PR #8, opened 2016-03-05
Parametrize

### LINKED ISSUE #1: Parametrization
Building a bit on https://www.reddit.com/r/Python/comments/48qz7w/nox_virtualenvtest_automation_similar_to_tox_but/d0m3q4l, another way to handle parametrization is like this:
```python
@nox.parametrize('python', ['/path/to/python2.7', '/path/to/python3.5', '/path/to/pypy', ...], ids=['py27', 'py34', 'pypy', ...])
@nox.parametrize('django', ['1.9', '1.8', ...])
def session_test(session, python, django):
   ...
```

This is quite identical to [pytest parametrization](http://pytest.org/latest/parametrize.html?highlight=parametrize). 

The advantage over your initial proposal is that you can customize the display names (via `ids`).

Also, would be good to know how far and where you'd take nox. IOW, what goals does nox have? 

PS. There is no contribution guide.

jonparrott:
Fixes #1

/cc @ionelmc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,13 @@ matrix:
   - python: '2.7'
     env: NOX_SESSION=lint
   - python: '2.7'
-    env: NOX_SESSION=py27
+    env: NOX_SESSION=interpreters(version='2.7')
   - python: '3.4'
-    env: NOX_SESSION=py34
+    env: NOX_SESSION=interpreters(version='3.4')
   - python: '3.5'
-    env: NOX_SESSION=py35
+    env: NOX_SESSION=interpreters(version='3.5')
 install: pip install codecov .
-script: nox --session $NOX_SESSION
+script: nox --session "$NOX_SESSION"
 after_success: codecov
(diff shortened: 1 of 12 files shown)


ionelmc:
Awesome stuff :D

jonparrott:
@ionelmc before I merge, can I validate that this satisfies your use case?

ionelmc:
It mostly does. I miss a "skip" option tho - eg: you can't run django 1.9 on python 2.6. That could be added later right?

For reference, you can "skip a parameter" in pytest: http://pytest.org/latest/skipping.html?highlight=parametrize#skip-xfail-with-parametrize

jonparrott:
@ionelmc yeah. You're welcome to file another bug for that.

In the meantime, that one isn't hard to do within the session itself:

```python
def session_blah(session, foo):
    if foo in ('bad', 'worse'):
        print('Not running with arg {}.format(foo)
        return
```

PR #10, opened 2016-06-07
Log error when running session that doesn't exist.

tswast:
I found it difficult to get the escaping correct for parameterized
sessions. By returning an error I was able to debug this more easily.
(Also prevents me from thinking that I'm running tests when I'm actually
not.)

### DIFF
--- a/nox/main.py
+++ b/nox/main.py
@@ -17,6 +17,7 @@ from __future__ import print_function
 import argparse
 import imp
 from inspect import isfunction
+import itertools
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


jonparrott:
Instead of adding a new function and such, why not add a simple empty check [here](https://github.com/jonparrott/nox/blob/master/nox/main.py#L93)?

tswast:
With an empty check I would not catch errors like `nox -s tests doesnotexist` where `tests` is a real session and `doesnotexist` is not.

I can't check that the lengths of the config sessions and the running sessions match either, since a single name can map to multiple running sessions for a parameterized session.

jonparrott:
I see, how about just checking for overlapping sets there?

```set(requested_sessions).issubset(set(available_sessions))```?

tswast:
Done.

The set comparison had to be a little more complicated than that, since I had to extract the names and signatures, but much simpler than my original implementation.

jonparrott:
@tswast this looks great, but can you expand the tests to include a few parametric sessions?

tswast:
> can you expand the tests to include a few parametric sessions?

Done.

jonparrott:
Thanks, @tswast!

Issue #11, opened 2016-09-22
Can nox be used to run multiple sessions in a certain order?


tswast:
I would love to be able to run `nox` or `nox -r` and have it run tests, lint, and extract and compile translations.

The way I am thinking of doing this is providing two sessions for the two sets of command in [my translation guide](https://github.com/whogoesfirst/who-goes-first/blob/master/docs/TRANSLATING.md).

- `nox -s extract_translations` would run `pybabel extract -F babel.cfg -o messages.pot .` followed by `pybabel update -i messages.pot -d ./translations`
- `nox -s compile_translations` would run `pybabel compile -d ./translations`

The problem with doing it this way is that I want to be certain that the extract_translations session runs before the compile_translations session. I could combine them all in one, but in general I'd want to be able to extract first, edit the output in POEdit and then compile, but if I'm being lazy / don't want to translate everything right away I'd want to run both sessions.

Maybe I shouldn't worry about this and just do the extract_translations session for now? That's the most important one to run, I think, to make sure translators have the latest messages to work with.

tswast:
After sleeping on this, I think it was a stupid idea. I don't want to compile translations as part of my test suite.

tswast:
I changed my mind again. I think I should compile the messages as part of my tests so that I can see if the translated pages also worked. In that case I would want to make sure extract_translations runs before the test session.

tswast:
As a workaround I can make a tests session that does everything I want in order and just run that instead of the plain `nox` command.

That's probably better, anyway, since then I can reuse the bench rather than creating a new session for each. I'll close this again.

jonparrott:
So - a few things: nox isn't meant to replace make, which is probably a better fit for that use case.

However, I do think that nox should run sessions in the order you specify, so I'll leave this open for that.

PR #13, opened 2017-03-08
Provide a means to preserve session order.

lukesneeringer:
Okay @jonparrott, I am a convert to nox. It is better. Mostly. ;-)

One thing that bothers me quite a bit (like @tswast) is that nox does not honor the order that sessions are declared in `nox.py`, instead running your sessions in alphabetical order. I really want unit tests to run before system tests, for example, and also I want my unit tests to amalgamate coverage on each interpreter and spit out a unified report at the end.

The current implementation relies on reading `module.__dict__`, which is unordered, so I am proposing an API change. I have added a decorator, `@nox.session`, which causes nox to recognize a function as a session. This is intended to be the preferred approach instead of relying on the convention of naming functions `session_*`, although the previous functionality remains in place for backwards compatibility.

If accepted, this will work as follows:

  * The preferred way to declare a nox session is with `@nox.session`. The function name is retained as-is.
  * All sessions that use the decorator are run in order. Sessions that use the naming convention still run alphabetically. If you mix and match, all the decorator sessions run first (in order), and then the naming convention sessions (alphabetically).

This has one side effect:

  * The `nox.session` module is renamed `nox.sessions`, to avoid a naming collision. This is ostensibly a backwards incompatible change, but I am fairly certain the old module was unlikely to be directly used by a downstream user.

Also worth noting, I designed `SessionRegistry` to be a little more robust than what the current feature set needs. I could see several additional uses for it in the future. Two obvious examples I can think of:

  * "run everything with an interpreter of Python 3.6"
  * declaring registries with explicit names and then running all sessions in a given registry.

Feedback welcome. :-)


lukesneeringer:
Additionally: Fixes #11.

lukesneeringer:
Sadly I have no idea why the AppVeyor build is failing, and the "environment can only contain strings" error message is not helpful...

jonparrott:
Hmm.. doesn't pytest do discovery and preserve order? I wonder how they pull that off.

> Also worth noting, I designed SessionRegistry to be a little more robust than what the current feature set needs. I could see several additional uses for it in the future. Two obvious examples I can think of:

Yeah this is super overkill. Can we simplify it for now (YAGNI)? We'll have the more robust one in the git history if we need it in the future, and we won't have to change the interface.

lukesneeringer:
> Yeah this is super overkill. Can we simplify it for now (YAGNI)? We'll have the more robust one in the git history if we need it in the future, and we won't have to change the interface.

Sure, although you will have to merge without squash to have it in git history. :-)

jonparrott:
I'm happy to do that, just do an interactive rebase so that we have pretty
history. :)

lukesneeringer:
As long as I make the new one a single commit, we will have a pretty history. Each of the two commits there now do a distinct thing.

lukesneeringer:
@jonparrott The simplified registry is in. I do ask that you preserve the history in the second commit if possible (being able to "group" sessions into a registry and run a registry is something that I know I am going to want; it would be great for google-cloud-python).

jonparrott:
> being able to "group" sessions into a registry and run a registry is something that I know I am going to want; it would be great for google-cloud-python.

I'd like to take (another) page out of pytest's book and use `-k 'something'` to do matching and add `nox.mark`.

setup.py
```diff
@@ -23,7 +23,7 @@
 setup(
     name='nox-automation',
 
-    version='0.9.1',
+    version='0.9.2',
```
jonparrott: Go ahead and bump to `0.10.0`

lukesneeringer:
> I'd like to take (another) page out of pytest's book and use -k 'something' to do matching and add `nox.mark`.

Got it. I might implement that next week.

jonparrott:
> Got it. I might implement that next week.

Awesome. :)

jonparrott:
I was curious how Sphinx autodoc can preserve member order in the docs. Turns out it analyzes the source code to do so (see [here](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/pycode/__init__.py#L291)). I would love something like that here so we can continue to look like `py.test`, but alas, it's complicated.

lukesneeringer:
One nice thing is that even if we implement this functionality now, we could still implement that later. It would require that you continue to support both syntaxes, but I do not see that as a huge problem. (I might be biased, though, because I actually like the decorator syntax -- "explicit is better than implicit".)

jonparrott:
> I actually like the decorator syntax -- "explicit is better than implicit".)

I do too, I just want to try to remain as close to `py.test` as possible.

> is there anything else this requires?

No, I'm just working on perf. I'll give it another once-over and merge soon.

jonparrott:
@lukesneeringer this is released. :)

lukesneeringer:
Cheers! Thanks @jonparrott.

If time permits, I might do a couple of the other small features discussed in this issue later this week.

jonparrott:
> If time permits, I might do a couple of the other small features discussed in this issue later this week.

That would be awesome.

PR #14, opened 2017-03-15
Change to the nox.py directory by default.

lukesneeringer:
See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3135#discussion_r105708321

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import hashlib
+import inspect
 import os
 import re
 import unicodedata
@@ -80,10 +81,10 @@ class SessionConfig(object):
         can also be specified globally using the
         ``--reuse-existing-virtualenvs`` argument when running ``nox``."""
 
-    def chdir(self, dir):
+    def chdir(self, dir, debug=False):
         """Set the working directory for any commands that run in this
-        session."""
-        self._commands.append(ChdirCommand(dir))
+        session after this point."""
+        self._commands.append(ChdirCommand(dir, debug=debug))
 
     def run(self, *args, **kwargs):
         """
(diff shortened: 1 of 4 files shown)


nox/sessions.py
```diff
@@ -177,6 +178,19 @@ def __init__(self, name, signature, func, global_config):
 
     def _create_config(self):
         self.config = SessionConfig(posargs=self.global_config.posargs)
+
+        # By default, nox should quietly change to the directory where
+        # the nox.py file is located; however, do not be overly stringent
+        # about this. If we cannot easily inspect that location, just give up
+        # and skip that command.
+        try:
+            cwd = os.path.dirname(inspect.getsourcefile(self.func))
```
jonparrott: The noxfile itself is available in the `global_config` object: https://github.com/jonparrott/nox/blob/3f6d1df8d613c76b0cc7a5aec4797d389b3b86da/nox/main.py#L142 any reason not to use that directly?

lukesneeringer: I did not notice that. :-)

This would be a slight behavior difference, as if the noxfile imported a function from another directory, the noxfile's directory would be used, not the imported function's directory (which is the behavior I would want). I _think_ I do not care about that distinction, at least not yet, but it is worth asking.

jonparrott:
I think it's less surprising for it to be the noxfile dir, not the function
dir.

lukesneeringer:
Sold.

jonparrott:
> Hmm. Perhaps it actually should be both? Each nox session should freshly ensure that the cwd is the noxfile (right now it is undefined behavior as to what happens if a previous session changes the working directory).

Fair point, let's do that.

lukesneeringer:
Done.

jonparrott:
Thanks, @lukesneeringer!

PR #18, opened 2017-03-22
Properly transform interpreter on Windows

lukesneeringer:
Once complete, this PR will properly transform a standard `pythonX.Y` interpreter designation to the appropriate standard path on Windows.

### Do not merge.

At the moment, this is a "thrash PR" because AppVeyor is what I am using to test work. It is not yet intended for merge

Additionally, test coverage is below 100%. I am trying to figure out what code works properly before I write tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,10 +13,12 @@ environment:
     - PYTHON: "C:\\Python27"
       PYTHON_VERSION: "2.7.11"
       PYTHON_ARCH: "32"
+      NOX_ENV: "interpreters(version='2.7')"
 
     - PYTHON: "C:\\Python27-x64"
       PYTHON_VERSION: "2.7.11"
       PYTHON_ARCH: "64"
+      NOX_ENV: "interpreters(version='2.7')"
 
     # Python 3.4.4 is the latest Python 3.4 with a Windows installer
     # Python 3.4.4 is the overall latest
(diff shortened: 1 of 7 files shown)


lukesneeringer:
Okay, this is in a good state now.

@jonparrott The coverage will be just short of 100% when the tests are run exclusively on Python 2 or Python 3, but the union of them is 100%. I think Codecov will report it correctly.

appveyor.yml
```diff
@@ -13,32 +13,38 @@ environment:
     - PYTHON: "C:\\Python27"
       PYTHON_VERSION: "2.7.11"
       PYTHON_ARCH: "32"
+      NOX_ENV: "interpreters(version='2.7')"
```
jonparrott: Nox uses 'session' instead of 'env', so let's keep it consistent and use `NOX_SESSION` as the env var.

nox/virtualenv.py
```diff
@@ -67,6 +70,46 @@ def _clean_location(self):
...
 
     @property
+    def _resolved_interpreter(self):
+        """Return the interpreter, appropriately resolved for the platform.
+
+        Based heavily on tox's implementation (tox/interpreters.py).
+        """
+        # Sanity check: If there is no assigned interpreter, then
+        # do nothing.
+        if not self.interpreter:
```
jonparrott: `if self.interpreter is None:`?

nox/virtualenv.py
```diff
@@ -67,6 +70,46 @@ def _clean_location(self):
...
+        match = re.match(r'^python(?P<maj>\d)\.(?P<min>\d)$', self.interpreter)
+        if match:
+            version = match.groupdict()
+            potential_paths = (
+                r'c:\python{maj}{min}\python.exe'.format(**version),
+                r'c:\python{maj}{min}-x64\python.exe'.format(**version),
+            )
+            for path in potential_paths:
+                actual = py.path.local(path)
+                if actual.check():
```
jonparrott: why not just `if py.path.local(path).check():`?

lukesneeringer: Previously I was returning `str(actual)`, but I changed that to returning path because of details in how the unit tests trick the system.

lukesneeringer:
Also, once this is merged, would you mind cutting an 0.11 release? :-)

PR #21, opened 2017-03-27
Changing case for github and readthedocs in README.

dhermes:


jonparrott:
@lukesneeringer @dhermes both of you are now collaborators here. :)

dhermes:
I've been ensnared by another @jonparrott project!

PR #24, opened 2017-04-13
Print virtualenv path when re-using virtualenv

### LINKED ISSUE #23: When using `-r`, the hashed env name is never printed
```
nox > Running session unit_tests(python_version='3.6')
nox > The virtualenv name was hashed to avoid being too long.
nox > Re-using existing virtualenv.
```

vs.

```
nox > Running session unit_tests(python_version='3.6')
nox > The virtualenv name was hashed to avoid being too long.
nox > virtualenv /path/to/.nox/775478c5 -p python3.6

```

jonparrott:
Resolves #23


dhermes:
Thanks!

PR #32, opened 2017-06-05
Fixing bad link in README (bad on PyPI).

dhermes:
Also managed to do some editorializing in the README :grinning: 

Currently the HTML generated [on PyPI][1] is:

```
<a href="docs" rel="nofollow">Read the Docs</a>
```

[1]: https://pypi.python.org/pypi/nox-automation/0.16.1

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,11 +3,13 @@ Nox - Flexible test automation for Python
 
 |Build Status| |Coverage Status| |PyPI Version| |Docs|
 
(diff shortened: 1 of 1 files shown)


dhermes:
@jonparrott Do you have a

```
python setup.py check --restructuredtext --strict
```

check?

jonparrott:
@dhermes [nope](https://github.com/jonparrott/nox/blob/master/nox.py#L35). We could use one.

dhermes:
Nose goes

PR #33, opened 2017-06-06
Add a '--version' argument.

tseaver:


nox/main.py
```diff
@@ -263,8 +265,17 @@ def main():
...
         help='Arguments that are passed through to the sessions.')
+    parser.add_argument(
+        '--version', action='store_true',
+        help='Output the nox version and exit.')
 
     args = parser.parse_args()
+
+    if args.version:
+        dist = pkg_resources.get_distribution('nox-automation')
+        sys.stderr.write("{}\n".format(dist.version))
```
dhermes: Why write to stderr? Also, you could instead be a bit more "modern" and idiomatic:

```
from __future__ import print_function

print('blah', file=sys.stderr)
```

tseaver: https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout, which cites Doug McIlroy's [Unix Reader](http://pic.plover.com/UnixReader/reader.pdf):

> All programs placed diagnostics on the standard output. This had al ways caused trouble when the output was redirected into a file, but became intolerable when the output was sent to an unsuspecting process. Nevertheless, unwilling to violate the simplicity of the standard-input-standard-output model, people tolerated this state of affairs through v6. Shortly thereafter Dennis Ritchie cut the Gordian knot by introducing the standard error file. That was not quite enough. With pipelines diagnostics could come from any of several programs running simultaneously .Diagnostics needed to identify themselves. Thus began a never quite finished pacification campaign: a few recalcitrant diagnostics still remain anonymous or appear on the standard output.

tseaver: I dunno how to mock the majyk `print` built-in created by that `__future__`.

dhermes: You can use `six.print_` if it needs to be mocked

tseaver: OK, it is tedious, but 5cc56b1 switches to use `print()`.

tests/test_main.py
```diff
@@ -498,6 +502,22 @@ def test_main():
         assert config.posargs == ['1', '2', '3', '-f', '--baz']
 
 
+def test_main_version():
```
jonparrott: Can you please just use pytest's `capsys` instead of all of this mocking?

tseaver: OK: 0d3193d

tseaver: And lint fixed in 0ecaf17.

jonparrott:
Thanks, @tseaver!

PR #35, opened 2017-06-14
Add session.skip

### LINKED ISSUE #34: Add `session.skip(reason)`
See https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3475

This way sessions can

- Fail
- Succeed
- Be Skipped

jonparrott:
Resolves #34

### DIFF
--- a/nox/main.py
+++ b/nox/main.py
@@ -157,9 +157,11 @@ def print_summary(results):
         elif status == SessionStatus.SKIP:
             log = logger.warning
             status = 'skipped'
-        else:
+        elif status == SessionStatus.FAIL:
             log = logger.error
             status = 'failed'
+        else:
+            raise ValueError('Unexpected session status: {}'.format(status))
 
         log('* {}: {}'.format(
             session.signature or session.name, status))
--- a/nox/main.py
+++ b/nox/main.py
@@ -22,13 +22,13 @@ import json
 import os
 import sys
 
+import pkg_resources
+from six import iterkeys
+
 from nox import registry
 from nox._parametrize import generate_calls
 from nox.logger import logger, setup_logging
 from nox.sessions import Session, SessionStatus
-
-import pkg_resources
-from six import iterkeys
 
 
 class GlobalConfig(object):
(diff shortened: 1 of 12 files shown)


nox/sessions.py
```diff
@@ -56,6 +56,16 @@ class _SessionQuit(Exception):
...
 
 
+class _SessionSkip(Exception):
+    pass
+
+
+class SessionStatus(object):
+    FAIL = False
+    SUCCESS = True
+    SKIP = 3
```
dhermes: These states are inconsistent within themselves. Why not just 1, 2, 3? (It seems the answer is for "backwards-compatibility with `success`")

nox/main.py
```diff
@@ -155,17 +150,25 @@ def filter_sessions_by_keywords(keywords, available_sessions):
...
+    for session, status in results:
+        if status == SessionStatus.SUCCESS:
+            log = logger.success
+            status = 'passed'
+        elif status == SessionStatus.SKIP:
+            log = logger.warning
+            status = 'skipped'
+        else:
+            log = logger.error
+            status = 'failed'
```
dhermes: I would prefer instead `elif status == SessionStatus.FAIL` and then an `else` branch that did the equivalent of a panic in Golang

nox/sessions.py
```diff
@@ -168,12 +178,21 @@ def install(self, *args):
         self._commands.append(InstallCommand(args))
 
     def log(self, *args, **kwargs):
+        """Outputs a log during the session."""
```
dhermes: Why not have a `if args or kwargs` before logging here?

nox/sessions.py
jonparrott: Crap, I meant to make `SKIP = 2`. It's supposed to be `0, 1, 2`, I used True and False for consistency with `success` and also because `False is 0 is True`.

nox/sessions.py
jonparrott: Because the other two have a side-effect, this one does not.

nox/main.py
jonparrott: Done

PR #36, opened 2017-06-27
Small documentation updates

duggelz:


nox/sessions.py
```diff
@@ -99,7 +99,9 @@ def __init__(self, posargs=None):
 
     def chdir(self, dir, debug=False):
         """Set the working directory for any commands that run in this
-        session after this point."""
+        session after this point.
+
+        cd() and chdir() are identical."""
```
duggelz: "Identical"?  "Equivalent"?  "Aliases"?  "the same"?

jonparrott: `alias` is preferred. :)

duggelz: `cd() is an alias for chdir().`

PR #38, opened 2017-07-20
Refactor the list of sessions into a manifest class.

lukesneeringer:
Greetings.

I have a proposed refactoring that I would like to do (mostly done actually) regarding how nox comes up with its session list.

### The Status Quo

Right now, the `main` function discovers sessions based on the Noxfile, filters them based on CLI arguments, ends up with a `list`, and then does a simple loop over that list and executes them in order.

### Proposal

I want to refactor how nox stores its session list into a somewhat more powerful "manifest" class, implemented in this PR (`manifest.py`). The manifest takes the original list of functions that comes back from `discover_session_functions` and applies all the filters as before.

_Iterating_ over the manifest is slightly different, in an extremely important way. Internally, the manifest keeps a queue (a list of sessions yet to be yielded) and a consumed list (a list of sessions already yielded). Importantly, _this allows the queue of remaining sessions to be modified during iteration_.

### Why do I want this?

Basically, I want to be able to add a session to the queue from another session. In particular, I want to implement something similar to handlers in Ansible, so I could do something like this...

```python
@nox.session
@nox.parametrize(version=('2.7', '3.6'))
def unit_tests(session, version):
    # Run the unit tests
    cover.notify()   # <-- THIS IS THE IMPORTANT PART. :-)

@nox.session
@nox.parametrize(version=('2.7', '3.6'))
def system_tests(session, version):
    # Run the system tests

@nox.session
def docs(session):
    # Build the docs.

@nox.session
def lint(session):
    # Lint all the things.

@nox.session
def cover(session):
    # All the .coverage files were written by unit_test earlier.
    session.install('coverage')
    session.run('coverage', 'report', '--fail-under=100')
    session.run('coverage', 'erase')
```

I want this because right now I find myself issuing `nox -s "unit_tests(version='3.6')" ; nox -s cover` _a lot_. Really what I want is I always want cover to run if _any_ of the sessions that wrote out coverage data ran.

Obviously not everyone will want this, but I hope in principle you can see how this is a useful feature. :-)

### Conclusion

This does not yet implement the notify feature (although it becomes easy once the manifest is a thing). However, is this something you are willing to accept in principle?

The manifest code is there and works, I would just need to update tests (but `test_run` is kind of gnarly and I wanted to get design approval before I spent an hour adapting all the mocks and such).

### DIFF
--- a/nox/main.py
+++ b/nox/main.py
@@ -17,7 +17,6 @@ from __future__ import print_function
 import argparse
 import imp
 from inspect import isfunction
-import itertools
 import json
 import os
 import sys
@@ -26,9 +25,9 @@ import pkg_resources
 from six import iterkeys
 
 from nox import registry
-from nox._parametrize import generate_calls
 from nox.logger import logger, setup_logging
-from nox.sessions import Session, SessionStatus
+from nox.manifest import Manifest
+from nox.sessions import SessionStatus
 
 
 class GlobalConfig(object):
@@ -70,82 +69,6 @@ def discover_session_functions(module):
 
     # Return the final dictionary of session functions.
     return funcs
-
-
-def _null_session_func(session):
-    """A do-nothing session for patemetrized sessions that have no available
-    parameters."""
-    session.skip('This session had no parameters available.')
-
-
-def make_sessions(session_functions, global_config):
-    """Create session objects from the session functions and the global
-    configuration."""
-    sessions = []
-    for name, func in session_functions.items():
-        if not hasattr(func, 'parametrize'):
-            sessions.append(Session(name, None, func, global_config))
-        else:
-            calls = generate_calls(func, func.parametrize)
-            for call in calls:
-                session = Session(
-                    name, name + call.session_signature, call, global_config)
-                sessions.append(session)
-            if not calls:
-                # Add an empty, do-nothing session.
-                sessions.append(Session(
-                    name, None, _null_session_func, global_config))
-
-    return sessions
-
-
-def filter_sessions_by_name(specified_sessions, available_sessions):
-    """Filter sessions based on the user-specified names."""
-    sessions = [x for x in available_sessions if (
-        x.name in specified_sessions or
-        x.signature in specified_sessions)]
-
-    missing_sessions = set(specified_sessions) - set(
-        itertools.chain(
-            [x.name for x in sessions if x.name],
-            [x.signature for x in sessions if x.signature]))
-    if missing_sessions:
-        logger.error('Sessions {} not found.'.format(', '.join(
-            missing_sessions)))
-        return False
-
-    return sessions
-
-
-class KeywordLocals(object):
-    """Eval locals using keywords.
-
-    When looking up a local variable the variable name is compared against
-    the set of keywords. If the local variable name matches any *substring* of
-    any keyword, then the name lookup returns True. Otherwise, the name lookup
-    returns False.
-    """
-    def __init__(self, keywords):
-        self._keywords = keywords
-
-    def __getitem__(self, variable_name):
-        for keyword in self._keywords:
-            if variable_name in keyword:
-                return True
-        return False
-
-
-def keyword_match(expression, keywords):
-    """See if an expression matches the given set of keywords."""
-    locals = KeywordLocals(set(keywords))
-    return eval(expression, {}, locals)
-
-
-def filter_sessions_by_keywords(keywords, available_sessions):
-    """Filter sessions using pytest-like keyword expressions."""
-    return [
-        x for x in available_sessions
-        if keyword_match(keywords, [x.signature or x.name])]
 
 
 def print_summary(results):
@@ -202,28 +125,27 @@ def run(global_config):
         return False
 
     session_functions = discover_session_functions(user_nox_module)
-    sessions = make_sessions(session_functions, global_config)
+    manifest = Manifest(session_functions, global_config)
 
-    if global_config.sessions:
-        sessions = filter_sessions_by_name(global_config.sessions, sessions)
-
-    if global_config.keywords:
-        sessions = filter_sessions_by_keywords(
-            global_config.keywords, sessions)
-
+    # If the user just asked for a list of sessinos, print that
+    # and be done.
     if global_config.list_sessions:
         print('Available sessions:')
-        for session in sessions:
+        for session in manifest:
             print('*', session.signature or session.name)
         return True
 
-    if not sessions:
+    if not manifest:
         return False
 
     success = True
     results = []
 
-    for session in sessions:
+    # Iterate over each session in the manifest, and execute it.
+    #
+    # Note that it is possible for the manifest to be altered in any given
+    # iteration.
+    for session in manifest:
         result = session.execute()
         results.append((session, result))
         success = success and result
--- a/nox/main.py
+++ b/nox/main.py
@@ -127,7 +127,7 @@ def run(global_config):
     session_functions = discover_session_functions(user_nox_module)
     manifest = Manifest(session_functions, global_config)
 
-    # If the user just asked for a list of sessinos, print that
+    # If the user just asked for a list of sessions, print that
     # and be done.
     if global_config.list_sessions:
         print('Available sessions:')
--- a/nox/main.py
+++ b/nox/main.py
@@ -12,22 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import print_function
+"""The nox `main` module.
+
+This is the entrypoint for the ``nox`` command (specifically, the ``main``
+function). This module primarily loads configuration, and then passes
+control to :meth:``nox.workflow.execute``.
+"""
+
+from __future__ import absolute_import, print_function
 
 import argparse
-import imp
-from inspect import isfunction
-import json
 import os
 import sys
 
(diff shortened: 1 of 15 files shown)


jonparrott:
This is a great idea. I'm all for it.

Feel free to delete/refactor the `test_run` stuff if it makes it easier for you, as long as we maintain coverage.

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Args:
+            keywords (Collection[str]): A collection of keywords which
+                session names are checked against.
+        """
+        self._queue = [
+            x for x in self._queue
+            if keyword_match(keywords, [x.signature or x.name])]
+
+    def make_session(self, name, func):
```
jonparrott: Can you pull this into its own function? It'll make it a bit easier to test.

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+
+from nox._parametrize import generate_calls
+from nox.logger import logger
+from nox.sessions import Session
+
+
+class Manifest(object):
+    """Session manifest.
+
+    The session manifest provides the final list of sessions that should be
```
jonparrott: it's not really `final` is it? :)

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+    run by nox.
+
+    It is possible for this to be mutated during execution. This allows for
+    useful use cases, such as for one session to "notify" another.
```
jonparrott: `notify` seems weird to me, how would you feel about the terminology 'chain'? e.g.:

```
This allows for a session to chain into another session.
```

nox/main.py
```diff
@@ -202,28 +125,27 @@ def run(global_config):
...
-
-    if global_config.sessions:
-        sessions = filter_sessions_by_name(global_config.sessions, sessions)
-
-    if global_config.keywords:
-        sessions = filter_sessions_by_keywords(
-            global_config.keywords, sessions)
+    manifest = Manifest(session_functions, global_config)
 
+    # If the user just asked for a list of sessinos, print that
```
dhermes: s/sessions/sessions

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+    Args:
+        session_functions (Mapping[str, function]): The registry of discovered
+            session functions.
+        global_config (): The global configuration.
+    """
+    def __init__(self, session_functions, global_config):
+        self._all_sessions = []
+        self._config = global_config
+
+        # Create the sessions based on the provided session functions.
```
jonparrott: I generally prefer dumb initializers. How would you feel if the usage of this class was closer to this:

```
manifest = Manifest()

manifest.add_filter(filter_by_name(global_config.sessions))
manifest.add_filter(filter_by_keyword(global_config.keywords))

for session_func in session_funcs:
    manifest.add_session(session_func, global_config)
```

or

```
manifest = Manifest()
...
manifest.add_sessions(session_funcs, global_config)
```

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+    The session manifest provides the final list of sessions that should be
+    run by nox.
+
+    It is possible for this to be mutated during execution. This allows for
+    useful use cases, such as for one session to "notify" another.
+
+    Args:
+        session_functions (Mapping[str, function]): The registry of discovered
+            session functions.
+        global_config (): The global configuration.
```
dhermes: Missing type info

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    def __init__(self, session_functions, global_config):
+        self._all_sessions = []
+        self._config = global_config
+
+        # Create the sessions based on the provided session functions.
+        for name, func in session_functions.items():
```
dhermes: `six.iteritems`? Am I lame?

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+            self.make_session(name, func)
+
+        # By default, every available session starts off in the queue.
+        self._queue = copy.copy(self._all_sessions)
```
dhermes: Why not just `self._all_sessions[::]`?

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+        # If the configuration specified filters, apply those now.
+        if self._config.sessions:
+            self.filter_by_name(self._config.sessions)
+        if self._config.keywords:
+            self.filter_by_keywords(self._config.keywords)
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
```
dhermes: You need a `next = __next__` to make this work in Python 2.

nox/manifest.py
```diff
@@ -0,0 +1,166 @@
...
+        """Return the next item in the queue.
+
+        Raises:
+            StopIteration: If the queue has been entirely consumed.
+        """
+        if not len(self._queue):
```
dhermes: I think you mean `if not self._queue`?

nox/manifest.py
lukesneeringer: HA! :-) That is, in fact, kind of the point. I intended to communicate "source of truth for". Will update.

nox/manifest.py
lukesneeringer: So, it's not _quite_ a chain. For instance, say `unit_tests(interpreter='2.7')` and `unit_tests(interpreter='3.6')` _both_ add the `cover` session to the manifest. The ultimate idea for me is that cover (1) gets added to the very end (e.g. it does not run immediately after the first thing to add it), and (2) gets added once.

The terminology I am using here is [borrowed from Ansible](http://docs.ansible.com/ansible/latest/playbooks_intro.html#handlers-running-operations-on-change), and the best term I could think of. I think `chain` is wrong -- it implies a slightly different thing to me (run this session _next_). But maybe there is a different term that would be better?

It is worth noting that there is no reason why "notify" and "chain" (as I am defining the terms in this comment) could not _both_ be implemented. There are valid use cases for each.

nox/manifest.py
lukesneeringer: This list is never big enough for me to care. :-)

nox/manifest.py
lukesneeringer: Later, I have:

```python
def next(self):
    return self.__next__()
```

nox/manifest.py
lukesneeringer: Explicit is better than implicit. :-)

nox/manifest.py
lukesneeringer: Advantages:

  * There are potentially more ways to customize the use of the manifest.

Disadvantages:

  * It pulls that logic back into `main.py`, and I think there is value in reducing logic there.

Probably a net win overall.

nox/manifest.py
lukesneeringer: Basically the [don't make me think](https://en.wikipedia.org/wiki/Don't_Make_Me_Think) principle. The `[::]` idiom, while certainly used, is not something that I think most people see every day, and I still have to do cognitive work to say, "oh right; it is making a shallow copy".

Or I can just import `copy` and explicitly write, "I am making a shallow copy." I believe this is better for readability.

nox/manifest.py
lukesneeringer: Update: This is a term that will be part of the user-facing API, so it definitely matters to get it right. :-)

jonparrott: Hrm, in that case i'm leaning towards using Ansible's `notify` as a loanword. Nox is intended to be a meta-pytest so I've mostly tried to pull terminology from there, but I can't really find anything there that makes sense. `queue` might work, but it feel it's far too general.

nox/manifest.py
jonparrott: Could we make it a dumb initializer with a helper constructor then?

nox/manifest.py
jonparrott: pylint actually yells about this. I don't feel strongly, though (and I don't use pylint on nox).

lukesneeringer:
Just pinging to say I intend to finish this next week.

nox/manifest.py
lukesneeringer: `notify` it will remain, unless we think of something better. :-) And yes, I agree with mostly pulling terminology from py.test, but too saw no candidate.

nox/manifest.py
lukesneeringer: That sounds like a reasonable compromise. :-)

lukesneeringer:
> Feel free to delete/refactor the test_run stuff if it makes it easier for you, as long as we maintain coverage.

@jonparrott Please see lukesneeringer/nox#1 for this refactor. It ended up being much more significant than I planned (I refactored both the function and the tests) so I sent it up separately.

lukesneeringer:
Coverage back to 100% on lukesneeringer/nox#1.

Once this gets in, I can actually write the (absolutely trivial by comparison) useful features I want that depend on it. :-p

nox/manifest.py
jonparrott: I would still prefer this to be a separate function.

lukesneeringer: I can change this. Would it be okay if we merged this and I will follow up with that PR? Just so it does not get even longer?

jonparrott: SG

jonparrott:
LGTM, but travis is failing

PR #39, opened 2017-08-21
Addition of session.notify

lukesneeringer:
This adds the `session.notify` command, exemplified in nox's own `nox.py` module, where any of the four interpreters trigger the `cover` session.

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -60,7 +60,7 @@ class Command(object):
         self.path = path
         self.success_codes = success_codes or [0]
 
-    def run(self, path_override=None, env_fallback=None):
+    def run(self, path_override=None, env_fallback=None, **kwargs):
         path = self.path if path_override is None else path_override
 
         env = env_fallback.copy() if env_fallback is not None else None
@@ -169,3 +169,15 @@ class InstallCommand(object):
 
     def __str__(self):
         return ' '.join(self.deps)
+
+
+class NotifyCommand(Command):
+    """Notify the given session and add it to the queue."""
+
+    def __init__(self, target, debug=False):
+        self.target = target
+        self.debug = debug
+
+    def __call__(self, session, **kwargs):
+        logger.info('Notifying session: %s' % self.target)
+        session.manifest.notify(self.target)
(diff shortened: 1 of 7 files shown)


nox/command.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, args, env=None, silent=False, path=None,
         self.path = path
         self.success_codes = success_codes or [0]
 
-    def run(self, path_override=None, env_fallback=None):
+    def run(self, path_override=None, env_fallback=None, **kwargs):
```
jonparrott: why add `kwargs` and not use them?

nox/manifest.py
```diff
@@ -164,6 +164,37 @@ def make_session(self, name, func):
     def next(self):
         return self.__next__()
 
+    def notify(self, session):
```
jonparrott: How does this work with parametrized sessions?

lukesneeringer: You can specify either the name or the parameterized name:

So both of these would work:

  * `notify('unit_tests')`
  * `notify("unit_tests(python_version='3.6')")`

nox/command.py
lukesneeringer: The invocation of `Command` is now sending an additional argument, which is used by `NotifyCommand.__call__`. This means I do not have to have a separate if check for every command subclass (there could eventually be more).

nox/manifest.py
jonparrott: It would be nice t be able to do `notify(unit_test, python_version='3.6')`

nox/command.py
jonparrott: Gotcha

nox/manifest.py
lukesneeringer: It would have to be `notify('unit_tests', python_version='3.6')`. I agree that would be nice but assert it can be separate. (Both forms can work.)

jonparrott: that's fine

jonparrott: Do you want to add this in a follow-up PR? If so, LGTM.

PR #48, opened 2017-09-18
Making sure all `Command`s use run in `__call__`.

dhermes:
This way subclasses don't have to also set `__call__ = run`, they can just over-ride `run`.

**ASIDE 1**: @jonparrott Do you want me to write docstrings?
**ASIDE 2**: @jonparrott Is there a reason `InstallCommand` **does not** subclass `Command`?
**ASIDE 3**: I **would** label this as `hygiene`, but the label doesn't exist.

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -112,7 +112,8 @@ class Command(object):
             logger.error('Interrupted...')
             raise
 
-    __call__ = run
+    def __call__(self, *args, **kwargs):
+        return self.run(*args, **kwargs)
 
 
 class FunctionCommand(object):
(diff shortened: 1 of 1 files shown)


jonparrott:
> ASIDE 1: @jonparrott Do you want me to write docstrings?

No need to write trivial docstrings (this is *not* a user-facing API)

> ASIDE 2: @jonparrott Is there a reason InstallCommand does not subclass Command?

Probably just an oversight. Wanna see if inheriting breaks anything?

dhermes:
> Probably just an oversight. Wanna see if inheriting breaks anything?

Done and pushed.

dhermes:
ISTM it will make `Pylint` angry that the constructor doesn't call `super()`.

jonparrott:
> ISTM it will make Pylint angry that the constructor doesn't call super().

I see - undo that for now, I think we just need to revisit and make the base class more abstract.

dhermes:
@jonparrott Undid it. Unclear why the AppVeyor build is failing (it's the same error message I saw on Travis when updating `InstallCommand(object)` -> `InstallCommand(Command)`).

PR #53, opened 2017-09-19
Locate the Python interpreter via the py.exe launcher

pfmoore:
On Windows, if the default strategy for finding a versioned interpreter fails, ask the `py.exe` launcher to locate it for us.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -64,21 +64,18 @@ class ProcessEnv(object):
             path=self.bin if in_venv else None).run()
 
 
-def locate_via_py(v_maj, v_min):
+def locate_via_py(version):
     """Find the Python executable using the Windows launcher.
 
     Code taken from tox (tox/interpreters.py).
     """
-    ver = "-%s.%s" % (v_maj, v_min)
     script = "import sys; print(sys.executable)"
-    py_exe = distutils.spawn.find_executable('py')
+    py_exe = py.path.local.sysfind('py')
     if py_exe:
-        proc = subprocess.Popen(
-            (py_exe, ver, '-c', script), stdout=subprocess.PIPE,
-        )
-        out, _ = proc.communicate()
-        if not proc.returncode:
-            return out.decode('UTF-8').strip()
+        try:
+            return py_exe.sysexec('-' + version, '-c', script).strip()
+        except py.process.cmdexec.Error:
+            return None
 
 
 class VirtualEnv(ProcessEnv):
@@ -123,21 +120,11 @@ class VirtualEnv(ProcessEnv):
         # If this is a standard Unix "pythonX.Y" name, it should be found
         # in a standard location in Windows, and if not, the py.exe launcher
         # should be able to find it from the information in the registry.
-        match = re.match(r'^python(?P<maj>\d)\.(?P<min>\d)$', self.interpreter)
+        match = re.match(r'^python(?P<ver>\d\.\d)$', self.interpreter)
         if match:
-            version = match.groupdict()
-            # First try some standard locations.
-            potential_paths = (
-                r'c:\python{maj}{min}\python.exe'.format(**version),
-                r'c:\python{maj}{min}-x64\python.exe'.format(**version),
-            )
-            for path in potential_paths:
-                if py.path.local(path).check():
-                    return str(path)
-            # If we're still struggling, ask the Python launcher.
-            # TODO: Should this go first? The code above prefers 32-bit
-            # versions, which may not be what the user wants...
-            path_from_launcher = locate_via_py(version['maj'], version['min'])
+            version = match.group('ver')
+            # Ask the Python launcher to find the interpreter.
+            path_from_launcher = locate_via_py(version)
             if path_from_launcher:
                 return path_from_launcher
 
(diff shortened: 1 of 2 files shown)


pfmoore:
Some outstanding questions:

* Does this need tests? It probably should, but the existing tests say "Not testing multiple interpreters on Windows" and I can see why - it's going to be quite hard to test this without mocking a lot of stuff (and possibly reqriting some of the code to make it more testable)
* Should this check *replace* the existing hard-coded list of interpreters? In theory, it's more reliable and the `py.exe` launcher should always be present.
* If we do keep both approaches, should the default list or the `py.exe` check be first?

jonparrott:
Thank you so much for doing this, @pfmoore!

> Does this need tests? It probably should, but the existing tests say "Not testing multiple interpreters on Windows" and I can see why - it's going to be quite hard to test this without mocking a lot of stuff (and possibly reqriting some of the code to make it more testable)

I can handle this - it's likely we'll need to mock some stuff for coverage purposes but have a real test for when on an actual windows platform.

> Should this check replace the existing hard-coded list of interpreters? In theory, it's more reliable and the py.exe launcher should always be present.

I think so, yes.

> If we do keep both approaches, should the default list or the py.exe check be first?

Let's just drop the old method.

nox/virtualenv.py
```diff
@@ -62,6 +64,23 @@ def run(self, args, in_venv=True):
...
 
 
+def locate_via_py(v_maj, v_min):
+    """Find the Python executable using the Windows launcher.
+
+    Code taken from tox (tox/interpreters.py).
+    """
+    ver = "-%s.%s" % (v_maj, v_min)
+    script = "import sys; print(sys.executable)"
+    py_exe = distutils.spawn.find_executable('py')
```
jonparrott: TIL about `distutils.spawn.find_executable`! Should we use that over [`py.path.local.sysfind`](http://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.sysfind) which we use in [command](https://github.com/jonparrott/nox/blob/f87525654265afdb3da086d4116fdfb5867154f6/nox/command.py#L35)?

pfmoore: Honestly, I doubt it matters (I lifted that code from tox unchanged). I'd be inclined to go for consistency and follow current practice by using `py.path.local.sysfind` in both places. I'll change this.

(BTW, there's also `shutil.which` in Python 3.3+. Who said "one obvious way"? :smile:)

jonparrott: > (BTW, there's also shutil.which in Python 3.3+. Who said "one obvious way"? 😄)

I'd love to make nox 3+ only. Sigh.

pfmoore: I'll also make the change to drop the old approach, as noted in your other comment. I can do the tests too, if you want, but if you have a preferred style/approach for the tests and would rather cover that yourself, that's fine too.

jonparrott: We use pytest so as long as it's not terribly different from what's there go for it. :)

pfmoore:
Looking into the coverage failures...

pfmoore:
There's a remaining coverage failure - line 153 in `nox/virtualenv.py`:

```python
        if self.interpreter:
            cmd.extend(['-p', self._resolved_interpreter])
```

I'm not sure why the changes I've made would have caused this line to no longer be executed. @jonparrott do you have any suggestions?

jonparrott:
> I'm not sure why the changes I've made would have caused this line to no longer be executed. @jonparrott do you have any suggestions?

It's my fault, you can disregard that coverage issue.

pfmoore:
I like that sort of answer ;-) In which case I think this is good to go. Could you check it over for me?

jonparrott:
LGTM- I updated the docstring to mention pep 397. I'm gonna wait for CI to merge (I know it'll fail due to coverage, just want to make sure that's the only failure).

jonparrott:
Thank you so much for doing this @pfmoore! 🍰

jonparrott:
All good. Merging. :)

pfmoore:
Glad to help :smile:

dhermes:
:tada: Yay Windows!

nox/virtualenv.py
```diff
@@ -62,6 +62,24 @@ def run(self, args, in_venv=True):
...
+
+    This is based on :pep:397 which details that executing
+    ``py.exe -{version}`` should execute python with the requested
+    version. We then make the python process print out its full
+    executable path which we use as the location for the version-
+    specific Python interpreter.
+    """
+    script = "import sys; print(sys.executable)"
+    py_exe = py.path.local.sysfind('py')
+    if py_exe:
```
dhermes: @jonparrott Should this be a [`None`-check][1]?

[1]: http://iruntheinternet.com/lulzdump/images/nunchucks-nuns-chuck-norris-nun-12593488599.jpg

jonparrott: Possibly. If you want you can send a follow-up PR.

dhermes: Done #59

dhermes:
BTW I am living life happily in Windows land and I can confirm this PR works great (and `nox` was broken-ish before I `pip install git+https://github.com/jonparrott/nox/`)

jonparrott:
Yay!

PR #59, opened 2017-09-20
Adding None-check to `locate_via_py`.

dhermes:
Follow-up to #53. Also

- Using single quotes instead of double quotes
- Adding Args/Returns section to docstring

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -70,10 +70,17 @@ def locate_via_py(version):
     version. We then make the python process print out its full
     executable path which we use as the location for the version-
     specific Python interpreter.
+
+    Args:
+        version (str): The desired Python version.
+
+    Returns:
+        Optional[str]: The full executable path for the Python ``version``,
+        if it is found.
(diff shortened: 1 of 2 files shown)


jonparrott:
Seems it needs to *not* be a none-check.

dhermes:
@jonparrott It was just the mocks.

```python
In [1]: import py

In [2]: py.path.local.sysfind?
Signature: py.path.local.sysfind(name, checker=None, paths=None)
Docstring:
return a path object found by looking at the systems
underlying PATH specification. If the checker is not None
it will be invoked to filter matching paths.  If a binary
cannot be found, None is returned
Note: This is probably not working on plain win32 systems
but may work on cygwin.
File:      ~/.pyenv/versions/3.6.2/lib/python3.6/site-packages/py/_path/local.py
Type:      method

In [3]: py.path.local.sysfind('python')
Out[3]: local('/home/dhermes/.pyenv/versions/3.6.2/bin/python')
```

PR #64, opened 2017-10-12
Adding fast-finish to AppVeyor build matrix.

dhermes:
This way one failure stops the entire build.


dhermes:
This will fail because it doesn't have the changes from #65, but it is a config only change, so I am merging.

PR #68, opened 2017-10-16
Simplifying the AppVeyor config a slight bit.

dhermes:
This PR also adds 3.6 testing (not sure why it was absent before).

`install.ps1` is not necessary unless you really want the latest versions of each Python (i.e. `3.6.3` is latest but AppVeyor images are still on `3.6.1`)


appveyor.yml
```diff
@@ -6,54 +6,35 @@ matrix:
...
+    # Pre-installed Python versions, which AppVeyor may upgrade to
     # a later point release.
     # See: http://www.appveyor.com/docs/installed-software#python
 
-    # Python 2.7.11 is the latest Python 2.7 with a Windows installer
-    # Python 2.7.11 is the overall latest
-    # https://www.python.org/ftp/python/2.7.11/
     - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.11"
-      PYTHON_ARCH: "32"
```
dhermes: @jonparrott These env. vars are used by many helpers e.g. the common `CMD_IN_ENV` helpers, but they are largely irrelevant if you never need the Windows SDK. (They were in theory being using by `install.ps1`, but in practice never used.)

jonparrott:
Approved, but AppVeyor is failing?

dhermes:
> AppVeyor is failing?

Very strange! I saw this failure yesterday and also couldn't explain it. I think it means `py.exe` is not working "correctly" on AppVeyor (at least for Python 3.6).

I will remove Python 3.6 from this PR. If it passes, I can send a new PR and we can debug?

PR #69, opened 2017-10-17
Adding Python 3.6 to AppVeyor.

dhermes:
This is a follow-up to #68.

It **appears** that `py.exe` is broken on AppVeyor for Python 3.6.


jonparrott:
So any idea what's going on here?

dhermes:
Nope. I'd guess that the Python 3.6 binaries were somehow installed incorrectly and they are not "registered" with `py.exe`.

jonparrott:
That could be the culprit because as far as I know `py.exe` uses the registry to locate Python executables. Maybe @pfmoore has some insight?

jonparrott:
FYI here's the source code for the launcher: https://github.com/python/cpython/blob/master/PC/launcher.c

dhermes:
I can confirm the launcher works just fine on Windows with 3.6 (32-bit and 64-bit) from my own laptop. That's why I said it may be an issue with the AppVeyor install.

pfmoore:
I've just pushed a test as PR #71 - see the result [here](https://ci.appveyor.com/project/jonparrott/nox/build/1.0.265.master/job/coaf07rpx38xrcub). Looks like there is a problem at appveyor :-(

dhermes:
So I checked out the [`py.exe` PEP][1] and used [`reg.exe`][3] to [check][2] the registry for the info:

```
> reg.exe query HKEY_CURRENT_USER\Software\Python\PythonCore

HKEY_CURRENT_USER\Software\Python\PythonCore
    DisplayName    REG_SZ    Python Software Foundation
    SupportUrl    REG_SZ    http://www.python.org/

HKEY_CURRENT_USER\Software\Python\PythonCore\3.5
HKEY_CURRENT_USER\Software\Python\PythonCore\3.5-32
HKEY_CURRENT_USER\Software\Python\PythonCore\3.6
HKEY_CURRENT_USER\Software\Python\PythonCore\3.6-32
> reg.exe query HKEY_LOCAL_MACHINE\Software\Python\PythonCore

HKEY_LOCAL_MACHINE\Software\Python\PythonCore\2.6
HKEY_LOCAL_MACHINE\Software\Python\PythonCore\2.7
HKEY_LOCAL_MACHINE\Software\Python\PythonCore\3.3
HKEY_LOCAL_MACHINE\Software\Python\PythonCore\3.4
> reg.exe query HKEY_LOCAL_MACHINE\Software\Wow6432Node\Python\PythonCore

HKEY_LOCAL_MACHINE\Software\Wow6432Node\Python\PythonCore\2.6
HKEY_LOCAL_MACHINE\Software\Wow6432Node\Python\PythonCore\2.7
HKEY_LOCAL_MACHINE\Software\Wow6432Node\Python\PythonCore\3.3
HKEY_LOCAL_MACHINE\Software\Wow6432Node\Python\PythonCore\3.4
```

It seems that `3.5`, `3.5-32`, `3.6` and `3.6-32` are handled differently from the other versions, but it's unclear why `3.5` works fine and the rest do not.

[1]: https://www.python.org/dev/peps/pep-0514/#structure
[2]: https://ci.appveyor.com/project/jonparrott/nox/build/1.0.267.master/job/lvf0emkv5b6rnak6
[3]: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/reg-query

dhermes:
So I wrote a Python script `winreg_hack.py` that will recursively go through the registry and print all values. (See https://ci.appveyor.com/project/jonparrott/nox/build/1.0.271.master/job/srnhfad44h4hn8nd)

I ran it and captured it to a text file on my local machine (checked into this PR as `winreg.txt`) and on AppVeyor (checked into this PR as `appveyor_winreg.txt`).

dhermes:
Based on https://github.com/dhermes/nox/blob/appveyor-add-36/winreg_output/35-32.diff, ISTM that `HKEY_CURRENT_USER\Software\Python\PythonCore\3.5-32\PythonPath` is not populated on AppVeyor, but it needs to be. (Though the value is populated for 3.6 32- and 64-bit and they still don't work right.)

dhermes:
OK @jonparrott I have arrived at a solution: manually check in a **working** `py.exe` and replace the broken one.

I figured out what the issue is and made a simple [reproducible case][1]. The old launchers (3.4.4 in this case) didn't allocate enough space for the version name (e.g. `3.5-32\0`) in the registry and so iterating over versions does not complete.

----

@jonparrott @pfmoore I had an "impossible" time replacing `py.exe` using one of the standard installers, e.g. with [`python-3.6.3-amd64.exe`][2] I tried

```
cmd /c start /wait python-3.6.3-amd64.exe /quiet TargetDir="C:\\Python35-x64" Shortcuts=0 Include_launcher=1 InstallLauncherAllUsers=1
```

and the original `py.exe` was left intact. In the graphical installer, the "Install launcher for all users (recommended)" option is grayed out

installer

This makes me wonder if there is some (secret, to me) location in the registry that tells the installer: "I already have a `py.exe`, thanks but no thanks".

[1]: https://github.com/dhermes/appveyor-py-launcher
[2]: https://www.python.org/ftp/python/3.6.3/python-3.6.3-amd64.exe

pfmoore:
Out of curiosity, has anyone reported this to Appveyor? In my experience, they are usually pretty good about fixing things like this.

I've been letting this thread fly by without reading it in full, but I suspect that the problem is that they have py.exe from an older version of Python that doesn't understand how 3.5+ register themselves. The relevant PEP is [PEP 514](https://www.python.org/dev/peps/pep-0514/). The solution, as you worked out, is to install the latest launcher.

@zooba is the expert in the Windows installers - I wonder if he knows a simple way of saying "force upgrade the launcher to the latest version for all users"?

zooba:
Yep, we had to update the launcher for Python 3.5 when I fixed the bug where you couldn't reliably have the 32-bit and 64-bit versions installed side by side (by changing the registry key for the 32-bit version), and that combined with the Miniconda issue is what lead to doing PEP 514. If you're still on the 3.4 launcher, then that's the problem.

The easiest way to force upgrade the launcher is to download https://www.python.org/ftp/python/3.6.3/amd64/launcher.msi and run `msiexec /i launcher.msi ALLUSERS=1`. However, the problem here may be that this actually corrupts the 3.4 install by overwriting one of its "core" files, and if any sort of automatic repair gets run it may be "fixed" (for 3.5 the launcher was separated out so that installing/upgrading it on its own is safe - though we generally don't publicise the direct URL). It's also possible that AppVeyor blocks changes to files in the Windows directory, which would also prevent upgrading the launcher.

Copying the launcher directly into your local checkout is totally fine. It's designed for this and I do it all the time. Even if you only have Python 2 and don't have the right C runtime, it'll still be able to find and launch older versions.

dhermes:
> has anyone reported this to Appveyor

@pfmoore I reported it yesterday: https://github.com/appveyor/ci/issues/1870 (there is a link to this issue in my little [repro][1] repo).

@zooba Thanks for the info! I was just working on a change to build from source (`launcher.c`, it's self-contained) rather than to include the binary `py.exe`. Of the three choices:

1. Hard-copy of `py.exe` binary
1. Install with `launcher.msi`
1. Build from source

Which do you think is most appropriate?

> It's also possible that AppVeyor blocks changes to files in the Windows directory, which would also prevent upgrading the launcher.

I've tried "upgrading" `py.exe` on my **local machine** and the installers seem to just not worry about `py.exe` once they know it's already been installed prior. I even moved it to a different place on the filesystem with a different name and the installer (e.g. [`python-3.6.3-amd64.exe`][2]) still didn't repopulate `C:\Windows\py.exe`.

[1]: https://github.com/dhermes/appveyor-py-launcher
[2]: https://www.python.org/ftp/python/3.6.3/python-3.6.3-amd64.exe

zooba:
I'd just hard-copy the binary, but building from source is fine too if you can handle the multi-step build easily enough. (These days I personally use the nuget packages and don't rely on any system configuration at all, but that isn't for everyone :) ) The installer is going to register a shell extension and file associations that you don't need on a CI system.

A repair step is different from a regular install. Installs check whether the _product_ is already there (by looking for a registered GUID) and skip it if so, while a repair is going to check each file against other criteria (typically a version number and/or hash) and replace it. But simply running the install again - especially from the .exe - is not going to replace it.

Python 3.4 doesn't do this check in its installer. If you can re-run the installer (i.e. it doesn't abort itself for some other reason) it's always going to check and restore the launcher file. Python 3.5 and later may not.

dhermes:
@zooba I am not running a repair step. I'm using the 3.6 installer to uninstall and then manually re-installing (i.e., not a repair) from the GUI on a system where `C:\Windows\py.exe` is missing. After the install completes successfully, `C:\Windows\py.exe` is still missing. (See the screenshot [above][1], the installer makes it **seem** like it will install the launcher.)

[1]: https://github.com/jonparrott/nox/pull/69#issuecomment-337803399

dhermes:
So now the question is to you Mr. Maintainer Man @jonparrott:

> Build from source (it's [not so bad][1]) or hard-copy a working `py.exe` binary?

[1]: https://github.com/dhermes/nox/blob/appveyor-add-36/appveyor/build_py_launcher.cmd

jonparrott:
I'd hard-copy if it's just for testing purposes.

dhermes:
Then hard-copy we shall. I just reverted the previous commit ([build][1]) so we should be back at the hard-copied `py.exe` commit ([build][2]).

So we **can** merge (squash) before waiting on the current [build][3].

[1]: https://ci.appveyor.com/project/jonparrott/nox/build/1.0.300.master
[2]: https://ci.appveyor.com/project/jonparrott/nox/build/1.0.299.master
[3]: https://ci.appveyor.com/project/jonparrott/nox/build/1.0.299.master

pfmoore:
@zooba just to explain, the use case here is that the code relies on doing `py -X.Y` to reuse the interpreter location code from the launcher without needing to implement PEP 514 itself.

It might be nice if there was a Python library for doing that directly (it'd certainly avoid the problem here that nox is affected if the launcher installation is bugged). I know that at one point you were working on something like that - did it ever get anywhere? If not, I may take a stab at it myself.

dhermes:
Good call @pfmoore! The actual bug in `3.4.4` (i.e. the registry tag for versions was hard-coded as being at most 4 characters) would not be a thing if the launcher was written in pure Python.

pfmoore:
Ha - if I wrote a pure Python PEP 514 implementation, I bet I could introduce a few bugs :smile:

zooba:
@pfmoore Feel free to contribute to https://github.com/zooba/pep514tools

PR #80, opened 2018-01-11
Fixing tox-to-nox docs reference.

dhermes:
The extra is actually `tox_to_nox`, not `tox-to-nox`. Another acceptable way to "fix" this probably would be to rename the extra.


dhermes:
@jonparrott I made a commit to fix the [failure][1] (which isn't caused by this PR, but is very easy to fix).

[1]: https://travis-ci.org/jonparrott/nox/jobs/327834284

jonparrott:
Cool, merge at will

PR #81, opened 2018-01-11
Simulating `unittest.mock` backport in the Python 2 standard library.

dhermes:
Also sneaking in a missing license header on a file (I verified in the `git` history that the file is from 2016).


dhermes:
@lukesneeringer FYI this is a "nice" way to prepare for the EOL for 2.7.

Issue #79, opened 2018-01-10
Allow passing env to `session.install()`


dhermes:
For example, if I want to build a C extension in "debug" mode, I have to swap

```python
session.install('.')
```

for

```python
session.run('pip', 'install', '--upgrade', '.', env={'DEBUG': 'True'})
```

For context, here are the current [`InstallCommand` constructor][1] and [`SessionConfig.install()`][2].

[1]: https://github.com/jonparrott/nox/blob/v0.18.2/nox/command.py#L163-L165
[2]: https://github.com/jonparrott/nox/blob/v0.18.2/nox/sessions.py#L195

dhermes:
@theacodes @alanbato Might I "take over" PR #83 to fix this? Happy to send commits into it.

alanbato:
Sure! I think I was having trouble getting the tests to pass. Good luck! :)

theacodes:
Done in d7f4a0a

PR #83, opened 2018-01-18
Allow session.install() to be called with env param

alanbato:
This should allow the behavior desired in #79 

I relied on the subclassing of `ProcessEnv` by `VirtualEnv` to update the default environment with the one provided in the `session.install`. Since the call trace translates everything into a `Command` to be run, I rely on the validations made there instead of rolling my own.
If there's any way to make it cleaner or more reliable, please let me know.

Also, I added a test to verify this is working, but I'm not sure the way I did it is the best way to go, so if there's a better way to test this new behavior I'd be happy to hear it :)
🐍

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -167,5 +167,8 @@ class VirtualEnv(ProcessEnv):
 
         return True
 
-    def install(self, *args):
+    def install(self, *args, **kwargs):
+        env = kwargs.get('env', None)
+        if env:
+            self.env.update(env)
         self.run(('pip', 'install', '--upgrade') + args)
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -55,12 +55,12 @@ class ProcessEnv(object):
     def run(self, args, in_venv=True, env=None):
         """Runs a command. By default, the command runs within the
         environment."""
-        env_copy = self.env.copy() if in_venv else None
-        if env_copy is not None and env is not None:
-            env_copy.update(env)
+        if in_venv:
+            env = env or {}
+            env.update(self.env)
         return Command(
             args=args,
-            env=env_copy,
+            env=env,
             silent=True,
             path=self.bin if in_venv else None).run()
 
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -52,12 +52,15 @@ class ProcessEnv(object):
     def bin(self):
         return self._bin
 
-    def run(self, args, in_venv=True):
+    def run(self, args, in_venv=True, env=None):
         """Runs a command. By default, the command runs within the
         environment."""
+        env_copy = self.env.copy() if in_venv else None
+        if env_copy is not None and env is not None:
+            env_copy.update(env)
         return Command(
             args=args,
-            env=self.env if in_venv else None,
+            env=env_copy,
             silent=True,
             path=self.bin if in_venv else None).run()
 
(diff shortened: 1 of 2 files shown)


nox/virtualenv.py
```diff
@@ -167,5 +167,8 @@ def create(self):
 
         return True
 
-    def install(self, *args):
+    def install(self, *args, **kwargs):
+        env = kwargs.get('env', None)
+        if env:
+            self.env.update(env)
```
jonparrott: My concern here is that whatever env you pass to `install` becomes `sticky` (it's there even in the next call to `install`). Perhaps we should augment `run` to accept an `env` kwarg and pass that all the way down?

alanbato: The modified version of `run` looks like this now:
``` python
def run(self, args, in_venv=True, env=None):
        """Runs a command. By default, the command runs within the
        environment."""
        env_copy = self.env.copy() if in_venv else None
        if env_copy is not None and env is not None:
            env_copy.update(env)
        return Command(
            args=args,
            env=env_copy,
            silent=True,
            path=self.bin if in_venv else None).run()
 ```
But I think I relied on the `sticky`ness of my previous implementation to test it works, how would I go about it now? I don't know much about the way `mock_run.assert_called_with` works

Comment (alanbato) [reply nox/virtualenv.py]: Sound good :) I just pushed the changes that I showed here.

File: nox/virtualenv.py [hunk 2]
```diff
@@ -52,12 +52,15 @@ def __init__(self, bin=None, env=None):
     def bin(self):
         return self._bin
 
-    def run(self, args, in_venv=True):
+    def run(self, args, in_venv=True, env=None):
         """Runs a command. By default, the command runs within the
         environment."""
+        env_copy = self.env.copy() if in_venv else None
```
jonparrott: So this causes a weird situation where if you pass both `in_venv=False` and `env={...}` then `env` won't be applied.

Recommendation:

```python
command_env = {}
if in_venv:
   command_env.update(self.env)
```

alanbato: I thought about that use case too, but since `env` was only being passed down when `self.in_venv = true` I assumed that you were only able to use `env`s within a `venv` for some reason. Applying the suggested patch right now 👍

alanbato:
@jonparrott What did you think of the previous changes, and how can I include tests to prove it works? :)

jonparrott:
Hey @alanbato, just an FYI that I'm traveling for the next two weeks and won't be able to give this a lot of attention. I'd still love to see it merged, so if @dhermes wants to take it up we can get it in before then. :) Otherwise, I'll take a look as soon as I can.

alanbato:
Thanks for the heads up @jonparrott, enjoy your trip! I'm looking forward to see it merged too, but it can wait 😄

alanbato:
Hello @jonparrott! I hope you enjoyed your travels, just pinging you back in case you have a chance to look into this so we can work together and get this merged :)
Cheers!

nox/virtualenv.py
```diff
@@ -52,12 +52,15 @@ def __init__(self, bin=None, env=None):
     def bin(self):
         return self._bin
 
-    def run(self, args, in_venv=True):
+    def run(self, args, in_venv=True, env=None):
         """Runs a command. By default, the command runs within the
         environment."""
+        if in_venv:
+            env = env or {}
+            env.update(self.env)
```
jonparrott: We should be careful when modifying a parameter that was passed in. I'd prefer if we did `env.copy` here.

theacodes:
Superseded by d7f4a0a

PR #86, opened 2018-02-25
Don't skip install commands when re-using existing virtualenvs.

### LINKED ISSUE #82: `-r` flag should not skip `session.install()`
It seems the concept `--reuse-existing-virtualenvs` is getting mixed up with [`_should_install_deps`][2].
 [Here][1] is where the skip happens.

But if the `virtualenv` already exists and the deps have already been installed, then `pip install --upgrade` is usually a no-op.

The **one exception** (which is what past-tense @jonparrott may have been thinking) is that if the user has installed the "current" package, but not in editable (`-e`) mode, then it may confuse the `nox` users.

[1]: https://github.com/jonparrott/nox/blob/3614a4c926d91232c04b9da97ff6c5d541a2aa00/nox/sessions.py#L280-L285
[2]: https://github.com/jonparrott/nox/blob/3614a4c926d91232c04b9da97ff6c5d541a2aa00/nox/sessions.py#L234

jonparrott:
Resolves #82

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -92,9 +92,6 @@ By default nox deletes and recreates virtualenvs every time it is run. This is u
     nox -r
     nox --reuse-existing-virtualenvs
 
-.. note::
-    Re-using the existing virtualenv will skip any ``session.install()`` calls!
-
 
 Stopping if any session fails
 -----------------------------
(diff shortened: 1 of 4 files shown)


pradyunsg:
@theacodes Could you do a new release with the new changes on `master`? ✨

theacodes:
@pradyunsg done! https://pypi.org/project/nox-automation/0.19.0/

pradyunsg:
@theacodes Thanks! :)

PR #94, opened 2018-07-27
Make Nox imperative; Drop 2.7 and 3.4; Remove legacy naming support

theacodes:


nox/manifest.py
```diff
@@ -138,25 +154,43 @@ def make_session(self, name, func):
             Sequence[~nox.session.Session]: A sequence of Session objects
                 bound to this manifest and configuration.
         """
+        sessions = []
+
+        # If the func has the python attribute set to a list, we'll need to
+        # expand them.
+        if isinstance(func.python, (list, tuple, set)):
```
lukesneeringer: `collections.Collection`

lukesneeringer: Actually:

```python
if isinstance(func.python, collections.Collection) and not isinstance(func.python, str):
```

lukesneeringer: Nevermind, what you have is probably best.

nox/manifest.py
```diff
@@ -138,25 +154,43 @@ def make_session(self, name, func):
...
         # Simple case: If this function is not parametrized, then make
         # a simple session
         if not hasattr(func, 'parametrize'):
-            session = Session(name, None, func, self._config, self)
-            return (session,)
+            if func.python is not None:
+                long_name = '{}-{}'.format(name, func.python)
+            else:
+                long_name = name
+            session = SessionRunner(name, long_name, func, self._config, self)
```
lukesneeringer: Could you add the keywords here, even if it forces line breaks? There are enough arguments that it would be easier to read.

theacodes: Will do.

docs/config.rst
```diff
@@ -13,46 +13,89 @@ Defining sessions
...
 Nox sessions are configured via standard Python functions that are decorated
-with ``@nox.session`` or start with ``session_``. For example, these are all
-sessions::
+with ``@nox.session``. For example::
 
     import nox
 
-    def session_a(session):
+    @nox.session
+    def b(session):
```
lukesneeringer: Nitpick: The name `b` no longer makes sense as there is now only one sample. Perhaps `unit_tests` or something.

docs/config.rst
```diff
@@ -13,46 +13,89 @@ Defining sessions
...
         pass
 
-    def other_func(session):
+You can also specify that the virtualenv should *always* be reused or change the virtualenv name::
+
+    @nox.session(
+        python=nox.python(
+            python='3.6',
+            virtualenv='custom-name',
+            reuse=True))
```
lukesneeringer: I dislike the stutter that would be involved in this approach. Also, my intuition maintains a cognitive separation between the version of Python being used and the venv configuration as a whole. I would honestly rather see...

```python
@nox.session(python=['2.7', '3.5', '3.6', '3.7'], venv=nox.venv(prefix='...', reuse=True))
```

This has the advantage that you do not have to repeat the virtualenv configuration for every version of Python. It will be a (very) common use case to support 3-4 versions of Python and want virtualenv configuration, and the this involves repeating the `virtualenv=` and `reuse=` arguments over and over for that situation. Separating out the Python versions from the remaining virtualenv configuration is more concise and, I think, more intuitive.

The con to this approach is that you can not have radically distinct virtualenv configurations for different Python versions in the same session; e.g. you can not tell your Python 3.7 virtualenv to be cached and your Python 3.6 virtualenv to be reused. I think that is fine (nobody will want to do that in real life, basically).

If you _really_ think there needs to be different virtualenv configurations for different Python versions, I would rather see stacked `@nox.session` calls, e.g.

```python
@nox.session(python='3.6', virtualenv='tests-3.6')
@nox.session(python='3.7', virtualenv='tests-3.7', reuse=True)
def unit_tests(session):
    pass
```

docs/config.rst
```diff
@@ -13,46 +13,89 @@ Defining sessions
...
+    def tests(session):
+        pass
+
+When collecting your sessions, Nox will create a separate session for each interpreter. You can see these sesions when running ``nox --list-sessions``. For example this Noxfile::
+
+    @nox.session(python=['2.7', '3.5', 3.6', '3.7'])
+    def tests(session):
+        pass
+
+Will produce these sessions::
```
lukesneeringer: If there is only one interpreter, is the suffix applied? (I think it should be.)

docs/usage.rst
```diff
@@ -39,24 +39,16 @@ You can run every session by just executing `nox` without any arguments:
...
 By default nox will run all sessions defined in the noxfile. However, you can choose to run a particular set of them using ``--session``, ``-s``, or ``-e``::
 
-    nox --session py27
-    nox -s lint py27
-    nox -e py34
+    nox --session tests
+    nox -s lint tests
+    nox -e lint
 
 Nox will run these sessions in the same order they are specified.
```
lukesneeringer: I would really like a `-p/--python` argument to limit to sessions for a particular interpreter.

Example:
`nox -s tests -p 3.7`

nox.py
```diff
@@ -20,8 +20,8 @@
 ON_APPVEYOR = os.environ.get('APPVEYOR') == 'True'
 
 
-@nox.session
-def default(session):
+@nox.session(python=['3.4', '3.5', '3.6'])
```
lukesneeringer: I thought you support 3.5+ now; this should be `['3.5', '3.6', '3.7']`, yes?
(I care more about the presence of 3.7 than absence of 3.4.)

nox/registry.py
```diff
@@ -2,23 +2,49 @@
...
+
+def _to_python_config(value):
+    if not isinstance(value, PythonConfig):
+        return PythonConfig(python=value)
+    return value
+
+
+def session_decorator(func=None, python=None):
     """Designate the decorated function as a session."""
+    if func is None:
```
lukesneeringer: This is an unusual enough idiom that is worth explaining:

```python
# If `func` is provided, then this is the decorator call with the function
# being sent as part of the Python syntax (`@nox.session`).
# If `func` is None, however, then this is a plain function call, and it
# must return the decorator that ultimately is applied (`@nox.session(...)`).
#
# This is what makes the syntax with and without parentheses both work.
```

nox/registry.py
```diff
@@ -2,23 +2,49 @@
...
+        return functools.partial(session_decorator, python=python)
+
     # This adds the given function to the _REGISTRY ordered dictionary, which
     # is checked by `nox.main.discover_session_functions`.
+    if (isinstance(python, (list, tuple, set))
+            and not isinstance(python, str)):
```
lukesneeringer: This is redundant.

```
>>> issubclass(str, (list, tuple, set))
False
```

nox/sessions.py
```diff
@@ -74,48 +70,57 @@ class Status(enum.Enum):
...
+    def virtualenv(self):
+        """The virtualenv that all commands are run in."""
+        return self._runner.venv
 
-        cd() is an alias for chdir()."""
-        self._commands.append(ChdirCommand(dir, debug=debug))
+    @property
+    def bin(self):
+        """The bin directory for the virtualenv."""
+        return self._runner.venv.bin
```
lukesneeringer: What is this set to if there is no virtualenv? Or does it raise AttributeError?

nox/sessions.py
```diff
@@ -224,75 +228,50 @@ def skip(self, *args, **kwargs):
...
     def __init__(self, name, signature, func, global_config, manifest=None):
         self.name = name
         self.signature = signature
         self.func = func
         self.global_config = global_config
         self.manifest = manifest
+        self.venv = None
 
     def __str__(self):
         return utils.coerce_str(self.signature or self.name)
```
lukesneeringer: You presumably do not need `coerce_str` anymore, right? `str(self.signature or self.name)` should do the trick.

docs/config.rst
```diff
@@ -13,46 +13,89 @@ Defining sessions
...
+    def tests(session):
         pass
 
-    @nox.session
-    def b(session):
+You can also tell Nox to run your session against multiple Python interpreters. Nox will create a separate virtualenv and run the session for each interpreter you specify. For example, this session will run twice - once for Python 2.7 and once for Python 3.6::
+
+    @nox.session(python=['2.7', '3.6'])
+    def tests(session):
         pass
```
lukesneeringer: I suggest adding:

> When you provide a version number, Nox automatically prepends `python` to determine the name of the executable. However, Nox also accepts the full executable name. If you want to test using pypy, for example:
>
> ```python
> @nox.session(python=['2.7', '3.6', 'pypy-6.0'])
> def tests(session):
>     pass

docs/usage.rst
lukesneeringer: (Although it would be fair to do it in a follow-up, or tell me to do so. :-p)

theacodes: Let's do it in a separate PR https://github.com/theacodes/nox/issues/95

nox.py
theacodes: Done

nox/registry.py
theacodes: Good call. Done.

nox/registry.py
theacodes: Done

nox/sessions.py
theacodes: Raises AttributeError if `ProcessEnv` is being used instead. I think that's pretty fair.

nox/sessions.py
theacodes: Probably not, but I'll defer that for another PR for 2.7 cleanup.

theacodes:
@lukesneeringer thank you so much for taking the time to review this. I think it's ready for another round.

nox/sessions.py
lukesneeringer: Agreed.

docs/usage.rst
lukesneeringer: Ack.

docs/config.rst
lukesneeringer: This is sufficient for my goals. :-)

A few questions:

  1. Others had complained about "virtualenv name too long" warnings that nox used to give, especially those who used longer directory structures. Has anything changed, or will that complaint return?

  2. If you _do_ need more granular virtualenv configuration, what is the approach? New keyword arguments here? (That is fine if so.)

docs/config.rst
```diff
@@ -13,46 +13,78 @@ Defining sessions
...
-These are **not**::
+When collecting your sessions, Nox will create a separate session for each interpreter. You can see these sesions when running ``nox --list-sessions``. For example this Noxfile::
 
-    def some_session(session):
+    @nox.session(python=['2.7', '3.5', 3.6', '3.7'])
+    def tests(session):
         pass
 
-    def other_func(session):
+Will produce these sessions::
```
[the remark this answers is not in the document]
theacodes: Yep, this still works! I only apply the suffix to the *signature* (or long name as manifest.py calls it), so `nox -s tests` and ofc `nox -k tests` will run all three.

docs/config.rst
theacodes: 1. That compliant will return if we do not add the feature in (2).
2. I plan on mirroring py.test's approach for [parametrized ids](https://docs.pytest.org/en/latest/example/parametrize.html#different-options-for-test-ids). Issue filed for it at #96.

theacodes:
Okay, @lukesneeringer, I think that's everything? I'm going to go ahead and merge so I can send some simple follow-ups, but please let me know if I missed anything or if you have further concerns and I'll address them.

Thanks again!

lukesneeringer:
> Okay, @lukesneeringer, I think that's everything? I'm going to go ahead and merge so I can send some simple follow-ups, but please let me know if I missed anything or if you have further concerns and I'll address them.
> 
> Thanks again!

It was good. :-) I was just waiting for a confirmation on `-s` before I gave you a green checkmark.

pradyunsg:
This makes me super happy as all of the hacks that I had for nox's previous behavior are no longer needed: https://github.com/pradyunsg/zazo/commit/6abf4a5a9b1579027d1d85f332bbeb4be9b1ecb1

🎉

pradyunsg:
Does this mean it's no longer possible to use nox to test Py2.7/Py3.4 libraries?

theacodes:
> Does this mean it's no longer possible to use nox to test Py2.7/Py3.4 libraries?

Nope, you still can, it just means that you have to install nox itself into a python 3.5+ environment.

pradyunsg:
Hmm... That's gonna make it a little difficult to deal with Travis. Let's see how it can be done.

Issue #60, opened 2017-09-26
Pass through the architecture on OS X


dhermes:
This allows `nox` to test on both 32-bit and 64-bit architectures.

With a Python script that is architecture aware ([`check.py`][1]) it's possible to write a [`nox` config][2] that sends along the "correct" `arch -32` prefix for commands. (Unfortunately prepending `arch -32` to a command doesn't show up in `sys.argv`.)

-----

Running it with my magic trick turned on does the right thing:

```
$ /Library/Frameworks/Python.framework/Versions/3.6/bin/nox -s 'check(do_hack=True)'
nox > Running session check(do_hack=True)
nox > /Library/Frameworks/Python.framework/Versions/3.6/bin/python -m virtualenv ...
nox > chdir ${HOME}/nox-bug-report
nox > arch -64 python check.py
Architecture: 64-bit
      Prefix: ${HOME}/nox-bug-report/.nox/check-do_hack-true/bin/..
 Real prefix: /Library/Frameworks/Python.framework/Versions/3.6
nox > Session check(do_hack=True) was successful.
$
$ # Re-use
$ arch -32 \
>   /Library/Frameworks/Python.framework/Versions/3.6/bin/nox -s 'check(do_hack=True)' -r
nox > Running session check(do_hack=True)
nox > Re-using existing virtualenv at ${HOME}/nox-bug-report/.nox/check-do_hack-true.
nox > chdir ${HOME}/nox-bug-report
nox > arch -32 python check.py
Architecture: 32-bit
      Prefix: ${HOME}/nox-bug-report/.nox/check-do_hack-true/bin/..
 Real prefix: /Library/Frameworks/Python.framework/Versions/3.6
nox > Session check(do_hack=True) was successful.
```

Running it in the default mode for `nox` doesn't pass through the architecture

```
$ /Library/Frameworks/Python.framework/Versions/3.6/bin/nox -s 'check(do_hack=False)'
nox > Running session check(do_hack=False)
nox > /Library/Frameworks/Python.framework/Versions/3.6/bin/python -m virtualenv ...
nox > chdir ${HOME}/nox-bug-report
nox > python check.py
Architecture: 64-bit
      Prefix: ${HOME}/nox-bug-report/.nox/check-do_hack-false/bin/..
 Real prefix: /Library/Frameworks/Python.framework/Versions/3.6
nox > Session check(do_hack=False) was successful.
$
$ # Re-use
$ arch -32 \
[...]

PR #100, opened 2018-08-22
Adding support for 32-bit binaries on Windows.

dhermes:
An alternative implementation (and possibly a better one) would be to allow `arch` as an optional argument to `nox.session`. For example:

```python
@nox.session(python=["2.7", "3.7"], arch=["32", "64"])
def tests(session):
    session.install("pytest")
    session.install("-r", "requirements-dev.txt")
    session.install("-e", ".")
    session.run("py.test")
```

This would also give a "solution" to #60.

The real silly thing: I'm not sure anyone cares about 32-bit support. I used to test my OS X binaries for `bezier` in 32-bit, but then NumPy discontinued 32-bit support on OS X. Maybe that's a different beast since Apple controls the hardware?

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -125,6 +125,12 @@ Nox has provisional support for running on Windows. However, depending on your W
 * `Python issue 24493 <http://bugs.python.org/issue24493>`_
 * `Virtualenv issue 774 <https://github.com/pypa/virtualenv/issues/774>`_
 
+The Python binaries on Windows are found via the Python `Launcher`_ for
+Windows (``py``). For example, Python 3.5 can be found by determining which
+executable is invoked by ``py -3.5``. If a given test needs to use the 32-bit
+version of a given Python, then ``X.Y-32`` should be used as the version.
+
+.. _Launcher: https://docs.python.org/3/using/windows.html#python-launcher-for-windows
 
 Converting from tox
 -------------------
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -52,7 +52,7 @@ class ProcessEnv:
 
 
 def locate_via_py(version):
-    """Find the Python executable using the Windows launcher.
+    """Find the Python executable using the Windows Launcher.
 
     This is based on :pep:397 which details that executing
     ``py.exe -{version}`` should execute python with the requested
@@ -61,7 +61,9 @@ def locate_via_py(version):
     specific Python interpreter.
 
     Args:
-        version (str): The desired Python version.
+        version (str): The desired Python version to pass to ``py.exe``. Of the form
+            ``X.Y`` or ``X.Y-32``. For example, a usage of the Windows Launcher might
+            be ``py -3.6-32``.
 
     Returns:
         Optional[str]: The full executable path for the Python ``version``,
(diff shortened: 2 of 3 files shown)


dhermes:
PS Sorry for sending a PR without a corresponding issue. Consider this PR a point for discussion as much as anything. I.e. let's close it if it seems like a bad idea.

theacodes:
I feel like Nox should still support testing against these interpreters.

How well does Tox deal with this issue?

dhermes:
I don't see any mention of it in `tox` docs, so it's unclear. (I am a bit [scared][1] to boot to Windows on my local machine, I could spin up AppVeyor to find out what `tox does?)

[1]: https://twitter.com/bossylobster/status/1031779928809721856

theacodes:
If you don't mind doing the research.

dhermes:
So AFAICT, the recommendation is to use `tox -e py`, which just defaults to the Python version where `tox` is installed. From the [packaging guide][2]:

> In order to run tests using `tox`, therefore, projects should probably
> use the default `py` environment in `tox`, which uses the Python
> interpreter that was used to run `tox`. This will ensure that when
> AppVeyor runs the tests, they will be run with the configured interpreter.

I ran a [simple][1] config and confirmed this works. From the [`tox` source][3] it seems clear there is no support for declaring 32-bit or 64-bit.

[1]: https://ci.appveyor.com/project/dhermes/bezier/build/1.0.1028.tox-wat
[2]: https://packaging.python.org/guides/supporting-windows-using-appveyor/#testing-with-tox
[3]: https://github.com/tox-dev/tox/blob/657b18e2d35ef58ac34be53442491c4f0a62efed/src/tox/interpreters.py#L165

theacodes:
Okay, looking at this implementation I'm not sure how I feel about the environment variable functioning as an extra argument here.

Could we do something like:

```python
@nox.session(python=['3.6', '3.6-32'])
````

**Comment by dhermes** [comment]
I also don't feel great about using an environment variable. I used `PYTHON_ARCH` because those before us have done this for AppVeyor scripts.

There are only two architectures, but you'd still end up with a doubling of options (versus just adding a two-element list). For example:

```python
@nox.session(python=["2.7", "2.7-32", "3.5", "3.5-32", "3.6", "3.6-32", "3.7", "3.7-32"])
def foo(session):
    ...


@nox.session(python=["2.7", "3.5", "3.6", "3.7"], arch=["32", "64"])
def foo(session):
    ...
```

theacodes:
Hrm, the section option doesn't let folks do things like "only test 32 bit python 3.7".

@lukesneeringer and I discussed a syntax like this previously:

```python
@nox.session(python=['3.6', nox.python('3.6', arch='32')])
def tests(s):
    pass
```

WDYT?

dhermes:
My **initial** reaction was that I liked it. Then I thought about the implications a bit and liked it less, mostly because it makes the implementation more complex since it has to support "strings, AND ..." as the expected type. I think your suggestion of `3.6-32` is better and it has precedence: `py -3.6-32` is the preferred way with the Python Windows Launcher.

As for your 

> the ~~section~~ second option doesn't let folks do things like "only test 32 bit python 3.7".

I don't think it's actually a reason not to do it. `arch` is largely irrelevant on Linux and probably on OS X for most users. So the "extra" environments containing `arch` would only ever **need** to be run on Windows (or a Windows CI). (I realize this breaks the bare `nox`.)

theacodes:
Hrm, so which direction are you leaning closer to now?

theacodes:
Okay, so what would be the behavior for those things on Linux/Mac?

dhermes:
I think on Linux it'd just be a `skip`? Or some text UI version of a `NotImplementedError`?

On Mac it's possible, but there are a few sharp corners, but I'm happy to write code / give advice.

theacodes:
Cool, I'm happy with that as long as we have well defined cases for that.

theacodes:
@dhermes is this ready? It seems to still be failing on Travis.

PR #104, opened 2018-08-22
Adding `functools.wraps` to `generate_calls()`.

### LINKED ISSUE #103: nox.session and nox.parameterize no longer play nice together
With `2018.7.31.dev1`, the following "legacy" config

```python
import nox


@nox.session
@nox.parametrize("py", ["2.7", "3.7"])
def now(session, py):
    session.interpreter = "python{}".format(py)
    session.install("pytest")
    session.run("py.test")
```

fails with a hard-to-grok (I'm still source-diving) error:

```
$ nox -s "now(py='2.7')" -r
nox > Running session now(py='2.7')
nox > Session now(py='2.7') raised exception AttributeError("'function' object has no attribute 'python'")
  File ".../python3.7/site-packages/nox/sessions.py", line 283, in execute
    self._create_venv()
  File ".../python3.7/site-packages/nox/sessions.py", line 256, in _create_venv
    if self.func.python is False:
AttributeError: 'function' object has no attribute 'python'
nox > Session now(py='2.7') failed.
```

dhermes:
Fixes #103.

----

For now, this has no tests, simply because I wanted to get it out for discussion with @theacodes. I am in the process of writing a few tests right now.


dhermes:
@theacodes I'm struggling here with the tests. I made an overly-specific update to `test__parametrize.py` but I think there needs to be an end-to-end test in place. Can you do the honor of deciding where it goes and how to run it? I'm happy to supply cases to be placed in there.

theacodes:
> Can you do the honor of deciding where it goes and how to run it?

Maybe makes sense for manifest to test the two combinations?

dhermes:
ISTM that `test_manifest.py` is too "unit testy", i.e. it invokes methods on `Manifest` directly, rather than running the whole system.

theacodes:
Eh, I'm fine with mixing "unit" tests and "integration" tests, as I really don't (personally) make a distinction.

dhermes:
@theacodes PTAL. I went a little wild and did some editorializing (that I maybe should put in a new PR):

- Added a LICENSE header to `tests/resources/noxfile.py`
- Removed your `nox._testing.Namespace` class and just replaced with `argparse.Namespace`.

Rather than putting the end-to-end test in `test_manifest.py` I put it in `test_main.py`. It (right now) has way too many magic strings in it, so we should discuss how to make the test less dependent on the contents of the source file. I **could** just dump contents in a tempfile, but I actually like the idea of having a sample repository of configs you could point users to.

tests/resources/noxfile_nested.py
```diff
@@ -12,8 +12,14 @@
...
 
+import nox
 
-class Namespace:
-    """A class that makes the provided kwargs available as attributes."""
-    def __init__(self, **kwargs):
-        self.__dict__.update(kwargs)
+
+def eat(cheese):
+    print("Noms, {} so good!".format(cheese))
```
theacodes: if you're just printing, you don't need this indirection.

theacodes:
Looks fine, just one little comment. Definitely appreciate the cleanup work.

tests/resources/noxfile_nested.py
dhermes: Yeah I saw that such calls were considered [legacy][1] now. I was just looking for an excuse to use the `session` object. Any better ideas?

[1]: https://github.com/theacodes/nox/blob/ad99797b630ca7abeb6bdf115ad0b4d788c30d26/nox/sessions.py#L158-L160

theacodes: Eh, just leave it unused?

theacodes:
Go for it! Thank you! :)

PR #105, opened 2018-08-22
Using `os.pathsep` instead of a hardcoded `':'`.

dhermes:
@theacodes Do I need any tests for this?

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -47,7 +47,7 @@ class ProcessEnv:
 
         if self.bin:
             self.env['PATH'] = os.pathsep.join(
-                [self.bin, self.env.get('PATH')])
+                [self.bin, self.env.get('PATH', '')])
 
     @property
     def bin(self):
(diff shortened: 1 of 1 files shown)


nox/virtualenv.py
```diff
@@ -46,7 +46,8 @@ def __init__(self, bin=None, env=None):
             self.env.pop(key, None)
 
         if self.bin:
-            self.env['PATH'] = ':'.join([self.bin, self.env.get('PATH')])
+            self.env['PATH'] = os.pathsep.join(
+                [self.bin, self.env.get('PATH')])
```
dhermes: @theacodes Do you want this to blow up if `PATH` isn't in the environment variables? `join()` **will** blow up (`TypeError`) if `None` is returned. Is a `KeyError` from `self.env['PATH']` preferable?

theacodes: Hrm PATH being unset is probably a pretty bizarre scenario, but according to [this](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html#tag_08_03):

> If PATH is unset or is set to null, the path search is implementation-defined.

So maybe we can just treat it like an empty string if it's not present?

dhermes: Done.

PR #119, opened 2018-08-24
Using more specific regex in `_resolved_interpreter()`.

dhermes:


nox/virtualenv.py
```diff
@@ -108,13 +108,9 @@ def _resolved_interpreter(self):
...
         # part and add Python to the front of it.
-        match = re.match(r"^([\d\.]+?)$", self.interpreter)
+        match = re.match(r"^(?P<xy_ver>\d(\.\d)?)(\.\d+)?$", self.interpreter)
         if match:
-            parts = match.group(1).split(".")
-            if len(parts) > 2:
-                parts = parts[:2]
-
-            self.interpreter = "python{}".format(".".join(parts))
+            self.interpreter = "python{}".format(match.group("xy_ver"))
```
dhermes: @theacodes I can save

```python
xy_version = match.group("xy_ver")
```

and then re-use it in the call to `locate_via_py` at line 131. Does that make the code too hard to read?

dhermes: OK turns out this is relevant for #100, so I'll just address it there and use this PR as a diffbase

PR #124, opened 2018-08-27
Storing `platform.system()` as global in `nox.virtualenv`.

dhermes:
Also re-factoring unit tests to account for this change. This is a follow up to #100.

Also some minor editorializing (switch usage of `dir` to `dir_` as a variable name, shadowing builtins grumble grumble).


theacodes:
Woooo

PR #127, opened 2018-08-28
Don't modify `Virtualenv.interpreter` in `_resolved_interpreter`.

dhermes:
Also cache and re-use the result.


dhermes:
Not very controversial, eh @theacodes?

theacodes:
Nah I think the value of `intepreter` should be the same you specify to `nox.session`.

PR #128, opened 2018-08-28
Adding `Session.__slots__`.

dhermes:
This way old code being migrated will get failures when trying to set previously supported attributes such as `session.interpreter` and `session.virtualenv_dirname`.

----

Note that `TestSession.test_install` is currently broken because `Session.run` cannot be monkey-patched in the presence of `__slots__`. I have three workarounds for this in mind but would like to discuss:

1. Define the behavior of `Session.run` in a function `_run` defined in the `nox.sessions` module and the monkey-patch that function
1. Instantiate a do-nothing subclass of `Session` that does not have `__slots__`, then just monkey-patch `run` on that instance
1. Ditch `__slots__` and use `__(get|set|del)attr__` to give helpful error messages for non-existent attributes

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -74,8 +74,20 @@ class Session:
     your Nox session.
     """
 
+    __slots__ = ("_runner",)
+
     def __init__(self, runner):
         self._runner = runner
+
+    @property
+    def __dict__(self):
+        """Attribute dictionary for object inspection.
+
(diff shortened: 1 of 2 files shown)


theacodes:
(2) sounds pretty reasonable.

dhermes:
@theacodes PTAL, I went with (2) as commanded

theacodes:
> as commanded

don't give me that much power.

PR #135, opened 2018-09-17
Addition of run_by_default flag and nox.param()

### LINKED ISSUE #40: Add a default session list
Invoking `nox` with no command line arguments runs every session.  It would be useful if the nox.py could specify a default list of sessions to run in that case, for example

```
nox.default_sessions([ 'lint', 'tests' ])
```

SethMichaelLarson:
Here's my initial implementation of allowing the following:

```python
import nox

@nox.session(run_by_default=True)
def test_default(session):
    pass

@nox.session
@nox.parametrize(["django", "param"], [("1.9", 1), nox.param("2.0", 2, run_by_default=True))
def test_django_versions(session, django, param):
    pass
```

Also adds the "reverse" of `run_by_default=True` by allowing the following:

```python
@nox.session
def is_default(session):
    pass

@nox.session(run_by_default=False)
def is_not_default(session):
    pass
```

Basically the logic is as follows:
1. If there's any sessions marked with `run_by_default=True`, run only those sessions.
2. Else if there's any sessions marked with `run_by_default=False`, then run all unmarked sessions. 
3. Otherwise run all sessions.

This implementation is following some suggestions within #40 but doesn't incorporate the `nox.options...` suggestion I've seen in an issue elsewhere.

Will require additional test cases and updates to documentation. I've tested it locally by modifying the existing `noxfile.py`. I was just wondering if this was the solution that looks good to everyone, it seemed a little heavy-handed when I was writing it so I'm open to suggestions! :)

Fixes #40

### DIFF
--- a/nox/__init__.py
+++ b/nox/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from nox._parametrize import parametrize_decorator as parametrize
+from nox._parametrize import param, parametrize_decorator as parametrize
 from nox.registry import session_decorator as session
 
-__all__ = ["parametrize", "session"]
+__all__ = ["param", "parametrize", "session"]
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -68,7 +68,7 @@ def main():
         "-e",
         "--sessions",
         nargs="*",
-        default=_get_default_sessions(),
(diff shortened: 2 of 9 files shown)


theacodes:
Thanks for doing this @SethMichaelLarson, I think I'm going to close this in favor of a more general-purpose solution proposed in #141. Please take a look at let me know what you think?

Issue #93, opened 2018-07-22
When a Python version isn't installed, nox fails (without going on to the next session)


pfmoore:
I have a `nox.py` file for testing my project, and I want to specify all of the Python versions I support. That's easy enough, and works fine for CI. But when running nox on my local PC, I don't have all those Python versions installed. As a result, I get `RuntimeError: Unable to locate Python interpreter "python3.5".` when I run plain `nox`, and the run stops there.

Two things would be a huge improvement:

1. Respect the `--stop-on-first-error` setting, and by default skip sessions that can't be created because the Python interpreter isn't installed.
2. Avoid the ugly traceback, and simply print an error "Python interpreter XXX not installed - skipping".

Manually specifying just the environments I want to run isn't really practical (apart from anything else, the syntax for specifying a parametrised environment is much more verbose than tox's `-e py36`).

theacodes:
This is a great idea, and should be pretty easy to do. I'm cool with both 1 and 2.

theacodes:
Seems that making Nox imperative fixed this. :)

What it looks like now:

<img width="1132" alt="screen shot 2018-08-23 at 8 00 13 pm" src="https://user-images.githubusercontent.com/250995/44562624-3af0ae80-a70f-11e8-87df-a325c03f6908.png">

Let me know if there's more I can do here, @pfmoore!

PR #138, opened 2018-09-19
If `X.Y-32` is used on non-Windows, just skip.

### LINKED ISSUE #136: Skip sessions that want to use 32-bit Python on non-Windows
The current experience is not a good one:

```
$ nox -s unit-3.5-32 -r
nox > Running session unit-3.5-32
nox > Session unit-3.5-32 raised exception RuntimeError("Locating 32-bit Python ('3.5-32') is only supported on Windows.")
  File ".../.pyenv/versions/3.7.0/lib/python3.7/site-packages/nox/sessions.py", line 295, in execute
    self._create_venv()
  File ".../.pyenv/versions/3.7.0/lib/python3.7/site-packages/nox/sessions.py", line 281, in _create_venv
    self.venv.create()
  File ".../.pyenv/versions/3.7.0/lib/python3.7/site-packages/nox/virtualenv.py", line 209, in create
    cmd.extend(["-p", self._resolved_interpreter])
  File ".../.pyenv/versions/3.7.0/lib/python3.7/site-packages/nox/virtualenv.py", line 165, in _resolved_interpreter
    raise self._resolved
RuntimeError: Locating 32-bit Python ('3.5-32') is only supported on Windows.
nox > Session unit-3.5-32 failed.
```

dhermes:
Fixes #136.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -122,9 +122,11 @@ class VirtualEnv(ProcessEnv):
 
         Based heavily on tox's implementation (tox/interpreters.py).
         """
+        from nox.sessions import _SessionSkip
+
         # If there is no assigned interpreter, then use the same one used by
         # Nox.
-        if isinstance(self._resolved, RuntimeError):
+        if isinstance(self._resolved, Exception):
             raise self._resolved
 
         if self._resolved is not None:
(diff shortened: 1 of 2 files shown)


nox/virtualenv.py
```diff
@@ -122,9 +122,11 @@ def _resolved_interpreter(self):
 
         Based heavily on tox's implementation (tox/interpreters.py).
         """
+        from nox.sessions import _SessionSkip
```
dhermes: @theacodes I don't like this. It was to avoid a circular import since `nox.sessions` imports `nox.virtualenv`. What do you think we should do?

theacodes: Why not just raise a new `InterpreterNotFound` error? We could use that as a more general purpose solution to #93.

dhermes:
I expected some tests to fail that verified the contents of the logged error message (which were previously in `RuntimeError().args`). However, there was no testing of this content. WDYT @theacodes?

nox/virtualenv.py
dhermes: Care to take the wheel here? I'm not really sure where to put it / what other corners to test it in.

dhermes:
For posterity, this was closed in favor of #140

PR #140, opened 2018-09-21
Skip sessions with missing interpreters

### LINKED ISSUE #136: Skip sessions that want to use 32-bit Python on non-Windows

theacodes:
Note: This changes Nox's behavior. Previously, missing interpreters would cause a failed session. Now they just cause a warning. The previous behavior can be used via ``--error-on-missing-interpreters``.

Replaces #138 
Closes #136 
Releated #93

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -104,6 +104,14 @@ By default nox will continue to run all sessions even if one fails. You can use 
     nox --stop-on-first-error
 
 
+Failing sessions when the interpreter is missing
+------------------------------------------------
+
+By default, Nox will skip sessions where the Python interpreter can't be found. If you want Nox to mark these sessions as failed, you can use ``--error-on-missing-interpreters``::
+
+    nox --error-on-missing-interpreters
+
+
 Controlling color output
 ------------------------
 
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -357,7 +357,7 @@ class Result:
             return "was successful"
         status = self.status.name.lower()
         if self.reason:
-            return f"{status}: {self.reason}"
+            return "{}: {}".format(status, self.reason)
         else:
             return status
 
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -239,15 +239,11 @@ class Session:
 
     def error(self, *args, **kwargs):
         """Immediately aborts the session and optionally logs an error."""
-        if args or kwargs:
-            logger.error(*args, **kwargs)
-        raise _SessionQuit()
+        raise _SessionQuit(*args, **kwargs)
 
     def skip(self, *args, **kwargs):
         """Immediately skips the session and optionally logs a warning."""
-        if args or kwargs:
-            logger.warning(*args, **kwargs)
-        raise _SessionSkip()
+        raise _SessionSkip(*args, **kwargs)
 
 
 class SessionRunner:
@@ -304,11 +300,17 @@ class SessionRunner:
             # Nothing went wrong; return a success.
             return Result(self, Status.SUCCESS)
 
-        except _SessionQuit:
-            return Result(self, Status.ABORTED)
+        except nox.virtualenv.InterpreterNotFound as exc:
+            if self.global_config.error_on_missing_interpreters:
+                return Result(self, Status.FAILED, reason=str(exc))
+            else:
+                return Result(self, Status.SKIPPED, reason=str(exc))
 
-        except _SessionSkip:
-            return Result(self, Status.SKIPPED)
+        except _SessionQuit as exc:
+            return Result(self, Status.ABORTED, reason=str(exc))
+
+        except _SessionSkip as exc:
+            return Result(self, Status.SKIPPED, reason=str(exc))
 
         except nox.command.CommandFailed:
             return Result(self, Status.FAILED)
@@ -325,16 +327,18 @@ class SessionRunner:
 class Result:
     """An object representing the result of a session."""
 
-    def __init__(self, session, status):
+    def __init__(self, session, status, reason=None):
         """Initialize the Result object.
 
         Args:
             session (~nox.sessions.SessionRunner):
                 The session runner which ran.
             status (~nox.sessions.Status): The final result status.
+            reason (str): Additional info.
         """
         self.session = session
         self.status = status
+        self.reason = reason
 
     def __bool__(self):
         return self.status.value > 0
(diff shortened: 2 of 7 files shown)


docs/usage.rst
```diff
@@ -104,6 +104,14 @@ By default nox will continue to run all sessions even if one fails. You can use
     nox --stop-on-first-error
 
 
+Failing sessions when the interpreter is missing
+------------------------------------------------
+
+By default, Nox will skip sessions where the Python interpreter can't be found. If you want Nox to mark these sessions as failed, you can use ``--error-on-missing-interpreters``::
+
+    nox --error-on-missing-interpreters
```
dhermes: Probably worth having a "primary" flags group so flags like this one can appear below the fold in `python -m nox --help`?

nox/sessions.py
```diff
@@ -239,15 +239,11 @@ def log(self, *args, **kwargs):
 
     def error(self, *args, **kwargs):
         """Immediately aborts the session and optionally logs an error."""
-        if args or kwargs:
-            logger.error(*args, **kwargs)
-        raise _SessionQuit()
+        raise _SessionQuit(*args, **kwargs)
```
dhermes: :+1: on moving the logging into the exception classes

nox/sessions.py
```diff
@@ -325,16 +327,18 @@ def execute(self):
...
 
-    def __init__(self, session, status):
+    def __init__(self, session, status, reason=None):
         """Initialize the Result object.
 
         Args:
             session (~nox.sessions.SessionRunner):
                 The session runner which ran.
             status (~nox.sessions.Status): The final result status.
+            reason (str): Additional info.
```
dhermes: Pedantry alert: `Optional[str]`

nox/sessions.py
```diff
@@ -351,7 +355,11 @@ def imperfect(self):
         """
         if self.status == Status.SUCCESS:
             return "was successful"
-        return self.status.name.lower()
+        status = self.status.name.lower()
+        if self.reason:
```
dhermes: nunchucks-nuns-chuck-norris-nun-12593488599

nox/virtualenv.py
```diff
@@ -31,6 +31,12 @@
 _SYSTEM = platform.system()
 
 
+class InterpreterNotFound(EnvironmentError):
```
dhermes:
```
$ python2.7 -c 'print(EnvironmentError)'
<type 'exceptions.EnvironmentError'>
$ python3.7 -c 'print(EnvironmentError)'
<class 'OSError'>
```

nox/virtualenv.py
```diff
@@ -134,59 +140,46 @@ def _resolved_interpreter(self):
...
-        # or may be a full path to an interpreter in the filesystem; accept this.
+        # If the cleaned interpreter is on the PATH, go ahead and return it.
         if py.path.local.sysfind(cleaned_interpreter):
             self._resolved = cleaned_interpreter
             return self._resolved
 
+        # The rest of this is only applicable to Windows, so if we don't have
+        # an interpreter by now, raise.
+        if _SYSTEM != "Windows":
+            self._resolved = InterpreterNotFound(self.interpreter)
```
dhermes: :+1:

nox/virtualenv.py
```diff
@@ -134,59 +140,46 @@ def _resolved_interpreter(self):
...
+            raise self._resolved
+
         # Allow versions of the form ``X.Y-32`` for Windows.
-        if not xy_version:
-            match = re.match(r"^\d\.\d-32?$", cleaned_interpreter)
-            if match:
-                xy_version = cleaned_interpreter
+        match = re.match(r"^\d\.\d-32?$", cleaned_interpreter)
```
dhermes: @theacodes It's possible that the user has already specified `python2.7` in which case calling `re.match` here is "wasted computation".

nox/virtualenv.py
```diff
@@ -134,59 +140,46 @@ def _resolved_interpreter(self):
...
+            # preserve the "-32" suffix, as the Python launcher expects
+            # it.
+            xy_version = cleaned_interpreter
 
-        # Sanity check: We only need the rest of this behavior on Windows.
-        if _SYSTEM != "Windows":
-            if xy_version.endswith("-32"):
-                self._resolved = RuntimeError(
-                    "Locating 32-bit Python ({!r}) is "
-                    "only supported on Windows.".format(cleaned_interpreter)
```
dhermes: Was this error message not more informative than the proposed behavior on line 164?

docs/usage.rst
theacodes: Maybe, filed #142.

nox/sessions.py
theacodes: Parameters that have a default value of `None` are implicitly `Optional`.

nox/sessions.py
theacodes: Intentional, it can be an empty string.

nox/virtualenv.py
theacodes: Fixed, good catch. (OSError is fine, but let's use the non-deprecated name).

nox/virtualenv.py
theacodes: It's fine.

nox/virtualenv.py
theacodes: Yes, but this branch has been removed, we no longer check for `32` until after we early-out because of Windows.

PR #147, opened 2018-10-12
Warn when programs not in the virtualenv are used, allow erroring and silencing the warning.

### LINKED ISSUE #130: `nox` uses `PATH` to find executables
This is a bit `tox`-like to suggest, so feel free to reject.

I was recently migrating my Pelican blog from a `Makefile` to a `noxfile`. I had `pelican` installed in a few of my base environments, but forgot to install it into the virtual environment. However, the sessions ran just fine because they referenced the `pelican` executable on my `${PATH}` even though none existed in the `virtualenv`.

theacodes:
Add --warn-on-external-run flag and the "external" keyword arg to session.run.

Closes #130

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -66,7 +66,16 @@ def _clean_env(env):
     return clean_env
 
 
-def run(args, *, env=None, silent=False, path=None, success_codes=None, log=True):
+def run(
+    args,
+    *,
+    env=None,
+    silent=False,
+    path=None,
+    success_codes=None,
+    log=True,
+    external=False
+):
     """Run a command-line program."""
 
     if success_codes is None:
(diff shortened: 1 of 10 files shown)


theacodes:
@dhermes you don't have to review in you don't want to. This one was pretty trivial, but happy to hear any input you have.

nox/command.py
```diff
@@ -80,6 +89,24 @@ def run(args, *, env=None, silent=False, path=None, success_codes=None, log=True
     if log:
         logger.info(full_cmd)
 
+        is_external_tool = path is not None and not cmd_path.startswith(path)
+        if is_external_tool:
+            if external == "error":
+                logger.error(
+                    "Error: {} is not installed into the virtualenv, it located at {}. "
```
dhermes: it located -> it is located

theacodes: Good catch, 1646e52.

PR #152, opened 2018-10-17
Run sessions in the same order specified on the command line.

### LINKED ISSUE #150: Nox is no longer executing sessions in the order specified on the command line

theacodes:
Closes #150

### DIFF
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -118,12 +118,15 @@ class Manifest:
         """
         # Filter the sessions remaining in the queue based on
         # whether they are individually specified.
-        specified_sessions = set(specified_sessions)
-        self._queue = [
-            x
-            for x in self._queue
-            if (x.name in specified_sessions or set(x.signatures) & specified_sessions)
-        ]
+        queue = []
+        for session_name in specified_sessions:
+            for session in self._queue:
+                if session_name == session.name or session_name in set(
+                    session.signatures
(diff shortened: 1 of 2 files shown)


nox/manifest.py
```diff
@@ -118,12 +118,15 @@ def filter_by_name(self, specified_sessions):
...
         # Filter the sessions remaining in the queue based on
         # whether they are individually specified.
-        specified_sessions = set(specified_sessions)
-        self._queue = [
-            x
-            for x in self._queue
-            if (x.name in specified_sessions or set(x.signatures) & specified_sessions)
-        ]
+        queue = []
+        for session_name in specified_sessions:
```
dhermes: Don't you still want a `set()` for `specified_sessions`?

You could instead do an "ordered set":

```python
>>> import collections
>>> specified_sessions = ["a", "c", "a", "c", "a", "b"]
>>> o_set = collections.OrderedDict((val, i) for i, val in enumerate(specified_sessions))
>>> list(o_set.keys())
['a', 'c', 'b']
>>> "d" in o_set
False
>>> "a" in o_set
True
```

theacodes: Eh, I'm not overly concerned. If someone really wants to specify a session twice we'll run it twice.

dhermes: Ha OK.

Issue #143, opened 2018-10-05
Add support for breakpointing


theacodes:
It would be useful in some cases to "drop in to" a Nox session - that is, obtain a shell that has the same `PATH` and environment variables as would be available to `session.run`. We can achieve something like this today by running a session and then `source .nox/session-name/bin/activate` but it's cumbersome.

A naive implementation would resemble `pipenv`'s `shell` command. It would still require that a session have been run at least once.

A much much more fancy implementation would be to take advantage of something like [PEP 553](https://www.python.org/dev/peps/pep-0553/) and support `breakpoint()`. This allows users to define *where* in the session breakpointing can occur (say, after `install` or `build` steps but before `pytest`):

```
@nox.session()
def tests(session):
    session.install('requirements-dev.txt')
    breakpoint()
    session.run('pytest', 'tests')
```

dhermes:
Such a cool idea :grin:

theacodes:
Pipenv's implementation is probably a good starting point: https://github.com/pypa/pipenv/blob/master/pipenv/shells.py

Tolker-KU:
This seems like a very cool feature. Any thoughts on going ahead with it?

PR #153, opened 2018-11-02
Add --install-only flag to install dependencies without running anything

bmccutchon:
The main use is for downloading dependencies for later offline use (with `--reuse-existing-virtualenvs`). A typical session where this would be useful will have the form:

```python
@nox.session
def tests(session):
    session.install(...)
    session.run("spam", ...)
```

### DIFF
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -54,8 +54,8 @@ class GlobalConfig:
         self.no_error_on_missing_interpreters = args.no_error_on_missing_interpreters
         self.error_on_external_run = args.error_on_external_run
         self.no_error_on_external_run = args.no_error_on_external_run
-        self.skip_pytest = args.skip_pytest
-        self.no_skip_pytest = args.no_skip_pytest
+        self.install_only = args.install_only
+        self.no_install_only = args.no_install_only
         self.posargs = args.posargs
         self.report = args.report
 
@@ -98,8 +98,8 @@ class GlobalConfig:
             options.error_on_external_run,
             self.no_error_on_external_run,
         )
-        self.skip_pytest = _default_with_off_flag(
-            self.skip_pytest, options.skip_pytest, self.no_skip_pytest
+        self.install_only = _default_with_off_flag(
+            self.install_only, options.install_only, self.no_install_only
         )
         self.report = self.report or options.report
 
@@ -211,14 +211,14 @@ def main():
     )
 
     secondary.add_argument(
-        "--skip-pytest",
+        "--install-only",
         action="store_true",
-        help="Skip invocations of the form session.run('py.test', ...) in the noxfile.",
+        help="Skip session.run invocations in the noxfile.",
     )
     secondary.add_argument(
-        "--no-skip-pytest",
+        "--no-install-only",
         action="store_true",
-        help="Disables --skip-pytest if it is enabled in the Noxfile.",
+        help="Disables --install-only if it is enabled in the Noxfile.",
     )
 
     secondary.add_argument(
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -98,9 +98,6 @@ class GlobalConfig:
             options.error_on_external_run,
             self.no_error_on_external_run,
         )
-        self.install_only = _default_with_off_flag(
-            self.install_only, options.install_only, self.no_install_only
-        )
         self.report = self.report or options.report
 
 
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -55,7 +55,6 @@ class GlobalConfig:
         self.error_on_external_run = args.error_on_external_run
         self.no_error_on_external_run = args.no_error_on_external_run
         self.install_only = args.install_only
-        self.no_install_only = args.no_install_only
         self.posargs = args.posargs
         self.report = args.report
 
@@ -211,11 +210,6 @@ def main():
         "--install-only",
         action="store_true",
         help="Skip session.run invocations in the Noxfile.",
-    )
-    secondary.add_argument(
-        "--no-install-only",
-        action="store_true",
-        help="Disables --install-only if it is enabled in the Noxfile.",
     )
 
     secondary.add_argument(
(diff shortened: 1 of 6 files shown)


dhermes:
ISTM this could be enabled via `posargs` and some helpful documentation. I'm a bit worried about having "flag soup", i.e. too many flags makes it harder to understand the interface.

stsewd:
Here are the docs about an example using posargs https://nox.thea.codes/en/stable/tutorial.html?highlight=posargs#passing-arguments-into-sessions

bmccutchon:
I'm aware of the posargs solution. I'm of the opinion that this shouldn't require modifications to the noxfile to work.

theacodes:
I'm wondering about thoughts on two things:

1. Would breakpointing (#143) satisfy this usecase?
2. Would inverting this to `--install-only` be a more easy to swallow proposal?

bmccutchon:
It seems like breakpointing is an interactive thing, whereas I'd like this to also work in an automated test environment. If people prefer `--install-only`, I'm fine with that.

theacodes:
Curious as to @lukesneeringer @stsewd and @dhermes' thoughts on the new proposal.

dhermes:
I am :+1: on `--install-only` and don't think breakpointing makes much sense to solve this.

stsewd:
@theacodes `--install-only` looks like a good approach.

theacodes:
Cool, @bmccutchon we have consensus, do you mind updating this pull to implement `--install-only`?

bmccutchon:
Will do.

bmccutchon:
It seems like the `--install-only` flag would only skip `run` commands. In particular:

* I don't see much reason for it to skip `chdir`,
* Skipping `notify` seems bad, since the notified session could have more `install` directives,
* Skipping `log` seems undesirable and not very useful, and
* Skipping `error` or `skip` seems like a bad idea.

theacodes:
seems reasonable

bmccutchon:
I made the change.

stsewd:
Seems like something weird happened with the git history, there are some merged commits.

theacodes:
Yeah, you may need to rebase onto my master branch and force push.

bmccutchon:
The history should be fixed now.

nox/__main__.py
```diff
@@ -205,6 +210,17 @@ def main():
         help="Disables --error-on-external-run if it is enabled in the Noxfile.",
     )
 
+    secondary.add_argument(
+        "--install-only",
+        action="store_true",
+        help="Skip session.run invocations in the noxfile.",
```
stsewd: `in the noxfile` -> `in the Noxfile`

nox/sessions.py
```diff
@@ -177,6 +177,10 @@ def run(self, *args, env=None, **kwargs):
         if not args:
             raise ValueError("At least one argument required to run().")
 
+        if self._runner.global_config.install_only:
+            logger.info("Skipping {} run, as --install-only is set.".format(args[0]))
+            return
```
stsewd: Maybe is better to raise a custom exception? 

Something like https://github.com/theacodes/nox/blob/c05f131c3b7255bde09cf597c93fc97d0f1ab5c2/nox/sessions.py#L59-L60

theacodes: Nah, this is fine.

nox/__main__.py
```diff
@@ -96,6 +98,9 @@ def merge_from_options(self, options):
             options.error_on_external_run,
             self.no_error_on_external_run,
         )
+        self.install_only = _default_with_off_flag(
+            self.install_only, options.install_only, self.no_install_only
```
theacodes: I don't forsee users wanting to set `install only` in their Noxfiles, so it's probably fine to make this a non-noxfile-changable-flag. We can always add it alter.

nox/__main__.py
bmccutchon: Done.

nox/__main__.py
bmccutchon: Done.

nox/sessions.py
bmccutchon: I think it's desirable to skip `run` commands rather than terminating when encountering them. That way, Nox will install all dependencies in a session of the form

```python
session.install('spam')
session.run('spam', ...)
session.install('eggs')
session.run('eggs', ...)
```

This will likely be more common when the second install/run pair is in another session invoked by the first with `notify`, as Nox's own noxfile does here:

https://github.com/theacodes/nox/blob/c05f131c3b7255bde09cf597c93fc97d0f1ab5c2/noxfile.py#L32

So, when I run `nox -s tests --install-only` in the root of this repo, "coverage" should be installed.

tests/test_sessions.py
```diff
@@ -112,6 +114,17 @@ def raise_value_error():
         with pytest.raises(nox.command.CommandFailed):
             assert session.run(raise_value_error)
 
+    @mock.patch.object(nox.command, "run")
```
theacodes: Can you add a test at asserts that `install` commands are still called?

nox/__main__.py
```diff
@@ -205,6 +207,17 @@ def main():
         help="Disables --error-on-external-run if it is enabled in the Noxfile.",
     )
 
+    secondary.add_argument(
+        "--install-only",
+        action="store_true",
+        help="Skip session.run invocations in the Noxfile.",
+    )
+    secondary.add_argument(
```
theacodes: You won't need this arg if we're not allowing Noxfile modification.

nox/sessions.py
```diff
@@ -177,6 +177,10 @@ def run(self, *args, env=None, **kwargs):
         if not args:
             raise ValueError("At least one argument required to run().")
 
+        if self._runner.global_config.install_only:
```
theacodes: since `install` just calls `run`, what keeps this from skipping install commands as well?

theacodes:
Thank you for doing this, @bmccutchon, and thanks for your thoughts and reviews, @stsewd and @dhermes!

Issue #156, opened 2018-11-04
Calling session.install() with keyword argument silent causes TypeError


erinwild-rubikloud:
For example, running `session.install("requests", "urllib3", silent=False)` results in the following error:
```
TypeError: run() got multiple values for keyword argument 'silent'
```

This is because `session.install()` calls `self.run("pip", "install", "--upgrade", *args, silent=True, external="error", **kwargs)` which doesn't allow for `silent` or `external` to be set in kwargs.

I'll be opening a pull request shortly which checks if `silent` is found in kwargs before attempting to set a default.

PR #157, opened 2018-11-04
Allow silent argument to be set in session.install

erinwild-rubikloud:
Addresses #156 by checking if `silent` is in the kwargs passed to `session.install()`. If not, it sets it to the default True.


theacodes:
Thanks, @erinwild-rubikloud!

PR #160, opened 2018-11-21
Fix tox casing to be consistent with their docs, remove . from pytest

theacodes:
Ref: https://github.com/tox-dev/tox/issues/1088

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,7 +10,7 @@ Welcome to Nox
    usage
    contrib
 
-``nox`` is a command-line tool that automates testing in multiple Python environments, similar to `Tox`_. Unlike Tox, Nox uses a standard Python file for configuration.
+``nox`` is a command-line tool that automates testing in multiple Python environments, similar to `tox`_. Unlike tox, Nox uses a standard Python file for configuration.
 
 Install nox via `pip`_::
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -39,9 +39,9 @@ For each session, Nox will automatically create `virtualenv`_ with the appropria
 
 To learn how to install and use Nox, see the :doc:`tutorial`. For documentation on configuring sessions, see :doc:`config`. For documentation on running ``nox``, see :doc:`usage`.
 
-.. _Tox: https://tox.readthedocs.org
+.. _tox: https://tox.readthedocs.org
```
dhermes: Sphinx / docutils / rST doesn't respect casing (not that it matters, and I also match casing)

theacodes: Interesting. :) Good to know, I'll do all of my links in TeEnAgEr CaSe from now on.

SethMichaelLarson: [mOcKiNg SpOnGeBoB cAsInG](https://i.kym-cdn.com/entries/icons/original/000/022/940/mockingspongebobbb.jpg)

obestwalter:
Thanks @theacodes!

PR #163, opened 2019-01-02
Don't colour output if `NO_COLOR` is set

srstevenson:
This follows the informal standard defined at https://no-color.org, allowing users who don't want colour output from command line applications to set a single environment variable to disable it in all supported applications.


theacodes:
Cool, thanks for doing this! I'll merge if the tests all pass.

srstevenson:
Thanks!

PR #168, opened 2019-02-18
Detect and handle when nox is frozen with PyInstaller

s0undt3ch:
This PR allows Nox to handle the case where it has been frozen by PyInstaller.

### Background
At first I tried looking into what would it take to make Nox py2 compatible, a couple of `SyntaxErrors` but it could be done, however, Nox has adopted Py3 like all projects should by now, so, it didn't make sense.
We were still left with a problem.
Out project is moving to Py3 only but we still need to support a Py2+Py3 code base, which would't be a problem if only we also didn't have to test against distributions which lack Python 3(installing python from source is not an option for our test scenarios), for example, CentOS 6.
The alternative here would be to freeze Nox into a single binary which we could then use to test on these systems using Nox.

There were a couple of rough edges than needed to be handled, hence, this PR.
Nox, correctly relies on `sys.executable` to call out the correct virtualenv module, but, in a single binary frozen scenario, that's not available, the interpreter is built-in into the frozen binary.
We had to teach Nox to search for a matching(freeze time) python binary, or as low as Python 2.7

#### Freeze Process
If it's of interest to the Nox project, I can add the required PyInstaller spec file and hooks which freezes Nox.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -203,11 +203,11 @@ class VirtualEnv(ProcessEnv):
         unless Nox was frozen with PyInstaller. In the later case, returns
         system python interpreter matching the interpreter used when Nox was
         frozen with PyInstaller or as low as Python 2.7, if found."""
-        if self._runtime is not None:
-            return self._runtime
-
         if isinstance(self._runtime, Exception):
             raise self._runtime
+
+        if self._runtime is not None:
+            return self._runtime
 
         if not _FROZEN:
             self._runtime = sys.executable
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -213,16 +213,34 @@ class VirtualEnv(ProcessEnv):
             self._runtime = sys.executable
             return self._runtime
 
+        def find_binary(major=None, minor=None):
+            if _SYSTEM != "Windows":
+                # Most common case first
+                interpreter = "python"
+                if major:
+                    interpreter += "{}".format(major)
+                if minor:
+                    interpreter += ".{}".format(minor)
+                resolved = py.path.local.sysfind(interpreter)
+                if resolved is not None and not isinstance(resolved, str):
+                    resolved = resolved.realpath().strpath
+                return resolved
+            else:
+                if not major and not minor:
+                    # py.exe needs major and minor both available
+                    return None
+                return locate_via_py("{}.{}".format(major, minor))
+
         # At this stage, we know Nox was frozen.
         # Let's first try a binary named python<major>.<minor>
-        interpreter = "python{}.{}".format(*sys.version_info)
-        if py.path.local.sysfind(interpreter):
+        interpreter = find_binary(sys.version_info.major, sys.version_info.minor)
+        if interpreter:
             self._runtime = interpreter
             return self._runtime
 
         # Let's now try a binary named python<major>
-        interpreter = "python{}".format(*sys.version_info)
-        if py.path.local.sysfind(interpreter):
+        interpreter = find_binary(sys.version_info.major)
+        if interpreter:
             self._runtime = interpreter
             return self._runtime
 
@@ -231,15 +249,16 @@ class VirtualEnv(ProcessEnv):
         # is apparently not the case.
         # Latest pip only works with Python 2.7+ so, no point to even try
         # to create a virtualenv under Python 2.6
-        interpreter = "python2.7"
-        if py.path.local.sysfind(interpreter):
+        interpreter = find_binary(major=2, minor=7)
+        if interpreter:
             self._runtime = interpreter
             return self._runtime
 
         # We're running out of options, let's see if we find a python2 or a
         # python binary, and if we do, let's check its version
-        for interpreter in ("python2", "python"):
-            if py.path.local.sysfind(interpreter):
+        for args in ((2,), ()):
+            interpreter = find_binary(*args)
+            if interpreter:
                 try:
                     output = nox.command.run(
                         [interpreter, "-V"], silent=True, log=False
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -29,6 +29,14 @@ _BLACKLISTED_ENV_VARS = frozenset(
     ["PIP_RESPECT_VIRTUALENV", "PIP_REQUIRE_VIRTUALENV", "__PYVENV_LAUNCHER__"]
 )
 _SYSTEM = platform.system()
+try:
+    # This attribute is set at runtime when running under a frozen
+    # pyinstaller binary
+    sys.frozen
+    _FROZEN = True
+except AttributeError:
+    # Catching an AttributeError is actually faster than (get|has)attr()
+    _FROZEN = False
 
 
 class InterpreterNotFound(OSError):
@@ -109,6 +117,7 @@ class VirtualEnv(ProcessEnv):
         self.location = os.path.abspath(location)
         self.interpreter = interpreter
         self._resolved = None
+        self._runtime = None
         self.reuse_existing = reuse_existing
         super(VirtualEnv, self).__init__()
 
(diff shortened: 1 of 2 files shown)


s0undt3ch:
Finally got CI passing.

Any input?

theacodes:
I'm busy with PyCascades this week, maybe @dhermes can review if he has time.

s0undt3ch:
:+1: K, not worries then.

nox/virtualenv.py
```diff
@@ -29,6 +29,14 @@
...
 )
 _SYSTEM = platform.system()
+try:
+    # This attribute is set at runtime when running under a frozen
+    # pyinstaller binary
+    sys.frozen
+    _FROZEN = True
+except AttributeError:
+    # Catching an AttributeError is actually faster than (get|has)attr()
+    _FROZEN = False
```
dhermes: 1.  What about instead using?
    ```python
    _FROZEN = getattr(sys, 'frozen', None) is not None
    ```
1.  ISTM that `_FROZEN` isn't specific enough. `_SYS_FROZEN` instead? Or `_SYS_HAS_FROZEN`?

nox/virtualenv.py
```diff
@@ -137,7 +146,12 @@ def _resolved_interpreter(self):
             return self._resolved
 
         if self.interpreter is None:
-            self._resolved = sys.executable
+            if not _FROZEN:
+                self._resolved = sys.executable
+                return self._resolved
+
+            # Nox was frozen with PyInstaller
```
dhermes: Since this is the "rare" case I'd prefer it to be in the "early-exit-for-a-special-case" block.

nox/virtualenv.py
```diff
@@ -182,6 +196,92 @@ def _resolved_interpreter(self):
...
+            raise self._runtime
+
+        if self._runtime is not None:
+            return self._runtime
+
+        if not _FROZEN:
+            self._runtime = sys.executable
+            return self._runtime
+
+        def find_binary(major=None, minor=None):
```
dhermes: Please move this out as a helper function (I am big :-1: on closures, they are hard to reason about and test and the enclosed scope is not really accessible)

nox/virtualenv.py
```diff
@@ -109,6 +117,7 @@ def __init__(self, location, interpreter=None, reuse_existing=False):
         self.location = os.path.abspath(location)
         self.interpreter = interpreter
         self._resolved = None
+        self._runtime = None
```
dhermes: What about `_runtime_interpreter_cached` instead?

nox/virtualenv.py
```diff
@@ -182,6 +196,92 @@ def _resolved_interpreter(self):
         self._resolved = InterpreterNotFound(self.interpreter)
         raise self._resolved
 
+    @property
+    def _runtime_interpreter(self):
+        """
+        Returns the interpreter to use when shelling out, ``sys.executable``,
+        unless Nox was frozen with PyInstaller. In the later case, returns
+        system python interpreter matching the interpreter used when Nox was
+        frozen with PyInstaller or as low as Python 2.7, if found."""
```
dhermes: Instead

```python
"""The Python interpreter to invoke from the shell.

This is typically ``sys.executable``, but there are a few exceptions.
If Nox was "frozen" with PyInstaller, then a special path will be used.
If no executable can be resolved, this will raise a stored exception.
"""
```

nox/virtualenv.py
```diff
@@ -182,6 +196,92 @@ def _resolved_interpreter(self):
...
+
+        if self._runtime is not None:
+            return self._runtime
+
+        if not _FROZEN:
+            self._runtime = sys.executable
+            return self._runtime
+
+        def find_binary(major=None, minor=None):
+            if _SYSTEM != "Windows":
```
dhermes: Please make `"Windows"` a constant.

nox/virtualenv.py
```diff
@@ -182,6 +196,92 @@ def _resolved_interpreter(self):
         self._resolved = InterpreterNotFound(self.interpreter)
         raise self._resolved
 
+    @property
+    def _runtime_interpreter(self):
```
dhermes: This property above seems to be mostly duplicated effort for locating the interpreter that is done elsewhere

nox/virtualenv.py
```diff
@@ -196,11 +296,22 @@ def create(self):
             logger.debug("Re-using existing virtualenv at {}.".format(self.location))
             return False
 
-        cmd = [sys.executable, "-m", "virtualenv", self.location]
+        cmd = [self._runtime_interpreter]
+        if not _FROZEN:
+            cmd.extend(["-m", "virtualenv"])
+        else:
+            # Nox was frozen with PyInstaller
+            # We have 2 options, import virtualenv and patch sys.argv to simlate
```
dhermes:
```suggestion
            # We have 2 options, import virtualenv and patch sys.argv to simulate
```

nox/virtualenv.py
```diff
@@ -196,11 +296,22 @@ def create(self):
...
+            # a normal CLI call, or, find a python interpreter matching the one
+            # Nox was frozen with, or close enough, and feed it the path to the,
+            # also frozen, virtualenv.py module
+            cmd.append(os.path.join(sys._MEIPASS, "site-packages", "virtualenv.py"))
```
dhermes: Can you provide documentation explaining why this is the way it must be done? ISTM that `python -m` would be **perfect** for the "frozen" case so I'm unclear why it isn't supported. It's also a bit worrying that this relies on `_MEIPASS` which is non-public.

tests/test_virtualenv.py
```diff
@@ -152,6 +154,46 @@ def test_create_interpreter(make_one):
     assert dir_.join("bin", "python3").check()
 
 
+@pytest.mark.skipif(IS_WINDOWS, reason="Not applicable on Windows.")
+@mock.patch("sys._MEIPASS", new="/tmp/_MEIxxx", create=True)
```
dhermes: Does `/tmp/_MEIxxx` need to exist?

nox/virtualenv.py
s0undt3ch: Catching the attribute exception is actually more performant than `hasattr`, but I'll change.

nox/virtualenv.py
s0undt3ch: Ok

nox/virtualenv.py
s0undt3ch: ok

nox/virtualenv.py
s0undt3ch: ok

nox/virtualenv.py
s0undt3ch: ok

nox/virtualenv.py
s0undt3ch: ok

nox/virtualenv.py
s0undt3ch: This is my attempt to follow `_resolved_interpreter`, what do you suggest?

nox/virtualenv.py
s0undt3ch: https://pyinstaller.readthedocs.io/en/v3.3.1/runtime-information.html#using-file-and-sys-meipass

As you mention, `python -m` would be the perfect choice, if not for the fact that the `python` referred is `sys.executable` which, on a frozen single binary, it's the actual frozen binary.
That leaves us with a few options:
1. Have the frozen nox accept the `-m` argument. Then use that value to actually import `virtualenv`, patch `sys.argv` to fake a regular call to it(`python virtualenv.py`), then call it's `main()` function. This has many potential failure points.
2. Provide an upstream change to virtualenv to have it's `main()` function accept a list of args, `def main(args=None)`. If accepted, we would include virtualenv in the frozen nox binary and we would just import it and provide reasonable `args` for it to work, without munging with `sys.argv`. This is likely the best option, but it needs to be accepted upstream and doesn't solve the immediate problem.
3. Include `virtualenv.py` as a data file in the frozen binary, and call it with an available runtime python interpreter. This is the path followed here.

tests/test_virtualenv.py
s0undt3ch: What do you mean?

s0undt3ch:
I found additional issues when running the frozen binary under opensuse and ubuntu, decided to create a branch in my fork which re-adds py2 support(since it takes less time to maintain that than to pursue the issues we found).
Closing.

I'm assuming you don't wan't to add py2 support back in into nox correct?

PR #172, opened 2019-03-11
Allow not passing "--upgrade" to `session.install`

s0undt3ch:
I don't really know the reasoning why `--upgrade` was added by default to the `pip` call in `session.install`.
I would reason against it, but, since it has been the default, at least allow the user not to upgrade.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -243,7 +243,7 @@ class Session:
         if "silent" not in kwargs:
             kwargs["silent"] = True
 
-        self._run("pip", "install", "--upgrade", *args, external="error", **kwargs)
+        self._run("pip", "install", *args, external="error", **kwargs)
 
     def notify(self, target):
         """Place the given session at the end of the queue.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -46,7 +46,7 @@ If you run this via ``nox`` you should see output similar to this::
 
     nox > Running session lint
     nox > virtualenv /tmp/example/.nox/lint
-    nox > pip install --upgrade flake8
+    nox > pip install flake8
     nox > flake8
     nox > Session lint successful. :)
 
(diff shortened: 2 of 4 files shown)


theacodes:
This is a great point. I'm not sure why I ended up going with `--upgrade` by default. I'd actually be curious if we should just remove that altogether.

@dhermes WDYT?

dhermes:
I think it should be **off** by default and have an `upgrade=True` override.

nox/sessions.py
```diff
@@ -242,8 +242,11 @@ def install(self, *args, **kwargs):
 
         if "silent" not in kwargs:
             kwargs["silent"] = True
+        upgrade = kwargs.pop("upgrade", True)
```
dhermes: I am all for this but I'd make `False` the default

theacodes:
> have an upgrade=True override.

At that point we could just pass it through like every other pip parameter, e.g.:

```python
session.install("--upgrade", "-r", "requirements.txt")
```

dhermes:
Ha, DERP. Good call

theacodes:
Cool, sound good to you, @s0undt3ch?

nox/sessions.py
s0undt3ch: That's a backwards incompatible change, but I'm cool with if you're cool with it too.

s0undt3ch:
Sorry, just woke and now I read the whole list of comments.
Yeah, I'd rely on the existing way to pass flags, no keyword arguments, just pass the actually pip CLI flag.

s0undt3ch:
I'll update the PR.

s0undt3ch:
PR updated.

dhermes:
Seems like some documentation needs to be updated as well: https://github.com/theacodes/nox/search?q=upgrade&unscoped_q=upgrade

s0undt3ch:
Docs updated.

theacodes:
Thank you, @s0undt3ch!

Issue #173, opened 2019-03-13
Log pip install command output


s0undt3ch:
**How would this feature be useful?**
Besides making it similar to tox(allowing for a smoother transition to nox), it's often useful in case of issues to, even though installing the requirements didn't fail, know which versions were actually installed.

**Describe the solution you'd like**
One solution would be do always show that output.
The other solution would be to add a `--verbose` boolean flag which, when passed, would additionally log the output of the pip install command (https://github.com/s0undt3ch/nox/commit/b5049abd13682a4e8b10fb67c413de53645cd1b4)

**Describe alternatives you've considered**
The only alternative is "patching" nox or submitting a PR to fix/enhance nox.

s0undt3ch:
This is particularly useful in CI contexts

theacodes:
Don't we already show that output if anything goes wrong?

I'm -1 for showing pip output always by default, it's incredibly noisy.

s0undt3ch:
Only if something goes wrong, yes, if nothing goes wrong, we can never get that output.
I'm also -1 with always showing that output, but the additional flag to show the output of all commands? I'd go for that and that's what I did on a branch which I link above.
If that's accepted, I'll submit a PR.

theacodes:
Yeah since it's plumbed through the normal run thing we should let users
set that flag. Sounds good.

s0undt3ch:
Ok, I'll submit a PR.

erinwild-rubikloud:
See the comment I left on your PR for a way to achieve this currently:

> It's already possible to override the silent flag in session.install (see #157). If the desired outcome is conditional reporting particularly in a CI context, the function in the noxfile could potentially handle this via extra arguments.
> 
> For example, `nox -s tests -- verbose` when additional verbosity is desired:
> 
> ```python
> @nox.session(python='3.7')
> def tests(session):
>     verbosity = True if 'verbose' in session.posargs else False
>     session.install('numpy', silent=verbosity)
> ```

vaneseltine:
Was this closed by https://github.com/theacodes/nox/pull/174?

theacodes:
Yeah I think so.

PR #174, opened 2019-03-16
Allow showing the commands output

s0undt3ch:
Refs #173

### DIFF
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -41,7 +41,7 @@ def main():
         print(dist.version, file=sys.stderr)
         return
 
-    setup_logging(color=args.color)
+    setup_logging(color=args.color, verbose=args.verbose)
 
     # Execute the appropriate tasks.
     exit_code = workflow.execute(
--- a/nox/command.py
+++ b/nox/command.py
@@ -119,6 +119,9 @@ def run(
 
             raise CommandFailed("Returned code {}".format(return_code))
 
+        if output:
+            logger.output(output)
+
         return output if silent else True
 
     except KeyboardInterrupt:
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -261,6 +261,13 @@ However, this will never output colorful logs:
 .. _opt-report:
 
 
+Controling commands verbosity
+-----------------------------
+
+By default, Nox will only show output of commands that fail, or, when the commands get passed `silent=False`.
+By passing `--verbose` to Nox, all output of all commands run is shown, regardless of the silent argument.
+
+
 Outputting a machine-readable report
 ------------------------------------
 
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -158,6 +158,15 @@ options.add_options(
         help="Arguments following ``--`` that are passed through to the session(s).",
         finalizer_func=_posargs_finalizer,
     ),
+    _option_set.Option(
+        "verbose",
+        "-v",
+        "--verbose",
+        group="secondary",
+        action="store_true",
+        help="Logs the output of all commands run including commands marked silent.",
+        noxfile=True,
+    ),
     *_option_set.make_flag_pair(
         "reuse_existing_virtualenvs",
         ("-r", "--reuse-existing-virtualenvs"),
--- a/nox/logger.py
+++ b/nox/logger.py
@@ -26,7 +26,7 @@ class NoxFormatter(ColoredFormatter):
         self._simple_fmt = logging.Formatter("%(message)s")
 
     def format(self, record):
-        if record.level == OUTPUT:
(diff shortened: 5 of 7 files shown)


s0undt3ch:
This was the least intrusive approach but it means changing the only debug log call to info.
The alternative would be to introduce another log level.
Let me know if its preferred to go down that route.

theacodes:
Hmm, I'm not sold on this approach. I don't like the idea of introducing a new flag (I'm *very* conservative about this).

What do you (and others) think about being able to just override the "silent" arg to `install`?

```python
session.install("requests", silent=False)
```

s0undt3ch:
Since nox is pytest based, have you though about allowing noxfiles to extend the available CLI flags?
I know, off topic, but then again, not so much.

s0undt3ch:
Also, to answer your question, with your suggestion you can either have it always on or off(the output).
I'm not interested on that output locally, but I am interested on that output in a CI context...

nox/command.py
```diff
@@ -124,6 +124,9 @@ def run(
 
             raise CommandFailed("Returned code {}".format(return_code))
 
+        if output:
+            logger.debug(output)
```
theacodes: So the problem with using `logger.debug` is that the output will actually differ from the normal operation because the logger appends `nox >` to the begging of each log-line. I would prefer that this goes to `stderr` just like the block above that sais `if silent: sys.stderr.write(output)`.

theacodes:
Cool - this mostly looks good then, just one comment about using `logging` to output the content.

s0undt3ch:
If we don't use logging we'll need to pass down the verbose flag to where we want verbose to be respected.
That's a whole lot of additional changes...
An additional verbose log level without the prefix you mention?

s0undt3ch:
Do you want the verbose flag passed down?

theacodes:
I like the idea of an additional verbose level that doesn't include the
prefix. We already extend the logger to add the "success" level.

s0undt3ch:
Ok. I'll go down that route.

erinwild-rubikloud:
It's already possible to override the silent flag in session.install (see #157). If the desired outcome is conditional reporting particularly in a CI context, the function in the noxfile could potentially handle this via extra arguments.

For example, `nox -s tests -- verbose` when additional verbosity is desired:
```python
@nox.session(python='3.7')
def tests(session):
    verbosity = True if 'verbose' in session.posargs else False
    session.install('numpy', silent=verbosity)
```

s0undt3ch:
True, but you won't get any output unless the command fails.
I'm after output even if the install command doesn't fail.

erinwild-rubikloud:
I will verify it and update you but I’m fairly sure setting the silent flag means that it doesn’t have to fail for it to be printed. This is precisely what I was looking for when I submitted the PR.

It looks like there hasn’t been a release since then, so perhaps that’s all that needs to happen @theacodes

s0undt3ch:
@erinwild-rubikloud you stand correct, passing `silent=False` shows the pip install output.

theacodes:
I'm glad we got you what you need, but it may still be useful to have a
verbose mode that immediately turn on output for all of our "silent"
commands. Is this something you'd still be interested in adding?

s0undt3ch:
Sure, I'll get back to it once I get some spare time.

s0undt3ch:
Ready for another set of eyes

nox/__main__.py
```diff
@@ -216,6 +218,10 @@ def main():
         "--report", help="Output a report of all sessions to the given filename."
     )
 
+    secondary.add_argument(
+        "--verbose", help="Show verbose output.", action="store_true"
```
erinwild-rubikloud: Might be useful to mention here that it overrides the silent option.

theacodes: Yeah, please do expand on this. Something like "Shows the output of all commands run, regardless of the silent argument."

theacodes:
This looks mostly good except for the one comment. Can you add documentation as well? (If not, I can amend the PR to do it.)

s0undt3ch:
Applied suggestions.

theacodes:
This LGTM, but I'm going to merge #187 and fix this up to use the new options pattern first. :)

It'll make it in before the next release.

theacodes:
Okay, @s0undt3ch. I've rebased this to work with the changes for #187.

I *do* have an open question though: How does this handle cases such as `silent=True` and where the command fails?

s0undt3ch:
@theacodes Nothing is logged but the error is printed to stderr.

Also, fixed conflict and rebased.

s0undt3ch:
Conflicts resolved again.

theacodes:
Thank you thank you!

s0undt3ch:
Thank You!

PR #176, opened 2019-04-11
Use the relative path of the virtualenv in the "creating virtualenv" log message.

theacodes:


dhermes:
LGTM after-the-fact. I recall being the cause of the absolute paths, so I'm happy to see this shortened for pretty purposes

theacodes:
Yeah I laid the groundwork for this but never made the tiny change to make
it work. It was annoying me so I finally did it. :)

PR #179, opened 2019-04-12
Fix documentation link for Docker Cloud vs Hub

TheKevJames:
Looks like Docker changed some links around and the old link is now only available when logged in. Changed the link to something which should work even for users without DockerHub accounts.


TheKevJames:
@theacodes noticed this while pushing a docker build for 2019.4.11, which should be live on DockerHub soon!

theacodes:
Perfect, thank you, @TheKevJames!

PR #180, opened 2019-04-13
Allow specifying `stdout` to Popen

s0undt3ch:


s0undt3ch:
The reason behind this change is that on a particular session, we need access to the output, even if the command failed, for CI reporting reasons.

theacodes:
Can you show me an example of when you'd use this and why the existing mechanisms aren't giving your the output you expect?

mostealth:
When running in CI you might want to save pylint output to file (to be later parsed). pylint does not support outputting to file hence the only solution I found was to monkey patch subprocess.Popen so that I inject my own stdout/stderr file before calling session.run

s0undt3ch:
What @mostealth said, that's our exact same scenario and why I created this PR.

This solution in use on our repo can be seen [here](https://github.com/saltstack/salt/blob/2018.3/noxfile.py#L819-L888)

nox/popen.py
```diff
@@ -16,13 +16,11 @@
 import sys
 
 
-def popen(args, env=None, silent=False):
-    stdout = None
-
-    if silent:
+def popen(args, env=None, silent=False, stdout=None, stderr=subprocess.STDOUT):
```
theacodes: What happens when both `silent` and `stdout` are specified? What happens when the command fails but `stdout` isn't `PIPE`? Can we add test cases for that?

theacodes:
That seems like a great reason. I had a comment about covering some edge cases, but I'm happy to merge this once we get that sorted. :)

s0undt3ch:
Rebased and added two new tests based on the comments provided.

theacodes:
Okay, I made `silent` and `stdout` mutually exclusive since `silent` doesn't have an effect if `stdout` is passed.

Does this still look good to you, @s0undt3ch? Also, it should be technically possible to override stderr now.  Do you mind adding a test for that?

s0undt3ch:
Added tests for custom `stderr`

theacodes:
Thank you, @s0undt3ch!

s0undt3ch:
Anytime!

PR #181, opened 2019-04-19
Allow bypassing the install only flag

s0undt3ch:
The reason behind this change is because, for our particular use case, we run some code against the virtualenv python in order to choose what requirements to install.
With this change, we can call nox with `--install-only` which skips all of the `session.run` calls, with the exception of those for which we explicitly pass `bypass_install_only`.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -177,7 +177,8 @@ class Session:
         if not args:
             raise ValueError("At least one argument required to run().")
 
-        if self._runner.global_config.install_only:
+        bypass_install_only = kwargs.pop("bypass_install_only", False)
+        if self._runner.global_config.install_only and not bypass_install_only:
             logger.info("Skipping {} run, as --install-only is set.".format(args[0]))
             return
 
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -126,6 +126,18 @@ class TestSession:
 
         assert "install-only" in caplog.text
 
+    def test_run_install_only_bypass(self, caplog):
+        caplog.set_level(logging.INFO)
+        session, runner = self.make_session_and_runner()
+        runner.global_config.install_only = True
+
+        with mock.patch.object(nox.command, "run") as run:
+            session.run("spam", "eggs", bypass_install_only=True)
+
+        assert "install-only" not in caplog.text
+
+        run.assert_called_once_with(("spam", "eggs"), env=mock.ANY, path=mock.ANY)
+
     def test_run_install_only_should_install(self):
         session, runner = self.make_session_and_runner()
         runner.global_config.install_only = True


dhermes:
This seems unnecessary. You can use `session.posargs` to pass through your option and then use a `return` statement in your session.

dhermes:
You could also subclass `Session` and override `install`

s0undt3ch:
You're missing the point. It has nothing to do with posargs.

Nox is responsible to install dependencies.
We lock dependencies based on python version and distro.
We need to know what's the python version on the virtualenv to install the right dependencies, and for that we get `sys.version_info` from the virtualenv's python, **through a `session.run`**.
If we can't know the version info, we can't install the right set of dependencies, hence, the proposed bypass.

dhermes:
If you want to know the currently running Python:

```python
@nox.session(py=("3.7", ...))
def foo(session):
    interpreter = session.virtualenv.interpreter
    ...
```

s0undt3ch:
I'm not creating PRs just because...

I know I can get the python version from there, but our goal is to make our users life easier, so, we also have `2` and `3` as valid python versions(thus nox sessions too), but we lock our requirements against the python `major` AND `minor` version information.
So that won't work.

s0undt3ch:
```python
@nox.session(py=("2", ...))
def foo(session):
    interpreter = session.virtualenv.interpreter
    print(interpreter)
    ...
```
```
❯ nox -e 'runtests-zeromq-2(coverage=False)' -- -v --proxy
nox > Running session runtests-zeromq-2(coverage=False)
nox > Creating virtualenv using python2 in /home/vampas/projects/SaltStack/salt/branches/2017.7.9/.nox/runtests-zeromq-2-coverage-false
nox > Session runtests-zeromq-2(coverage=False) was successful.
nox > Running session runtests-parametrized-2(coverage=False, crypto=None, transport='zeromq')
nox > Creating virtualenv using python2 in /home/vampas/projects/SaltStack/salt/branches/2017.7.9/.nox/runtests-parametrized-2-coverage-false-crypto-none-transport-zeromq
2
```
notice the above lonely `2`...

theacodes:
Yeah this is too far. I don't like the idea of an argument that reverses a flag that itself reverses an argument.

It's just Python, you can use `nox.command.run` directly or use `subprocess` with `session.virtualenv.bin` if you need to here.

s0undt3ch:
Your suggestions mean recreating the logic nox has on the noxfile or don't use nox at all(subprocess), that defeats the purpose of having nox to help out setting up the virtualenv...

theacodes:
I think for now this still just doesn't appeal to my sensibilities. Another alternative is to use session arguments to do what you want:

```python
@nox.session
def thing(session):
     if "setup-only" in session.posargs:
         ...
     else:
         ...
```

Or have a shared session like:

```
@nox.session
def setup(sesssion):
    ...

@nox.session
def thing(session)
    setup(session)
    ...
```

Happy to revisit this later as "no is temporary, but yes is forever" - https://twitter.com/solomonstre/status/715277134978113536?lang=en

PR #185, opened 2019-05-13
Make --list list all available sessions, not just the selected ones.

### LINKED ISSUE #161: List all sections when a list of sections is set using nox.options
**Describe the bug**
Setting `nox.options.sessions` results in `nox -l` only those sections, not all the sections.

**How to reproduce**
Create a `noxfile.py` with 3 sessions (format, tests, docs) and set `nox.options.sessions = ["format"]`, then run `nox -l`:

```
Available sessions:
* format
```

**Expected behavior**
It would be better if `nox -l` would actually list all the sessions.

---

Related Feature Request: The sessions that are run by default be marked with a `*` while the others won't have a bullet:

```
Available sessions:
* format
  tests
  docs
```

theacodes:
Closes #161

### DIFF
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -57,6 +57,7 @@ class GlobalConfig:
         self.install_only = args.install_only
         self.posargs = args.posargs
         self.report = args.report
+        self.color = not args.nocolor or args.forcecolor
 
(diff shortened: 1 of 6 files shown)


theacodes:
output

/cc @pradyunsg WDYT?

PR #193, opened 2019-05-16
Use 'pytest' instead of 'py.test' in examples and configuration

nicoddemus:
'pytest' is the recommended command since version 3.0. 

The command was called 'py.test' only for historical reasons:

https://docs.pytest.org/en/latest/faq.html#why-can-i-use-both-pytest-and-py-test-commands

Gratz on the project! 👍

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -137,13 +137,13 @@ class Session:
 
         Commands must be specified as a list of strings, for example::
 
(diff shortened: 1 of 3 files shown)


theacodes:
Nice. :) Thank you @nicoddemus

theacodes:
I think we'd gotten most of these that were user-facing previously, but always good to get the rest taken care of. :)

PR #196, opened 2019-05-18
Add interactive property to session.

AGeekInside:
I took a stab at adding the interactive attribute to session. The tests are present, but I think it'll need some testing in a non-interactive pipeline to confirm it works. The unit tests in test_sessions feel a bit clunky with the mocks, so suggestions on making that cleaner would be great.

I need to sort out how to update the docs as well to include the new attribute.

If this looks reasonable to folks, I can take a look at adding the '--non-interactive' flag.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -118,7 +118,7 @@ class Session:
 
     @property
     def interactive(self):
-        """Returns true if the Nox is being run in an interactive session or False otherwise."""
+        """Returns True if the Nox is being run in an interactive session or False otherwise."""
         return sys.stdout.isatty()
 
     def chdir(self, dir):
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -119,7 +119,7 @@ class Session:
     @property
     def interactive(self):
         """Returns True if Nox is being run in an interactive session or False otherwise."""
-        return sys.stdout.isatty()
+        return not self._runner.global_config.non_interactive and sys.stdin.isatty()
 
     def chdir(self, dir):
         """Change the current working directory."""
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -118,7 +118,7 @@ class Session:
 
     @property
     def interactive(self):
-        """Boolean showing if the session was ran in an interactive sessions."""
+        """Returns true if the Nox is being run in an interactive session or False otherwise."""
         return sys.stdout.isatty()
 
     def chdir(self, dir):
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -118,7 +118,7 @@ class Session:
 
     @property
     def interactive(self):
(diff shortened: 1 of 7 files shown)


nox/sessions.py
```diff
@@ -115,6 +116,11 @@ def bin(self):
         """The bin directory for the virtualenv."""
         return self._runner.venv.bin
 
+    @property
+    def interactive(self):
+        """Boolean showing if the session was ran in an interactive sessions."""
```
theacodes: Use active voice, please: `Returns True if the Nox is being run in an interactive session or False otherwise.`

tests/test_sessions.py
```diff
@@ -83,6 +83,21 @@ def test_properties(self):
         assert session.bin is runner.venv.bin
         assert session.python is runner.func.python
 
+    def test_interactive(self):
+        with mock.patch("nox.sessions.sys") as mock_sys:
```
theacodes: Is it possible to just patch `sys.stdout.istty`?

nox/sessions.py
AGeekInside: Just changed it in the latest commit

tests/test_sessions.py
AGeekInside: Yep. I just pushed a commit with the change.

theacodes:
This is wonderful, @AGeekInside. I'm going to merge #187 first and add a `--non-interactive` option to negate this before merging. Thank you so much!

tests/test_sessions.py
```diff
@@ -83,6 +83,17 @@ def test_properties(self):
...
+    def test_interactive(self):
+        with mock.patch("nox.sessions.sys.stdout.isatty") as m_isatty:
+
+            m_isatty.return_value = True
+            session, runner = self.make_session_and_runner()
+            assert session.interactive == m_isatty.return_value
+
+            m_isatty.return_value = False
+            session, runner = self.make_session_and_runner()
+            assert session.interactive == m_isatty.return_value
```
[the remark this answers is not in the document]
AGeekInside: I updated the assertion to use 'is'.

tests/test_sessions.py
```diff
@@ -83,6 +83,17 @@ def test_properties(self):
         assert session.bin is runner.venv.bin
         assert session.python is runner.func.python
 
+    def test_interactive(self):
+        with mock.patch("nox.sessions.sys.stdout.isatty") as m_isatty:
+
+            m_isatty.return_value = True
+            session, runner = self.make_session_and_runner()
+            assert session.interactive == m_isatty.return_value
```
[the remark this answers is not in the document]
AGeekInside: I updated the assertion to use 'is'.

theacodes:
Thank you, @AGeekInside!

Issue #99, opened 2018-08-22
Add `needs_nox` parsing to specify minimum version


dhermes:
The current switch from `0.19.1` to `>= 2018` will probably cause some breakage and it may be useful to specify

```python
needs_nox = "2018.7.31"
```

much in the same way Sphinx supports a

```python
needs_sphinx = "1.0"
```

theacodes:
Where does `needs_nox` need to go?

dhermes:
In `nox.py`; `needs_sphinx` goes in `docs/conf.py`. (Maybe `NEEDS_NOX` to be PEP8 friendly?)

theacodes:
Interesting. What would be our behavior if the version wasn't satisfied?

(Also, keep in mind I'm planning on `s/nox.py/noxfile.py` so we can support `python -m nox`).

dhermes:
> What would be our behavior if the version wasn't satisfied?

Similar to what Sphinx does? At c64cc8e43ca41ef1e9eae71073f0b696538e3dc3, the diff

```diff
diff --git a/docs/conf.py b/docs/conf.py
index ab3d6ac..bd8d482 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,7 @@ sys.path.insert(0, os.path.abspath('.'))
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#needs_sphinx = '1.0'
+needs_sphinx = '1.8.0'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
```

yields the following

```
$ nox -s docs
...
nox > cd docs
nox > sphinx-build -b html -d _build/doctrees . _build/html
Running Sphinx v1.6.2
making output directory...

Sphinx version error:
This project needs at least Sphinx v1.8.0 and therefore cannot be built with this version.
nox > Command sphinx-build -b html -d _build/doctrees . _build/html failed with exit code 1
nox > Session docs failed.
```

> (Also, keep in mind I'm planning on `s/nox.py/noxfile.py` so we can support `python -m nox`).

Yay!

theacodes:
Sounds reasonable.

cjolowicz:
I'd be interested in picking this up.

theacodes:
Go for it.

PR #210, opened 2019-07-05
`needs_nox` in noxfile

Paulius-Maruska:
Hello, in an attempt to familiarize myself with nox code, I managed to implement #99

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -316,7 +316,7 @@ Produces these sessions when running ``nox --list``:
 Minimum required nox version
 ----------------------------
 
-Minimum required Nox version can be specified in your ``noxfile.py`` by simply defining
+Minimum required Nox version can be specified in your ``noxfile.py`` file by simply defining
 a ``needs_nox`` global variable. For example:
 
 .. code-block:: python
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -313,6 +313,22 @@ Produces these sessions when running ``nox --list``:
     * tests(mysql, new)
 
 
+Minimum required nox version
+----------------------------
+
+Minimum required Nox version can be specified in your ``noxfile.py`` by simply defining
+a ``needs_nox`` global variable. For example:
+
+.. code-block:: python
+
+    import nox
+
+    needs_nox = '2019.5.30'
+
+    @nox.session(name='test')  # name argument was added in 2019.5.30
+    def pytest(session):
+        session.run('pytest')
+
 
 The session object
 ------------------
--- a/nox/_version.py
+++ b/nox/_version.py
@@ -1,0 +1,60 @@
+# Copyright 2017 Alethea Katherine Flowers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from distutils.version import LooseVersion
+from os import path
+from re import compile
+
+from pkg_resources import get_distribution
+
+from nox.logger import logger
+
+
+def nox_version():
+    # obtain version of the currently running nox instance
+    dist = get_distribution("nox")
+    return LooseVersion(dist.version)
+
+
+def parse_needs_nox(noxfile):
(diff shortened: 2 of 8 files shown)


theacodes:
@crwilcox @dhermes @stsewd do any of you have a moment to look at this? I don't have the bandwidth right now.

@Paulius-Maruska thank you for submitting this!

nox/_version.py
```diff
@@ -0,0 +1,60 @@
...
+        return None
+
+    # find a "needs_nox" line and parse the version
+    regex = compile(r"""^needs_nox\s*=\s*(['"])(?P<version>[^\1]*)\1.*$""")
+    with open(noxfile, "r") as noxfile:
+        for line in noxfile:
+            m = regex.match(line)
+            if m:
+                needs_version = LooseVersion(m.group("version"))
+                return needs_version
```
stsewd: Probably this should return the group and create the `LooseVersion` object in the `needs_nox` function

Paulius-Maruska:
I have updated the PR, and also rebased it on the latest master branch.

tests/test_tasks.py
```diff
@@ -64,6 +64,28 @@ def test_load_nox_module_not_found():
...
+
+
+def test_load_nox_module_returns_2_when_needed_version_is_higher():
+    config = _options.options.namespace(
+        noxfile=os.path.join(RESOURCES, "noxfile_future.py")
+    )
+    assert tasks.load_nox_module(config) == 2
+
+
+def test_load_nox_module_returns_2_when_noxfile_does_not_even_import():
```
crwilcox: It isn't clear to me what this validates. Is this validating the ordering of errors? That we will mention that we have an invalid arg to the decorator before the version?

tests/test__version.py
```diff
@@ -0,0 +1,74 @@
...
+
+def test_needs_nox_returns_none_when_provided_module_does_not_have_it():
+    class Module:
+        pass
+
+    needed = _v.needs_nox("not/important/noxfile.py", Module)
+    assert needed is None
+
+
+def test_is_version_sufficient_returns_true_for_lower_version():
```
crwilcox: should there be a case for exact match?

docs/config.rst
```diff
@@ -313,6 +313,22 @@ Produces these sessions when running ``nox --list``:
     * tests(mysql, new)
 
 
+Minimum required nox version
+----------------------------
+
+Minimum required Nox version can be specified in your ``noxfile.py`` file by simply defining
```
crwilcox: nitpick: may make sense to drop the word 'simply' as there can be negative feelings about that verbiage in prose.

```suggestion
Minimum required Nox version can be specified in your ``noxfile.py`` file by defining
```

theacodes:
In terms of placating linters and continuing to be somewhat idiomatic, what do y'all think of making `needs_version` a member of the `nox` module?

```python
import nox

nox.needs_version = ">=2018.7.8"
nox.options.sessions = ["lint", "tests-3.6"]

@nox.session
def lint(s):
    ...

...
```

crwilcox:
I think adding this to the module would make sense. It also may be more discoverable via language intelligence services in IDEs. Though my big sticking point is matching PyLint, just to avoid the bug that will follow if this doesn't :D

nox/_version.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def is_version_sufficient(noxfile, needed_version):
+    """Return True only when currently running nox version matches or is
+    newer than the version needed.
+    """
+    if needed_version and needed_version > nox_version():
+        logger.error(
+            "Noxfile {} needs at least Nox {} and therefore it cannot be used "
+            "with this version.".format(noxfile, str(needed_version))
```
stsewd: This message could include the current nox version too

nox/_version.py
```diff
@@ -0,0 +1,67 @@
...
+    regex = compile(r"""^needs_nox\s*=\s*(['"])(?P<version>[^\1]*)\1.*$""")
+    with open(noxfile, "r") as noxfile:
+        for line in noxfile:
+            m = regex.match(line)
+            if m:
+                needs_version = m.group("version")
+                return needs_version
+
+
+def needs_nox(noxfile, noxfile_module=None):
```
stsewd: Np: this function is doing two things. And only using one arg. If noxfile and noxfile module are given, noxfile module is only used

Paulius-Maruska:
I'm sorry for the silence this past week - I was attending the EuroPython2019 conference and was quite busy with the various activities here. :)

I will try to address the issues / respond next week as early as I can.

theacodes:
Take your time, and thanks for keeping us updated. 🙂

theacodes:
I'm going to close this PR due to inactivity. That's not to say I'm not interested in merging it! If you're still up for it just reply and I'll re-open the PR. :)

Issue #199, opened 2019-05-26
Allow use of venv rather than virtualenv


pfmoore:
**How would this feature be useful?**
Some projects (for example, `pipx`) have test suites that cannot run under virtualenv, but require venv. Being able to choose whether to use virtualenv or venv would allow using nox to run the test suite.

**Describe the solution you'd like**
The simplest solution I can think of is to use a parameter to `nox.session`, something like `@nox.session(stdlib_venv=True)`. But honestly, it doesn't matter much to me how it gets specified.

**Describe alternatives you've considered**
I can't really think of a good alternative, other than "use something else" :-( Ideally, `virtualenv` and `venv` would be sufficiently compatible that it doesn't matter which you use, but that simply doesn't seem to be the case.

stsewd:
`nox` is python3 only, so I guess we can just replace virtualenv?

theacodes:
No, unfortunately we still need it to create virtual environments for 2.7
sessions.

pfmoore:
... and there are things that only work in virtualenv (there are some nasty interactions if, for example, you create a venv from within a virtualenv). So I definitely think it's something where the user needs to be able to choose. Of course, as someone who wants to ditch Python 2.7 ASAP, my *personal* preference would be to go venv by default with an option to use virtualenv, but that would be backward incompatible, so I'm fine with venv being the one that needs to be opt-in.

theacodes:
Thinking of future-proofing this for `conda`, `pyenv`, or `pipenv` support as well (see #183 #4 #159) we should probably just extend the pattern we have with `ProcessEnv` and `VirtualEnv` to include `Venv`, `Conda and `PyEnv` backends as well. We could abstract a bit more common code of the current `VirtualEnv` class to make it the base for these new backends.

Session decorator wise, I think it makes the most sense to have something like:

```python
@nox.session(py="3.7", venv_backend="venv")
```

Thoughts?

stsewd:
I like the idea of having different backends, shouldn't be this a global option that could be overridden by each session?

theacodes:
I think per session alone makes the most sense, but happy to hear reasoning
for another global opt.

cs01:
I'm trying to migrate pipx's test suite to nox right now and am running into this issue. 

The problem is the pip executable is pointing to nox's rather than one that pipx created with venv:
```
>> ll /tmp/pipx_tests__fc881ve/subdir/pipxhome/.cache/3d7e26ae000b6c1/bin/python 
[...]

PR #215, opened 2019-07-13
Add `session(venv_backend='conda')` option to use Conda environments.

### LINKED ISSUE #183: Include Conda-install package functionality
Reading the documentation , nox provides automated pip-install packages feature ( that can be defined in the configuration file).
 It would be useful to support conda-install as well , to be able to include installing miniconda/anaconda package install .

I would like to work on this as my first contribution , if anyone could help me get started, would be really helpful!

tswast:
Also adds a `session.conda_install()` function to install packages with
conda, instead of pip.

Towards #199

Closes #183

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -207,12 +207,11 @@ class Session:
         if self._runner.global_config.error_on_external_run:
             kwargs.setdefault("external", "error")
 
-        # If we aren't using a virtualenv allow all external programs.
-        if not isinstance(self.virtualenv, (CondaEnv, VirtualEnv)):
+        # Allow all external programs when running outside a sandbox.
+        if not self.virtualenv.is_sandboxed:
             kwargs["external"] = True
 
-        # Allow the global conda command from conda environments.
-        if isinstance(self.virtualenv, CondaEnv) and args[0] == "conda":
+        if args[0] in self.virtualenv.allowed_globals:
             kwargs["external"] = True
 
         # Run a shell command.
@@ -258,7 +257,6 @@ class Session:
             "conda",
             "install",
             "--yes",
-            "--use-index-cache",
             "--prefix",
             self.virtualenv.location,
             *args,
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -254,6 +254,7 @@ class Session:
             "conda",
             "install",
             "--yes",
+            "--use-index-cache",
             "--prefix",
             self.virtualenv.location,
             *args,
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -208,7 +208,11 @@ class Session:
             kwargs.setdefault("external", "error")
 
         # If we aren't using a virtualenv allow all external programs.
-        if not isinstance(self.virtualenv, VirtualEnv):
+        if not isinstance(self.virtualenv, (CondaEnv, VirtualEnv)):
+            kwargs["external"] = True
+
+        # Allow the global conda command from conda environments.
+        if isinstance(self.virtualenv, CondaEnv) and args[0] == "conda":
             kwargs["external"] = True
 
         # Run a shell command.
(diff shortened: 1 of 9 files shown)


tswast:
Tested locally, but needs unit tests and CI tests. I'm thinking of using CircleCI for the conda tests, since it supports a docker base image. See use of conda in pandas-gbq config:

https://github.com/pydata/pandas-gbq/blob/9460ad6444ef750cbd46aae12897001ce72867ac/.circleci/config.yml#L40-L42

tswast:
For some reason Travis builds aren't triggering anymore.

I'm having trouble getting this PR to work on Windows with AppVeyor. It seems there are many more possible paths for binaries in the Windows version of conda.

https://github.com/conda/conda/blob/3ca57b16effb02f961631305fcd941d96cd856eb/conda/activate.py#L511-L517

tswast:
```
PackagesNotFoundError: The following packages are not available from current channels:
  - python
  - pip
Current channels:
  - https://conda.anaconda.org/conda-forge/win-64
  - https://conda.anaconda.org/conda-forge/noarch
  - https://repo.anaconda.com/pkgs/main/win-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/win-64
  - https://repo.anaconda.com/pkgs/r/noarch
  - https://repo.anaconda.com/pkgs/msys2/win-64
  - https://repo.anaconda.com/pkgs/msys2/noarch
```

No idea why it's working for Python 3.5, but not Python 3.6.

nox/sessions.py
```diff
@@ -208,12 +208,64 @@ def _run(self, *args, env=None, **kwargs):
             kwargs.setdefault("external", "error")
 
         # If we aren't using a virtualenv allow all external programs.
-        if not isinstance(self.virtualenv, VirtualEnv):
+        if not isinstance(self.virtualenv, (CondaEnv, VirtualEnv)):
```
theacodes: flip this to check if its an instance of `ProcessEnv`?

nox/sessions.py
```diff
@@ -208,12 +208,64 @@ def _run(self, *args, env=None, **kwargs):
...
+            kwargs["external"] = True
+
+        # Allow the global conda command from conda environments.
+        if isinstance(self.virtualenv, CondaEnv) and args[0] == "conda":
```
theacodes: Hrm special casing seems weird here. Maybe we can make `*Env` classes have a `whitelisted_globals` property?

nox/sessions.py
```diff
@@ -208,12 +208,64 @@ def _run(self, *args, env=None, **kwargs):
...
+            raise ValueError("At least one argument required to install().")
+
+        if "silent" not in kwargs:
+            kwargs["silent"] = True
+
+        self._run(
+            "conda",
+            "install",
+            "--yes",
+            "--use-index-cache",
```
theacodes: What if people don't want to use the index cache?

nox/sessions.py
tswast: Both `CondaEnv` and `VirtualEnv` subclass `ProcessEnv`. Could use `type` or adjust the inheritance structure, but I added a `is_sandboxed` property, instead.

nox/sessions.py
tswast: Good idea. Done.

nox/sessions.py
tswast: Oops. You're right, removed. I added that so that I could code on the airplane.

Now they can add `--use-index-cache` from the `conda_install` function call, and by default it uses the default CLI params.

tswast:
And now AppVeyor is passing. It was probably the `--use-index-cache` that was messing it up.

Anyone know how to trigger Travis? It seems to have given up on this PR.

crwilcox:
@tswast I tried to get CI to respond but it seems the path forward is to close this pr and reopen :(

PR #224, opened 2019-07-25
Resolve to interpreter 'python' in PATH if '--version' fits

### LINKED ISSUE #219: Python interpreter not found on conda/win32
**Describe the bug**
Using Anaconda on Windows 10, `nox` skips all sessions that explicitly specify a python version using the session decorator, e.g. `@nox.session(python="3.6")`, even though a python interpreter of the correct version is present in the path:
```
(base) C:\dev\mytest>where python
C:\ProgramData\Anaconda3\python.exe

(base) C:\dev\mytest>python --version
Python 3.6.5 :: Anaconda, Inc.
```

**How to reproduce**

Create a `noxfile.py` with at least one session that explicitly specifies the python version, e.g.
```python
import nox

@nox.session(python="3.6")
def with_version36(session):
    print("With python 3.6")

@nox.session(python="3.7")
def with_version37(session):
    print("With python 3.7")

@nox.session(python="3.7")
def without_version(session):
    print("Without python version")
```

Install Anaconda, open Anaconda Prompt, install nox, navigate to the directory containing the noxfile, and run `nox`:
```
(base) C:\>pip install nox
...
Successfully installed colorlog-3.2.0 nox-2019.5.30

(base) C:\>cd C:\dev\mytest

(base) C:\dev\mytest>nox
nox > Running session with_version36
nox > Session with_version36 skipped: Python interpreter 3.6 not found
nox > Running session with_version37
nox > Session with_version37 skipped: Python interpreter 3.7 not found
nox > Running session without_version
nox > Creating virtualenv using python.exe in .nox\without_version
Without python version
...
```

Checking the location of the interpreter reveals that there is a python 3.6 in the path:
```
(base) C:\dev\mytest>where python
C:\ProgramData\Anaconda3\python.exe

(base) C:\dev\mytest>python --version
Python 3.6.5 :: Anaconda, Inc.
```

I also tried it out in another 3.7-conda environment, even matrioshka'd inside a pipenv environment, with the same results.

**Expected behavior**
I expected both sessions `with_version36` and `without_version` to run.
I think the python interpreter of the correct version should be discoverable on conda, too.

Is `nox` perhaps looking for python binaries by name (expecting python3.exe, python36.exe and so on), instead of finding all interpreters in the path and then checking `--version`?

schuderer:
To fix #219 

TL;DR: Trying to implement the suggestion in https://github.com/theacodes/nox/issues/219#issuecomment-512933886, this PR adds a final check at the end of the property ``nox.virtualenv._resolved_interpreter`` which looks for an interpreter called "python" in the ``PATH``, checks whether its ``--version`` fits the required version, and returns it as the resolved interpreter if it matches.

### Rationale
As described in #219, sessions specifying the interpreter version (e.g. ``@nox.session(python='3.7')``) don't find the existing matching python 3.7 if it is called "python" or "python.exe" and no python launcher ("py.exe") is available.

The same sessions work when not specifying python='3.7', even if the used interpreter is actually python 3.7. If this happens, it makes it harder for collaborators on Unix/Windows to work on cross-platform libraries together (like ``nox`` or [my current pet project](/schuderer/mllaunchpad)) without temporarily adapting the noxfile to their own system (or their system to the noxfile).

While the mechanisms applied in ``nox.virtualenvs`` do follow the Python guidelines, in practice, Anaconda installations on Windows, which are quite popular in my scientific/data-wrangling field, provide neither versioned interpreters (like 'python2', 'python27', 'python3', 'python37') nor a python launcher.

For Anaconda-specific projects, the recent addition of conda environment support to ``nox`` is certainly very interesting. However, I am arguing from the perspective of cross-platform library development and would like to be able to provide a noxfile to my project which works on Unix and Windows, be it with Anaconda or with "proper" Python installations.

One could argue that this is not ``nox``s problem and should be fixed in Anaconda (and one would be right), but it is a pretty small addition and would in one step increase the reach for nox to everyone using Anaconda on Windows (which are quite a few people). And it would make it easier for me (and ``nox`` ;)) to find Windows contributors also.

### Implementation
Added the check (with an extra module-level function for cleanliness), and some parametrised tests. I also fixed two incorrect return value mocks of ``py.path.local.sysfind`` (which returns ``None`` on not found, not ``False`` like in the mock return value), which my code initially tripped over because it checked for ``None``.

As the body of ``_resolved_interpreter`` is essentially a string of if-else conditionals, creating good tests was a bit of a challenge. I hope they're not too ugly. :)

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -85,6 +85,39 @@ def locate_via_py(version):
             return None
 
 
+def locate_using_path_and_version(version):
+    """Check the PATH's python interpreter and return it if the version
+    matches.
+
+    On systems without version-named interpreters and with missing
+    launcher (which is on all Windows Anaconda installations),
+    we search the PATH for a plain "python" interpreter and accept it
+    if its --version matches the specified interpreter version.
+
+    Args:
+        version (str): The desired Python version. Of the form ``X.Y``.
+
+    Returns:
+        Optional[str]: The full executable path for the Python ``version``,
+        if it is found.
+    """
+    if not version:
+        return None
+
+    path_python = py.path.local.sysfind("python")
+    if path_python:
+        try:
+            version_pattern = "^Python {}".format(version)
+            version_string = path_python.sysexec("--version").strip()
+            match = re.match(version_pattern, version_string)
+            if match:
+                return path_python
+        except py.process.cmdexec.Error:
+            return None
+
+    return None
+
+
 class VirtualEnv(ProcessEnv):
     """Virtualenv management class.
 
@@ -173,9 +206,13 @@ class VirtualEnv(ProcessEnv):
             xy_version = cleaned_interpreter
 
         path_from_launcher = locate_via_py(xy_version)
-
         if path_from_launcher:
             self._resolved = path_from_launcher
+            return self._resolved
+
+        path_from_version_param = locate_using_path_and_version(xy_version)
+        if path_from_version_param:
+            self._resolved = path_from_version_param
             return self._resolved
 
         # If we got this far, then we were unable to resolve the interpreter
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -104,13 +104,13 @@ def locate_using_path_and_version(version):
     if not version:
         return None
 
+    script = "import platform; print(platform.python_version())"
     path_python = py.path.local.sysfind("python")
     if path_python:
         try:
-            version_pattern = "^Python {}".format(version)
-            version_string = path_python.sysexec("--version").strip()
-            match = re.match(version_pattern, version_string)
-            if match:
+            prefix = "{}.".format(version)
+            version_string = path_python.sysexec("-c", script).strip()
+            if version_string.startswith(prefix):
                 return path_python
         except py.process.cmdexec.Error:
             return None
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -111,7 +111,7 @@ def locate_using_path_and_version(version):
             prefix = "{}.".format(version)
             version_string = path_python.sysexec("-c", script).strip()
             if version_string.startswith(prefix):
-                return path_python
+                return str(path_python)
         except py.process.cmdexec.Error:
             return None
 
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -23,6 +23,7 @@ import nox.virtualenv
 
 
 IS_WINDOWS = nox.virtualenv._SYSTEM == "Windows"
+RAISE_ERROR = "RAISE_ERROR"
 
 
 @pytest.fixture
@@ -33,6 +34,47 @@ def make_one(tmpdir):
         return (venv, location)
 
     return factory
+
+
+@pytest.fixture
+def make_mocked_interpreter_path():
+    def factory(path, sysexec_result):
+        def mock_sysexec(*_):
+            if sysexec_result == RAISE_ERROR:
+                raise py.process.cmdexec.Error(1, 1, "", "", "")
+            else:
+                return sysexec_result
+
+        attrs = {
+            "sysexec.side_effect": mock_sysexec,
+            "__str__": mock.Mock(return_value=path),
+        }
+        mock_python = mock.Mock()
+        mock_python.configure_mock(**attrs)
+
+        return mock_python
+
+    return factory
+
+
+@pytest.fixture
+def patch_sysfind(make_mocked_interpreter_path):
+    def patcher(sysfind, only_find, sysfind_result, sysexec_result):
+        mock_python = make_mocked_interpreter_path(sysfind_result, sysexec_result)
+
+        def mock_sysfind(arg):
+            if sysfind_result is None:
+                return None
+            elif arg.lower() in only_find:
+                return mock_python
+            else:
+                return None
+
+        sysfind.side_effect = mock_sysfind
+
+        return sysfind
+
+    return patcher
 
 
 def test_process_env_constructor():
@@ -279,64 +321,53 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
     sysfind.assert_any_call("py")
 
 
-@pytest.mark.parametrize(
-    ["input_", "path_result", "expected"],
-    [
-        ("3.7", r"c:\python37-x64\python.exe", r"c:\python37-x64\python.exe"),
-        ("3.7", r"c:\fails\python.exe", None),
-        ("python3.7", r"c:\python37-x64\python.exe", None),
-        ("2.7", r"c:\python37-x64\python.exe", None),
-        ("goofy", r"c:\python37-x64\python.exe", None),
-        ("3.7", None, None),
-        ("python3.7", None, None),
-        ("2.7", None, None),
-        ("goofy", None, None),
-    ],
-)
 @mock.patch("nox.virtualenv._SYSTEM", new="Windows")
 @mock.patch.object(py.path.local, "sysfind")
 def test__resolved_interpreter_windows_path_and_version(
-    sysfind, make_one, input_, path_result, expected
+    sysfind, make_one, patch_sysfind
+):
+    # Establish that if we get a standard pythonX.Y path, we look it
+    # up via the path on Windows.
+    venv, _ = make_one(interpreter="3.7")
+
+    # Trick the system into thinking that it cannot find
+    # pythonX.Y up until the python-in-path check at the end.
+    # Also, we don't give it a mock py launcher.
+    # But we give it a mock python interpreter to find
+    # in the system path.
+    correct_path = r"c:\python37-x64\python.exe"
+    patch_sysfind(
+        sysfind,
+        only_find=("python", "python.exe"),
+        sysfind_result=correct_path,
+        sysexec_result="3.7.3\\n",
+    )
+
+    # Okay, now run the test.
+    assert str(venv._resolved_interpreter) == correct_path
+
+
+@pytest.mark.parametrize("input_", ["2.7", "python3.7", "goofy"])
+@pytest.mark.parametrize("sysfind_result", [r"c:\python37-x64\python.exe", None])
+@pytest.mark.parametrize("sysexec_result", ["3.7.3\\n", RAISE_ERROR])
+@mock.patch("nox.virtualenv._SYSTEM", new="Windows")
+@mock.patch.object(py.path.local, "sysfind")
+def test__resolved_interpreter_windows_path_and_version_fails(
+    sysfind, input_, sysfind_result, sysexec_result, make_one, patch_sysfind
 ):
     # Establish that if we get a standard pythonX.Y path, we look it
     # up via the path on Windows.
     venv, _ = make_one(interpreter=input_)
 
-    # Trick the system into thinking that it cannot find python3.6
-    # (it likely will on Unix). Also, we don't give it a dummy
-    # py launcher. But we give it a dummy python interpreter to find
+    # Trick the system into thinking that it cannot find
+    # pythonX.Y up until the python-in-path check at the end.
+    # Also, we don't give it a mock py launcher.
+    # But we give it a mock python interpreter to find
     # in the system path.
-    if path_result:
+    patch_sysfind(sysfind, ("python", "python.exe"), sysfind_result, sysexec_result)
 
-        def mock_sysexec(_):
-            if path_result == r"c:\fails\python.exe":
-                raise py.process.cmdexec.Error(1, 1, "", "", "")
-            else:
-                return "Python 3.7.3 | Anaconda\\n"
-
-        attrs = {
-            "sysexec.side_effect": mock_sysexec,
-            "__str__.return_value": path_result,
-        }
-        mock_python = mock.MagicMock()
-        mock_python.configure_mock(**attrs)
-
-        def mock_sysfind(arg):
-            if arg.lower() == "python" or arg.lower() == "python.exe":
-                return mock_python
-            else:
-                return None
-
-        sysfind.side_effect = mock_sysfind
-    else:
-        sysfind.return_value = None
-
-    # Okay now run the test.
-    if expected:
-        assert str(venv._resolved_interpreter) == expected
-    else:
-        with pytest.raises(nox.virtualenv.InterpreterNotFound):
-            venv._resolved_interpreter
+    with pytest.raises(nox.virtualenv.InterpreterNotFound):
+        venv._resolved_interpreter
 
 
 @mock.patch("nox.virtualenv._SYSTEM", new="Windows")
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -246,7 +246,7 @@ def test__resolved_interpreter_windows_pyexe(sysfind, make_one, input_, expected
     attrs = {"sysexec.return_value": expected}
     mock_py = mock.Mock()
     mock_py.configure_mock(**attrs)
-    sysfind.side_effect = lambda arg: mock_py if arg == "py" else False
+    sysfind.side_effect = lambda arg: mock_py if arg == "py" else None
 
     # Okay now run the test.
     assert venv._resolved_interpreter == expected
@@ -269,7 +269,7 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
     attrs = {"sysexec.side_effect": py.process.cmdexec.Error(1, 1, "", "", "")}
     mock_py = mock.Mock()
     mock_py.configure_mock(**attrs)
-    sysfind.side_effect = lambda arg: mock_py if arg == "py" else False
+    sysfind.side_effect = lambda arg: mock_py if arg == "py" else None
 
     # Okay now run the test.
     with pytest.raises(nox.virtualenv.InterpreterNotFound):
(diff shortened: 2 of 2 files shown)


tests/test_virtualenv.py
```diff
@@ -279,6 +279,66 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
...
+@mock.patch("nox.virtualenv._SYSTEM", new="Windows")
+@mock.patch.object(py.path.local, "sysfind")
+def test__resolved_interpreter_windows_path_and_version(
+    sysfind, make_one, input_, path_result, expected
+):
+    # Establish that if we get a standard pythonX.Y path, we look it
+    # up via the path on Windows.
+    venv, _ = make_one(interpreter=input_)
+
+    # Trick the system into thinking that it cannot find python3.6
```
dhermes: This comment refers to 3.6 but none of the test cases do. Then the next comment is about Unix but this test is meant to be Windows only?

tests/test_virtualenv.py
```diff
@@ -279,6 +279,66 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
...
+    # (it likely will on Unix). Also, we don't give it a dummy
+    # py launcher. But we give it a dummy python interpreter to find
+    # in the system path.
+    if path_result:
```
dhermes: There are 4 if statements and 2 closures in this test case. I'd much rather see it broken up into distinct test cases with a "single path of execution" to make it clear (and hopefully :crossed_fingers: more concise) what is happening in each.

tests/test_virtualenv.py
```diff
@@ -279,6 +279,66 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
...
+            if path_result == r"c:\fails\python.exe":
+                raise py.process.cmdexec.Error(1, 1, "", "", "")
+            else:
+                return "Python 3.7.3 | Anaconda\\n"
+
+        attrs = {
+            "sysexec.side_effect": mock_sysexec,
+            "__str__.return_value": path_result,
+        }
+        mock_python = mock.MagicMock()
```
dhermes: 1. `MagicMock` needed over `Mock` because of `__str__`?
1. Any way we can reduce the surface area by using `MagicMock(spec=())`?

tests/test_virtualenv.py
```diff
@@ -279,6 +279,66 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
...
+        mock_python.configure_mock(**attrs)
+
+        def mock_sysfind(arg):
+            if arg.lower() == "python" or arg.lower() == "python.exe":
```
dhermes: Is this your preference over

```python
if arg.lower() in ("python", "python.exe"):
```

tests/test_virtualenv.py
```diff
@@ -279,6 +279,66 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
...
+    venv, _ = make_one(interpreter=input_)
+
+    # Trick the system into thinking that it cannot find python3.6
+    # (it likely will on Unix). Also, we don't give it a dummy
+    # py launcher. But we give it a dummy python interpreter to find
+    # in the system path.
+    if path_result:
+
+        def mock_sysexec(_):
+            if path_result == r"c:\fails\python.exe":
```
dhermes: If this "fails" value is going to be used as a sentinel value like this, could you make it a module global?

tests/test_virtualenv.py
```diff
@@ -279,6 +279,66 @@ def test__resolved_interpreter_windows_pyexe_fails(sysfind, make_one):
...
+                return mock_python
+            else:
+                return None
+
+        sysfind.side_effect = mock_sysfind
+    else:
+        sysfind.return_value = None
+
+    # Okay now run the test.
+    if expected:
```
dhermes: `if expected is not None`? (i.e. we want a `None`-check not a truthy check)

nox/virtualenv.py
```diff
@@ -85,6 +85,39 @@ def locate_via_py(version):
...
+        Optional[str]: The full executable path for the Python ``version``,
+        if it is found.
+    """
+    if not version:
+        return None
+
+    path_python = py.path.local.sysfind("python")
+    if path_python:
+        try:
+            version_pattern = "^Python {}".format(version)
```
dhermes: Rather than having to use a regex, why not just `platform.python_version()`:

```
$ python2.7 -c 'import platform; print(platform.python_version())'
2.7.15
$ python3.5 -c 'import platform; print(platform.python_version())'
3.5.6
$ python3.6 -c 'import platform; print(platform.python_version())'
3.6.7
$ python3.7 -c 'import platform; print(platform.python_version())'
3.7.1
$ pypy -c 'import platform; print(platform.python_version())'
2.7.13
$ pypy3 -c 'import platform; print(platform.python_version())'
3.5.3
```

Also, if you do go with `python --version`, why resort to a regex when you could just do

```python
prefix = "Python {}.".format(version)  # Note the extra dot
version_string.startswith(prefix)
```

nox/virtualenv.py
```diff
@@ -173,11 +206,15 @@ def _resolved_interpreter(self):
             xy_version = cleaned_interpreter
 
         path_from_launcher = locate_via_py(xy_version)
-
         if path_from_launcher:
             self._resolved = path_from_launcher
             return self._resolved
 
+        path_from_version_param = locate_using_path_and_version(xy_version)
```
dhermes: Any reason not to do this **before** the `if _SYSTEM != "Windows":` check on line 197? I.e. just checking for `python` on `PATH` and matching version against `python --version` could work fine on non-Windows operating systems.

tests/test_virtualenv.py
schuderer: Yes, not my most elegant test case indeed. 😬 I tried to split it up earlier and found that I would have to repeat a lot preparation code (creating mocks and making sure the checks above the check-under-test fail). I *could* create yet another fixture to help with this, but I'm not sure whether this makes it even less transparent. I'm unsure, should I choose conciseness or clarity in this case?

tests/test_virtualenv.py
schuderer: I did not get "__str__" to work using `Mock`, but did with `MagicMock`, and looking at its [documentation](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock), an empty `spec` should just make it a `Mock`. I will look into getting plain `Mock` to work again, otherwise maybe resort to `MagicMock(spec=("__str__"))`.

Disclaimer: First-time-user of unittest.mock ;)

tests/test_virtualenv.py
schuderer: Oh dear, did I really write this? :)

nox/virtualenv.py
schuderer: I don't want to put the python-in-path-check before the Windows check because, it being a stop-gap solution, then it would precede the python-launcher-check, which is preferrable if it exists. (py can find different python versions, if present, while python-in-path can only check whether the python in the path has the correct version. By putting our python-in-path-check above the py-check we would forfeit the possibility).

Personally, if something needs to be changed about this, I'd rather just remove the `Windows` check (as I understand it, it is not doing anything functionally, just skipping some (potentially) useless checks).

tests/test_virtualenv.py
schuderer: Thanks for taking the time to review! I was copying some logic from another test and did not fully adapt the comment, will change it.

One difficulty with testing `_resolved_parameter` is that it checks for things one after another, so every test has to make sure that the checks above the check-under-test fail (of which most are for Unix, hence the original comment). I am inclined to think that refactoring `_resolved_parameter` (separately from this PR, maybe next after) would help reducing the complexity of the tests (factoring out the checks themselves, separating them from the flow).

tests/test_virtualenv.py
schuderer: This is a bit of a hack to signify that sysexec should throw an exception in the parametrization table. I can put it in a constant, but it would essentially still be the same hack, I am open to suggestion on how to handle this kind of more complex parametrization matrix. For now, I hope I can remove it when splitting up the test case as per your review comment above.

tests/test_virtualenv.py
dhermes: SGTM. Bravo on diving in!

tests/test_virtualenv.py
dhermes: Try to break this up without any `if` statements? If it gets too unwieldy we can revert?

tests/test_virtualenv.py
dhermes: Yup, no issue with the hack, just wanted to avoid copy-pasting the value in multiple places.

nox/virtualenv.py
dhermes: That may actually be OK, but I'd rather do it in a separate PR. We could do that one in parallel with this one?

tests/test_virtualenv.py
schuderer: Thanks for all your comments. I pushed my changes just now. Most of the branching actually was part of patching/mocking sysfind, the resulting path object, and python-sysexec the right way for the various parameters. Off-loading the mocking and patching into two pytest fixtures solved the complexity problem (and as a side-effect, created fixtures that are probably useful for other tests in the module as well).

schuderer:
Hope CI runs successfully. Refreshing from master brought some conda-related test sessions which fail both on macOS (`Program conda not found.`, which is correct) as well as on Windows (Conda is installed, but says that colorlog and tox are not available from the current channels `https://repo.anaconda.com/pkgs/[main/win-64|main/noarch|r/win-64|r/noarch|msys2/win-64|msys2/noarch]`, maybe @tswast knows whether this is a potential issue one should look into).

This causes `nox -s cover` to always fail locally, as well (my changes are in the same module `virtualenv` as the conda-related functionality). If anyone has any hints on how to easily validate partial coverage, they are welcome (and maybe a good idea to add to CONTRIBUTING).

nox/virtualenv.py
schuderer: Sounds good, left it as it is for now in this PR. Before starting another PR, I'll be sorta-kinda waiting first what else might pop up in this PR. 😃

theacodes:
I've marked this PR as `open collective eligible`. Once merge, both the *submitter* (@schuderer) and reviewer (@dhermes) can submit an expense to our [Open Collective](https://opencollective.com/python-nox) to get paid for this.

Our funds aren't magnificent, so I'm setting the cap here at $20/participant.

tswast:
> (Conda is installed, but says that colorlog and tox are not available from the current channels https://repo.anaconda.com/pkgs/[main/win-64|main/noarch|r/win-64|r/noarch|msys2/win-64|msys2/noarch], maybe @tswast knows whether this is a potential issue one should look into).

Likely that's because these packages are being fetched from [conda-forge](https://conda-forge.org) in the CI builds.

```
conda config --add channels conda-forge 
```

dhermes:
> Our funds aren't magnificent, so I'm setting the cap here at $20/participant.

Can I give my $20 to @schuderer?

tests/test_virtualenv.py
dhermes: SGTM

nox/virtualenv.py
```diff
@@ -104,13 +104,13 @@ def locate_using_path_and_version(version):
     if not version:
         return None
 
+    script = "import platform; print(platform.python_version())"
```
dhermes: Why lift this out of the `if` statement but not out of the function as a module global?

tests/test_virtualenv.py
```diff
@@ -23,6 +23,7 @@
 
 
 IS_WINDOWS = nox.virtualenv._SYSTEM == "Windows"
+RAISE_ERROR = "RAISE_ERROR"
```
dhermes: Very srs bznz!

tests/test_virtualenv.py
```diff
@@ -35,6 +36,47 @@ def factory(*args, **kwargs):
...
+                raise py.process.cmdexec.Error(1, 1, "", "", "")
+            else:
+                return sysexec_result
+
+        attrs = {
+            "sysexec.side_effect": mock_sysexec,
+            "__str__": mock.Mock(return_value=path),
+        }
+        mock_python = mock.Mock()
+        mock_python.configure_mock(**attrs)
```
dhermes: Ditto here, is there some reason we can't use `__spec__`?

tests/test_virtualenv.py
```diff
@@ -47,6 +48,47 @@ def factory(*args, **kwargs):
     return factory
 
 
+@pytest.fixture
+def make_mocked_interpreter_path():
+    def factory(path, sysexec_result):
+        def mock_sysexec(*_):
+            if sysexec_result == RAISE_ERROR:
```
dhermes: First of all, thank you for re-factoring these tests at my request. I know that can be a bit annoying after you've written some useful unit tests.

The core of my re-factoring request was that I was hoping we could ditch `if` statements for distinct test cases. But now there is an `if` statement here and on line 78, so my original hesitation is still there.

theacodes:
As long as we understand the *intent* of the tests we can always replace them later if needed.

theacodes:
> Can I give my $20 to @schuderer?

Of course.

dhermes:
> As long as we understand the *intent* of the tests we can always replace them later if needed.

So let's merge?

theacodes:
You're a maintainer. Go for it. :)

schuderer:
> > Can I give my $20 to @schuderer?
> 
> Of course.

Haha, I intended to suggest to give it all to @dhermes 😆 - But seriously, I don't really want the money, as this would essentially be a 4x refund of my donation from a few weeks ago. I'd prefer to keep it in nox's pot/re-donate it to the project. Some stickers would be awesome, though 😁  (if you set up a merch store with the nice art you got, I bet people would buy stickers/mugs/shirts, even more so if it supports the project).

nox/virtualenv.py
schuderer: I was aiming for style consistency with `locate_via_py`. I can put it inside the `if` block if it helps (would prefer not to make it module global, as it is the only time this string is used). Should I?

tests/test_virtualenv.py
schuderer: Yes. I'm exceptionally proud of this line. :) Well, I at least hope it's more explicit than some fake path string like `"c:\this\breaks\python.exe"`.

tests/test_virtualenv.py
schuderer: I understood that a plain `Mock()` would be preferable if possible (and `MagicMock(spec=("__str__"))` if it's not possible). I admit I'm unsure what exactly you are referring to by "Ditto" and "__spec__". I can change it to something like `MagicMock(spec=("__str__"))`? I might be confusing things here, also due to my lack of experience with `unittest.mock`, if in doubt, could you suggest a code example? Thanks!

tests/test_virtualenv.py
schuderer: Not annoying at all! It's an interesting challenge, and, yes, there were absolutely too many `if` statements in the original test.

But thanks to your request, the number of `if` statements got at least cut in half. Now, there are no `if`s in the tests themselves, so their logic should be much clearer. There are only mock/patch-related `if` statements left. Here, I really wanted to avoid copy-pasting the same mock/patch code over several tests. This mock/patch construction unfortunately needs to be a bit complex: a patched `sysfind` returns a mocked `path` to the python interpreter, which has a `__str__`, and whose `sysexec` method returns the version string.
On every one of these steps, something could go wrong and should be tested. The only way I see to test all these cases without conditionals, would be to mock all variants of these situations separately, which, IMO, would make the tests themselves harder to read.

I do get that each fixture creates a new potential source for bugs, and for me, the `if`s that are left are clear and primitive enough that it's worth the tradeoff. I just don't see how I could split this whole thing up into separate `if`-less tests (with or without extra fixtures) that would still be readable and not lead to increasing the amount of lines of test code manyfold. I am open to suggestions on how someone with more experience than me would solve this.

tests/test_virtualenv.py
```diff
@@ -47,6 +48,47 @@ def factory(*args, **kwargs):
     return factory
 
 
+@pytest.fixture
+def make_mocked_interpreter_path():
```
schuderer:
```suggestion
def make_mocked_interpreter_path():
    """Provides a factory to create a mocked ``path`` to a python interpreter.
    This mocked ``path`` provides a ``__str__`` , as well as a ``sysexec`` method
    which returns the value of the factory parameter ``sysexec_result``
    (which can be a version string or ``RAISE_ERROR``).
    """
```

tests/test_virtualenv.py
```diff
@@ -47,6 +48,47 @@ def factory(*args, **kwargs):
...
+        mock_python.configure_mock(**attrs)
+
+        return mock_python
+
+    return factory
+
+
+@pytest.fixture
+def patch_sysfind(make_mocked_interpreter_path):
+    def patcher(sysfind, only_find, sysfind_result, sysexec_result):
```
schuderer:
```suggestion
    """Provides a function to patch ``sysfind`` with parameters for tests related to
    locating a Python interpreter in the system ``PATH``.
    """
    def patcher(sysfind, only_find, sysfind_result, sysexec_result):
        """Returns an extended ``sysfind`` patch object for tests related to locating a
        Python interpreter in the system ``PATH``.
        
        Args:
            sysfind: The original sysfind patch object
            only_find (Tuple[str]): The strings for which ``sysfind`` should be successful,
                e.g. ``("python", "python.exe")``
            sysfind_result (Optional[str]): The ``path`` string to create the returned
                mocked ``path`` object with which will represent the found Python interpreter,
                or ``None``.
                This parameter is passed on to make_mocked_interpreter_path.
            sysexec_result (str): A string that should be returned when executing the
                mocked ``path`` object. Usually a Python version string.
                Use RAISE_ERROR to have ``sysexec`` fail.
                This parameter is passed on to make_mocked_interpreter_path.
        """
```

schuderer:
> > (Conda is installed, but says that colorlog and tox are not available from the current channels [https://repo.anaconda.com/pkgs/[main/win-64|main/noarch|r/win-64|r/noarch|msys2/win-64|msys2/noarch]](https://repo.anaconda.com/pkgs/%5Bmain/win-64%7Cmain/noarch%7Cr/win-64%7Cr/noarch%7Cmsys2/win-64%7Cmsys2/noarch%5D), maybe @tswast knows whether this is a potential issue one should look into).
> 
> Likely that's because these packages are being fetched from [conda-forge](https://conda-forge.org) in the CI builds.
> 
> ```
> conda config --add channels conda-forge 
> ```

@tswast Thanks, for your hint, I got further. It still fails with permission error: access is denied on `os.unlink(fullname)` when trying to delete contents of the `%USERTEMP%\pytest-of-<userid>\pytest-6\test_condaenv_create0\condaenv\DLLs\pyexpat.pyd` (gets triggered when `test_condaenv_create` calls `venv.create()` for the second time). Full disclosure: I'm running the conda_test-sessions from within the conda (base) environment as this is my only Python on Windows. Not sure if this is anything to look into (I'm not really planning to use this functionality specifically, but willing to try out what causes this with you/someone if it helps making it work with some edge cases - of which, apparently, I am one 😁). I can create an issue if it makes sense, and can provide more info/try stuff out, but can't promise that I can actually work on the code to fix this.

others: please excuse the off-topic discussion

tswast:
> It still fails with permission error: access is denied on venv.create() > if not self._clean_location() > shutil.rmtree(self.location) when trying to delete contents of the %USERTEMP%\pytest-of-<userid>\pytest-6\test_condaenv_create0\condaenv\DLLs\pyexpat.pyd (gets triggered when test_condaenv_create calls venv.create() for the second time)

Hmmm. I think there can be some problems with the virtualenv package that comes with conda. I installed https://github.com/conda-forge/virtualenv-feedstock in my environment. Then again, that's what should be happening already based on https://github.com/theacodes/nox/blob/master/requirements-conda-test.txt

I'll try on my local Windows laptop later today and see if I hit the same issues.

omry:
Let's merge this? this is getting in my way for adding windows CI to my project!
:)

schuderer:
@dhermes Agree with @omry, if I see it correctly, all review questions have been addressed (last point: please accept my documentation suggestions above if they are ok, or if accepting them does not work, tell me whether I should implement them). The conda discussion is off-topic and has been moved to another issue.

@omry In the meantime, feel free to use my workaround: At https://github.com/schuderer/mllaunchpad I created a `py.bat` in the project root which gets used by nox's lookup mechanism. Works nicely at my end. As soon as nox releases this PR, I can simply delete `py.bat` and everything should just keep working.

theacodes:
This LGTM, there might still be some improvements we could make to the tests but I don't see that as a reason to hold this any longer.

omry:
awesome, what is the release cadence for nox? when can I expect to try it out from the pip package?

theacodes:
It's whenever. I can try to cut a new release soon.

omry:
great, please do.
if you have doubts you can make it an rc and I can test it with --pre.

schuderer:
@theacodes Um, what about my test-doc suggestions above, wanted to add them if they improve things, but got no feedback. New PR?

theacodes:
Sure! Sorry about that.

omry:
@schuderer, I tried your py.bat hack and it didn't quiet for for me, I also have some slight reservation about the license it comes with (LGPL).
I also tried to contact you through http://www.schuderer.net/kontakt.shtml, not sure you are monitoring it.

I will try out the actual fix in nox as soon as it's released.

@theacodes, I am eagerly awaiting the new version.

schuderer:
@omry You can always install `nox` from source, for example using:
```bash
$ pip install git+https://github.com/theacodes/nox@master
```

You can also put `git+https://github.com/theacodes/nox@master` in your project's `requirements.txt` for the time being, and I think it should work on CI as well.

And sorry about missing your message! I've been hit by a lot of spam lately, and the filter might have become a little too eager. Unfortunately, I can't find your message in my spam folder; it must have been more than a month ago and already have been cleaned out... The `py.bat` hack (more fitting term than "workaround", actually :) ) is indeed kind of brittle. I would have changed this file's license to MIT if that's a problem.

theacodes:
I've also cut a new release!

omry:
@schuderer no need to change the license, I will test the new nox version and we can work on fixing it if it's not quiet there yet instead of on focusing on temporary solutions.

@theacodes awesome, I will test it soon.

PR #228, opened 2019-08-04
add shell autocomplete

cs01:
To try this out in bash run the following. Other shells also have limited support. See instructions in the docs in this PR.
```
pip install .  # from nox directory since a new dependency is required
eval "$(register-python-argcomplete nox)"
nox <tab>
nox --<tab>
nox -s <tab>
```

A sample of the autocompletions for all items that start with a `--`:
image

The  `_session_completer` function provides a list of sessions when hitting tab in `nox -s <tab>`. argcomplete provides `parsed_args` to this function, which is the Namespace argparse would have returned if the program were actually run. This corresponds almost directly to nox's global_config, and it works out of the box :smile: .

Note that this adds a new dependency to nox in setup.py, `argcomplete`.

fixes https://github.com/theacodes/nox/issues/227

cc @theacodes

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -313,3 +313,40 @@ This will create a ``noxfile.py`` based on the environments in your ``tox.ini``.
 
 .. _Generative environments: http://tox.readthedocs.io/en/latest/config.html#generating-environments-conditional-settings
 .. _substitutions: http://tox.readthedocs.io/en/latest/config.html#substitutions
+
+
+Shell Completion
+----------------
+Add the appropriate command to your shell's config file
+so that it is run on startup. You will likely have to restart
+or re-login for the autocompletion to start working.
+
+bash
+
+.. code-block:: console
+
+    eval "$(register-python-argcomplete nox)"
+
+zsh
+
+.. code-block:: console
+
+    # To activate completions for zsh you need to have
+    # bashcompinit enabled in zsh:
+    autoload -U bashcompinit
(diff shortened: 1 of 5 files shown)


cs01:
I see travis is failing because this drops below 100% coverage. I don't really have the time/desire to write tests for this at the moment, but I might at some point. I created the PR with the ability to add edits from maintainers if anyone would like to add them.

cs01:
@theacodes are you okay with the addition of a new dependency, [argcomplete](https://github.com/kislyuk/argcomplete)? 

Some good things about it
* It has no dependencies of its own
* It's actively maintained
* It works for multiple shells
* It's dynamically generated base on nox's arguments. This means the nox maintainers don't have to continuously update and test custom shell completion scripts.

Downside
* Additional dependency

Other options
* Add an extra such as `nox[completions]` to nox's setup.py so it's not included in the base `nox` package. This requires a little extra overhead for those that want shell completions, but avoids forcing that requirement on *all* nox installations.

theacodes:
I'm happy taking it on if it has no other dependencies and is relatively
stable.

cs01:
Great! We should be good, and also have the extra safety of limiting the versions nox requires (>=1.9.4, <2.).

I will work on getting this test passing now.

cs01:
Tests have been added and CI is now passing.

theacodes:
This is rad, thank you @cs01!

PR #229, opened 2019-08-05
immediately exit if unknown arguments are passed

cs01:
This PR adds changes to raise an error if an unknown argument is passed.

### before (all tests would run):
```
nox lint 
> nox 
nox > Running session tests-3.5
nox > Session tests-3.5 skipped: Python interpreter 3.5 not found.
nox > Running session tests-3.6
...
```
### after (error is raised):
```
>> nox lint
usage: nox [-h] [--version] [-l] [-s [SESSIONS [SESSIONS ...]]] [-k KEYWORDS]
           [-r] [--no-reuse-existing-virtualenvs] [-f NOXFILE]
           [--envdir ENVDIR] [-x] [--no-stop-on-first-error]
           [--error-on-missing-interpreters]
           [--no-error-on-missing-interpreters] [--error-on-external-run]
           [--no-error-on-external-run] [--install-only] [--report REPORT]
           [--non-interactive] [--nocolor] [--forcecolor]
           ...
nox: error: Unrecognized argument(s): 'lint'

>> nox -s lint
nox > Running session lint
nox > Creating virtualenv using python3.7 in .nox/lint
...
```

fixes https://github.com/theacodes/nox/issues/226
cc @theacodes

### DIFF
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -20,7 +20,7 @@ and surfaced in documentation."""
 import argparse
 import collections
 import functools
-import sys
+
 
 Namespace = argparse.Namespace
 ArgumentError = argparse.ArgumentError
@@ -68,7 +68,7 @@ class Option:
         finalizer_func=None,
         default=None,
         hidden=False,
-        **kwargs,
+        **kwargs
     ):
         self.name = name
         self.flags = flags
@@ -143,7 +143,7 @@ def make_flag_pair(name, enable_flags, disable_flags, **kwargs):
         *enable_flags,
         noxfile=True,
         merge_func=functools.partial(flag_pair_merge_func, name, disable_name),
-        **kwargs,
+        **kwargs
     )
 
     kwargs["help"] = "Disables {} if it is enabled in the Noxfile.".format(
@@ -233,22 +233,8 @@ class OptionSet:
 
     def parse_args(self):
         parser = self.parser()
-        all_args = sys.argv[1:]
-        if "--" in all_args:
-            split_index = all_args.index("--")
-            nox_args = all_args[0:split_index]
-            posargs = all_args[split_index + 1 :]
-        else:
-            nox_args = all_args
-            posargs = []
+        args = parser.parse_args()
 
-        args = parser.parse_args(nox_args)
-        if args.posargs:
-            # posargs should not be present since we explicitly only parsed
-            # values before the '--'
-            parser.error(f"Unrecognized argument(s): '{' '.join(args.posargs)}'")
-
-        args.posargs = posargs
         try:
             self._finalize_args(args)
         except ArgumentError as err:
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -20,7 +20,7 @@ and surfaced in documentation."""
 import argparse
 import collections
 import functools
-
+import sys
 
 Namespace = argparse.Namespace
 ArgumentError = argparse.ArgumentError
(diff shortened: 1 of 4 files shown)


nox/_option_set.py
```diff
@@ -233,8 +233,22 @@ def _finalize_args(self, args):
...
+            posargs = all_args[split_index + 1 :]
+        else:
+            nox_args = all_args
+            posargs = []
+
+        args = parser.parse_args(nox_args)
+        if args.posargs:
+            # posargs should not be present since we explicitly only parsed
+            # values before the '--'
+            parser.error(f"Unrecognized argument(s): '{' '.join(args.posargs)}'")
```
theacodes: Would it actually be possible to move this logic into the posargs finalizer?

cs01: Yeah now that you mention it, it does look like it makes more sense there.

nox/_options.py
```diff
@@ -97,11 +97,30 @@ def _color_finalizer(value, args):
...
+        )
+
+    dash_index = posargs.index("--")
+    if dash_index != 0:
+        unexpected_posargs = posargs[0:dash_index]
+        raise _option_set.ArgumentError(
+            None, "Unknown argument(s) '{}'.".format(" ".join(unexpected_posargs))
+        )
+
+    # fmt: off
```
theacodes: Interesting, why did we have to turn black off for this?

theacodes:
> This looks great - just one last question before merging. :)

Of course, we'll need CI to pass. Let me know if you run into any trouble there.

nox/_options.py
cs01: flake8 competes with black here. If I let black run, this is what flake8 does:
```
nox > flake8 nox tests
nox/_options.py:120:34: E203 whitespace before ':'
nox > Command flake8 nox tests failed with exit code 1
nox > Session lint failed.
```

I could add E203 to the ignore list in flake8 if you want (that would be my personal preference).

cs01:
> > This looks great - just one last question before merging. :)
> 
> Of course, we'll need CI to pass. Let me know if you run into any trouble there.

I'm having trouble getting the unit test to pass. It did start failing, which is a good thing. It means the new stricter check is working! I just can't figure out how to assert the proper error was raised. Any help would be appreciated.

cs01:
Tests are now passing. I also updated the flake8 rule to comply with black's formatting (and removed the `fmt: off`.

tests/test_main.py
```diff
@@ -168,18 +168,20 @@ def test_main_session_from_nox_env_var(monkeypatch, env, sessions):
...
-    with mock.patch("nox.workflow.execute") as execute:
-        execute.return_value = 0
 
-        # Call the main function.
-        with mock.patch.object(sys, "exit") as exit:
-            nox.__main__.main()
-            exit.assert_called_once_with(0)
-        assert execute.called
+    with pytest.raises(
+        AttributeError, match="object has no attribute 'color'"
```
theacodes: I don't understand this test anymore. What's happening here?

cs01: The `_exit` function was a no-op, so instead of exiting, nox continued executing `main()` until it hit the next exception which was an attribute error. Since the arguments were not parsed, `args` didn't have the `color` attribute. That is pretty fragile and unintuitive, so I updated with a mocked exit that raises an exception so the unit test can catch it.

theacodes:
Thank you!

PR #230, opened 2019-08-05
add mypy to lint

cs01:
Type checking has lots of benefits for a codebase. I'm not sure if the lack of type checks in nox is intentional, but since nox supports Python 3.5+ it can use types. I went ahead and added mypy to the lint session and ignored all the existing mypy errors so it all passes. 

Future PRs can add more types to existing functions and gradually remove the ignored type checks. Of course if you aren't interested in types, feel free to decline this PR but it was something I would have benefited from as a newcomer trying to navigate the codebase, so I thought I'd put up a PR 😄 .

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -15,7 +15,7 @@
 import os
 import sys
 
-import py
+import py  # type: ignore
 
 from nox.logger import logger
 from nox.popen import popen
--- a/nox/logger.py
+++ b/nox/logger.py
@@ -14,12 +14,12 @@
 
 import logging
 
-from colorlog import ColoredFormatter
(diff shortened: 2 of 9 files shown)


theacodes:
> I'm not sure if the lack of type checks in nox is intentional, but since nox supports Python 3.5+ it can use types.

We were originally a 2+3 project, so that explains the lack of types. We can start adding them for sure. :)

Let me know when you think this is good to merge?

cs01:
Awesome, yeah it's good to go! The call to mypy passed CI, that's all we needed.

theacodes:
Thanks, @cs01!

Issue #123, opened 2018-08-27
reuse_existing should check the interpreter


dhermes:
If `reuse_existing=True` when creating a `Virtualenv`, but the interpreter has changed, then the environment should still be re-created.

theacodes:
Interesting, wouldn't that change the name of the virtualenv folder?

dhermes:
@theacodes I'm not 100% sure, but I think "not always" is the answer.

theacodes:
Cool, makes sense for us to check it then.

theacodes:
I'm not 100% how to do this yet. We don't ever know the full path to the interpreter, we just figure out enough info to get virtualenv to figure it out. I could try invoking the interpreter.

dhermes:
Yes. You can get the actual interpreter via:

```python
original = session.run("python", "-c", "import sys; print(sys.real_prefix)")
```

theacodes:
Invoking the interpreter doesn't quite get me any closer. We could invoke our resolved interpreter and invoke the one in the virtualenv and compare. Still not 100% sure on that.

dhermes:
Yes that's what you want to do.

theacodes:
From #231, we should also test the *type* of environment.

PR #231, opened 2019-08-05
add venv as an option for venv_backend

cs01:
### Summary

This PR adds support for a different backend, `venv`. It adds a keyword argument to the `VirtualEnv` class, which defaults to `False` and is only true when the nox session has `venv_backend=True`.

It also updates the resolution of the Python path, since using `venv` alone does not fix issues when creating a venv from within a virtual environment.

The path resolution is modified to determine if a virtual environment is being used. If so, it tries to find the desired Python interpreter version in the `base_prefix` rather than anywhere on the system's PATH. Note that Windows was not tested, and probably will require some more changes.

### Test Plan

```
# Create and enter virtual environment, then
cd ~/git/nox
pip install -e .
cd ~/git/pipx
# modify pipx's noxfile.py to have `venv_backend='venv'` for the unittests session
nox -s unittests-3.6  # tests pass
```

https://docs.python.org/3/library/sys.html#sys.prefix

> base_prefix: Set during Python startup, before site.py is run, to the same value as prefix. If not running in a virtual environment, the values will stay the same; if site.py finds that a virtual environment is in use, the values of prefix and exec_prefix will be changed to point to the virtual environment, whereas base_prefix and base_exec_prefix will remain pointing to the base Python installation (the one which the virtual environment was created from).

https://docs.python.org/3/library/sys.html#sys.base_prefix

Tox's implementation for venv support is included as a plugin in an additional package: https://github.com/tox-dev/tox-venv/blob/master/src/tox_venv/hooks.py

--- 

fixes https://github.com/theacodes/nox/issues/199
cc @pfmoore

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+from pathlib import Path
 import platform
 import re
 import shutil
@@ -174,6 +175,26 @@ class CondaEnv(ProcessEnv):
         return True
 
 
+def resolve_real_python_outside_venv(desired_intepreter: str) -> str:
+    """Return path to the real Python installation based
+
+    See also:
+    https://docs.python.org/3/library/sys.html#sys.prefix
+    https://docs.python.org/3/library/sys.html#sys.base_prefix
+    """
+    interpreter = (
+        desired_intepreter
+        if desired_intepreter
+        else f"{sys.version_info.major}.{sys.version_info.minor}"
+    )
+    python_str = f"python{interpreter}"  # i.e. python3.6
+
+    base_python_installation = Path(sys.base_prefix) / "bin" / python_str
+    if base_python_installation.is_file():
+        return str(base_python_installation)
+    raise InterpreterNotFound(interpreter)
+
+
 class VirtualEnv(ProcessEnv):
     """Virtualenv management class.
 
@@ -196,12 +217,15 @@ class VirtualEnv(ProcessEnv):
 
     is_sandboxed = True
 
-    def __init__(self, location, interpreter=None, reuse_existing=False):
+    def __init__(
+        self, location, interpreter=None, reuse_existing=False, *, venv: bool = False
+    ):
         self.location_name = location
         self.location = os.path.abspath(location)
         self.interpreter = interpreter
         self._resolved = None
         self.reuse_existing = reuse_existing
+        self.venv_or_virtualenv = "venv" if venv else "virtualenv"
         super(VirtualEnv, self).__init__()
 
     _clean_location = _clean_location
@@ -212,6 +236,7 @@ class VirtualEnv(ProcessEnv):
 
         Based heavily on tox's implementation (tox/interpreters.py).
         """
+
         # If there is no assigned interpreter, then use the same one used by
         # Nox.
         if isinstance(self._resolved, Exception):
@@ -220,8 +245,13 @@ class VirtualEnv(ProcessEnv):
         if self._resolved is not None:
             return self._resolved
 
+        currently_in_virtual_environment = sys.prefix != sys.base_prefix
+
         if self.interpreter is None:
-            self._resolved = sys.executable
+            if currently_in_virtual_environment:
+                self._resolved = resolve_real_python_outside_venv(self.interpreter)
+            else:
+                self._resolved = sys.executable
             return self._resolved
 
         # Otherwise we need to divine the path to the interpreter. This is
@@ -238,7 +268,10 @@ class VirtualEnv(ProcessEnv):
             cleaned_interpreter = "python{}".format(xy_version)
 
         # If the cleaned interpreter is on the PATH, go ahead and return it.
-        if py.path.local.sysfind(cleaned_interpreter):
+        if currently_in_virtual_environment and _SYSTEM != "Windows":
+            self._resolved = resolve_real_python_outside_venv(self.interpreter)
+            return self._resolved
+        elif py.path.local.sysfind(cleaned_interpreter):
             self._resolved = cleaned_interpreter
             return self._resolved
 
@@ -275,21 +308,26 @@ class VirtualEnv(ProcessEnv):
             return os.path.join(self.location, "bin")
 
     def create(self):
-        """Create the virtualenv."""
+        """Create the virtualenv or venv."""
         if not self._clean_location():
             logger.debug(
-                "Re-using existing virtualenv at {}.".format(self.location_name)
+                "Re-using existing virtual environment at {}.".format(
+                    self.location_name
+                )
             )
             return False
 
-        cmd = [sys.executable, "-m", "virtualenv", self.location]
-
         if self.interpreter:
-            cmd.extend(["-p", self._resolved_interpreter])
+            cmd = [self._resolved_interpreter]
+        else:
+            cmd = [sys.executable]
+        cmd += ["-m", self.venv_or_virtualenv, self.location]
 
         logger.info(
-            "Creating virtualenv using {} in {}".format(
-                os.path.basename(self._resolved_interpreter), self.location_name
+            "Creating virtual environment ({}) using {} in {}".format(
+                self.venv_or_virtualenv,
+                os.path.basename(self._resolved_interpreter),
+                self.location_name,
             )
         )
         nox.command.run(cmd, silent=True, log=False)
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import os
-from pathlib import Path
 import platform
 import re
 import shutil
@@ -175,26 +174,6 @@ class CondaEnv(ProcessEnv):
         return True
 
 
-def resolve_real_python_outside_venv(desired_intepreter: str) -> str:
-    """Return path to the real Python installation based
-
-    See also:
-    https://docs.python.org/3/library/sys.html#sys.prefix
-    https://docs.python.org/3/library/sys.html#sys.base_prefix
-    """
-    interpreter = (
-        desired_intepreter
-        if desired_intepreter
-        else f"{sys.version_info.major}.{sys.version_info.minor}"
-    )
-    python_str = f"python{interpreter}"  # i.e. python3.6
-
-    base_python_installation = Path(sys.base_prefix) / "bin" / python_str
-    if base_python_installation.is_file():
-        return str(base_python_installation)
-    raise InterpreterNotFound(interpreter)
-
-
 class VirtualEnv(ProcessEnv):
     """Virtualenv management class.
 
@@ -217,9 +196,7 @@ class VirtualEnv(ProcessEnv):
 
     is_sandboxed = True
 
-    def __init__(
-        self, location, interpreter=None, reuse_existing=False, *, venv: bool = False
-    ):
+    def __init__(self, location, interpreter=None, reuse_existing=False, *, venv=False):
         self.location_name = location
         self.location = os.path.abspath(location)
         self.interpreter = interpreter
@@ -236,7 +213,6 @@ class VirtualEnv(ProcessEnv):
 
         Based heavily on tox's implementation (tox/interpreters.py).
         """
-
         # If there is no assigned interpreter, then use the same one used by
         # Nox.
         if isinstance(self._resolved, Exception):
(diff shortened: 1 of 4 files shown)


nox/virtualenv.py
```diff
@@ -275,21 +308,26 @@ def bin(self):
...
+                    self.location_name
+                )
             )
             return False
 
-        cmd = [sys.executable, "-m", "virtualenv", self.location]
-
         if self.interpreter:
-            cmd.extend(["-p", self._resolved_interpreter])
+            cmd = [self._resolved_interpreter]
```
pfmoore: This change means that if virtualenv is used, it will now need to be installed *in the resolved interpreter*, whereas the current code only requires virtualenv to be installed in the current interpreter. It's not clear to me whether this would be a problem in practice, but it is a change in behaviour and if it's to be made, it should be done deliberately and not "by accident".

nox/virtualenv.py
```diff
@@ -275,21 +308,26 @@ def bin(self):
...
 
     def create(self):
-        """Create the virtualenv."""
+        """Create the virtualenv or venv."""
         if not self._clean_location():
             logger.debug(
-                "Re-using existing virtualenv at {}.".format(self.location_name)
+                "Re-using existing virtual environment at {}.".format(
+                    self.location_name
+                )
```
pfmoore: Is there a risk here that the existing environment used virtualenv, but the request was to use venv? If so, would this result in the caller getting the wrong type of environment? This is a real edge case, and probably isn't a major consideration, but it's worth noting.

nox/virtualenv.py
```diff
@@ -174,6 +175,26 @@ def create(self):
         return True
 
 
+def resolve_real_python_outside_venv(desired_intepreter: str) -> str:
+    """Return path to the real Python installation based
```
pfmoore: Incomplete sentence?

nox/virtualenv.py
```diff
@@ -220,8 +245,13 @@ def _resolved_interpreter(self):
         if self._resolved is not None:
             return self._resolved
 
+        currently_in_virtual_environment = sys.prefix != sys.base_prefix
+
         if self.interpreter is None:
-            self._resolved = sys.executable
+            if currently_in_virtual_environment:
+                self._resolved = resolve_real_python_outside_venv(self.interpreter)
```
pfmoore: As it stands, this has no hope of working on Windows. At this point, we're calling `resolve_real_python_outside_venv(None)` (side note, it might be better to explicitly use `None`, rather than have the reader need to track back to the if statement to determine that `self.interpreter` is always `None` at this point). There's no check for `_WINDOWS` here (unlike the call below) so the fact that `resolve_real_python_outside_venv` doesn't (yet) support Windows will cause immediate breakage.

nox/virtualenv.py
```diff
@@ -238,7 +268,10 @@ def _resolved_interpreter(self):
             cleaned_interpreter = "python{}".format(xy_version)
 
         # If the cleaned interpreter is on the PATH, go ahead and return it.
-        if py.path.local.sysfind(cleaned_interpreter):
+        if currently_in_virtual_environment and _SYSTEM != "Windows":
```
pfmoore: You're ignoring `cleaned_interpreter` here, which seems weird. Why go to the trouble of calculating it and then not use it?

nox/virtualenv.py
```diff
@@ -174,6 +175,26 @@ def create(self):
         return True
 
 
+def resolve_real_python_outside_venv(desired_intepreter: str) -> str:
+    """Return path to the real Python installation based
+
+    See also:
+    https://docs.python.org/3/library/sys.html#sys.prefix
+    https://docs.python.org/3/library/sys.html#sys.base_prefix
+    """
```
pfmoore: This definitely won't work on Windows. There are various implementations of this type of "locate Python" algorithm on the web, including a reusable library version at https://github.com/sarugaku/pythonfinder.

Unfortunately, reusable code for this seems hard to locate (hence the reason so many people write their own) but IMO we should use a library if at all possible (pythonfinder is the one I've heard of, but I have no problem if someone recommends a better one).

nox/virtualenv.py
cs01: Agreed

nox/virtualenv.py
cs01: Good catch, I didn't intentionally ignore it

nox/virtualenv.py
cs01: pythonfinder looks like it can replace the entire existing nox function unless I'm missing something. Is that correct?

cs01: I replaced all the code with pythonfinder but it isn't able to find any installations. I'll keep trying with it. If it works it will simplify the code tremendously.

theacodes:
I would definitely want to make the pythonfinder stuff a separate PR, if
that's okay with y'all.

Relatedly, I'm a bit wary about taking on new dependencies in Nox. It
definitely warrants a bit of discussion.

nox/virtualenv.py
cs01: Whoops, that was a remnant of my hack to get pipx tests working earlier. I'll add the `-p` flag back in if virtualenv is being used. I'll fix it.

nox/virtualenv.py
pfmoore: I've not used pythonfinder myself, I just knew about it from other projects. So I've no experience with it I'm afraid. I'll try to have a play with it myself, in case I get any ideas.

pfmoore:
I'm completely fine with the pythonfinder thing being a separate discussion. However, getting `resolve_real_python_outside_venv` to work without it is non-trivial (the existing code doesn't support Windows at all). I'll fish out some known-working code that does the job - both tox or virtualenv have this IIRC (and I have a suspicion I contributed the code in both cases!) Let me see what I can find.

But I do think that in the long term, duplicating this same bit of code in many projects is a bad idea...

cs01:
@pfmoore I'm pretty sure this is tox's: https://github.com/tox-dev/tox-venv/blob/master/src/tox_venv/hooks.py

theacodes:
If we land the PR to switch to pythonfinder *first*, does this PR become
easier/less complex?

cs01:
@theacodes Assuming pythonfinder does what I think it does, yes. 

This PR is really doing two things: adding the venv_backend option of venv, and updating Python resolution to make venv creation from within a virtual environment work. It makes sense to review the two changes separately, I can separate it out.

pfmoore:
> I'm pretty sure this is tox's

I'm not sure that's all of it. A key part is `tox_get_python_executable`, which is defined in the `tox.interpreters` subpackage). To get the available interpreters on Windows you need to look in the registry (PEP 514).

As a simpler workaround, you can let the launcher do the work for you, and call `py -X.Y -c "import sys; print(sys.executable)"` in a subprocess, but you need to be careful with things like setting `PYTHONIOENCODING` so that you support full Unicode paths.

The reason I'm in favour of a library solution is that it's really tricky to get all the edge cases correct. Of course, users can always just supply the full executable path, so it's not like it's critical, but it's not a good UX if the software can't find a Python installation you know is there.

> updating Python resolution to make venv creation from within a virtual environment work

Yeah, mixing virtualenvs and venvs is really messy. At this point I *think* you can safely create a virtualenv from within an active venv (see the discussions around https://github.com/pypa/virtualenv/pull/1345 for the ugly details) but I don't think creating a venv from within an active virtual environment (either venv or virtualenv) is even supported. So getting this fixed would be good (for general reliability, even if there's no specific issue it's causing right now).

pfmoore:
Something like the following should work for Windows, for what it's worth. Lightly tested, it works on my PC but I only have one Python version installed, so I couldn't test much...

```python
import os
import subprocess

def get_python(major, minor=None):
    env = os.environ.copy()
    env['PYTHONIOENCODING'] = "utf-8"
    version = "-" + major
    if minor:
        version += "." + minor
    try:
        proc = subprocess.run(
            ["py", version, "-c", "import sys; print(sys.executable)"],
            env=env,
            capture_output=True
        )
    except Exception:
        # Just say we couldn't find anything if we get any sort of error
        return None
    if proc.returncode != 0:
        return None
    return proc.stdout.decode("utf-8").strip()

if __name__ == '__main__':
    import sys
    print(get_python(*sys.argv[1:]))
```

pfmoore:
Whoa! I just looked at the list of dependencies for pythonfinder:

```
install_requires =
    attrs
    cached-property
    click
    crayons
    six
    packaging
    vistir[spinner]>=0.4
```

That's *way* too much for the job it's doing. Installing it pulled in 15 other packages, for a total of about 14M of data.

At this point, I'd say let's *not* use pythonfinder. That's just silly :-(

nox/virtualenv.py
cs01: Yes there is. Agreed this is an edge case. I suppose there could be a call made to identify whether it's a virtualenv or venv and to assert as necessary before continuing.

theacodes: We also have an outstanding bug to make sure the `-r` checks the interpreter version as well. We might be able to just add a note to that issue that checking the backend should be part of the check as well. #123.

nox/virtualenv.py
```diff
@@ -275,21 +276,25 @@ def bin(self):
...
+                    self.location_name
+                )
             )
             return False
 
-        cmd = [sys.executable, "-m", "virtualenv", self.location]
+        cmd = [sys.executable, "-m", self.venv_or_virtualenv, self.location]
 
-        if self.interpreter:
+        if self.interpreter and self.venv_or_virtualenv == "virtualenv":
```
theacodes: What happens if they request a *different* interpreter version than the one Nox is installed on?

cs01: Ah I think I see a bug. 

In the case of venv we should call
```
[self._resolved_interpreter, 'venv', '-m', 'self.location']
```
rather than use `sys.executable` since you can't specify the python interpreter to use in venv (it just uses the Python you create the venv with).

In the case of virtualenv it will remain unchanged
```
[sys.executable, 'virtualenv', '-m', 'self.location', '-p', self._resolved_interpreter]
```

cs01:
> Whoa! I just looked at the list of dependencies for pythonfinder:
> 
> ```
> install_requires =
>     attrs
>     cached-property
>     click
>     crayons
>     six
>     packaging
>     vistir[spinner]>=0.4
> ```
> 
> That's _way_ too much for the job it's doing. Installing it pulled in 15 other packages, for a total of about 14M of data.
> 
> At this point, I'd say let's _not_ use pythonfinder. That's just silly :-(

Just FYI I filed an issue in pythonfinder on this: https://github.com/sarugaku/pythonfinder/issues/78

cs01:
Got the tests passing :tada: .

So to summarize, this PR adds support for `venv` as an alternate backend in a nox session. It does not modify the logic when resolving the interpreter, which is required to create a new venv from within a venv. That can be done in a different PR, possibly with `pythonfinder`, depending on how https://github.com/sarugaku/pythonfinder/issues/78 evolves.

theacodes:
Thank you, @cs01 and @pfmoore!

Does this need a follow-up PR for documentation as well, or do we think it's sufficiently advanced enough that most users won't need it?

cs01:
> Thank you, @cs01 and @pfmoore!
> 
> Does this need a follow-up PR for documentation as well, or do we think it's sufficiently advanced enough that most users won't need it?

I don't really have a feeling for how many other people want to use this. AFAIK the only reason you'd want it is to test a program that creates virtual environments. So that would include pipx, pipenv, poetry, virtualenv, venv, and maybe some others. Because of this, I don't know if there is much value added to using the venv backend without the improved Python resolution, so I was planning on waiting until the improved Python resolution was added before documenting it.

theacodes:
Sounds good. Thank you.

gaborbernat:
Hmmm, I've been working on making virtualenv more like venv, and add python discovery part of it - see https://github.com/gaborbernat/virtualenv/blob/rewrite/src/virtualenv/interpreters/discovery/__init__.py 🤔 probably will be released in a month or so 🤔 so this entire ticket should self solve itself then 🤔

cs01:
venv is now a valid backend for nox, so nox can create venvs, but interpreter resolution in nox still needs to be improved since it still defaults to the python in the venv. 

Are you saying that in the future, nox can use virtualenv but still find the base interpreter from there (I think currently it can only do this from venv), or that the default interpreter in virtualenv will be the base interpreter and therefore able to create a new venv/virtualenv from there?  (This is the PR, open issue is at https://github.com/theacodes/nox/issues/233)

gaborbernat:
virtualenv's future discovery mechanism will always let you know the base interpreter too (for venv/old virtualenv/host python)👍

cs01:
Right now, nox has a `@property` called `_resolved_interpreter`. Just want to confirm I am understanding the new workflow: In the future we will be importing virtualenv into nox's `virtualenv.py` module and using a new method that allows us to get the base interpreter?

gaborbernat:
Not sure how this will integrate into nox; but virtualenv will contain the logic of interpreter discovery (with option to also access the host interpreter, not just the immediate discovered one). tox will definitely use this and no longer implement its own version of this functionality, nox might or might not choose to do so, 🤷‍♂ it's up to @theacodes

PR #241, opened 2019-08-25
add type hints to tasks.py

cs01:
Generate type hints database from types used during runtime:
```
pip install monkeytype
monkeytype run nox -s tests-3.6
```

See modules that can have types applied:
```
monkeytype list-modules
```

Apply types
```
monkeytype apply <module>
```

This exposed a circular dependency between session.py and manifest.py which was resolved in this PR. It ended up causing many LOC to be changed, but the change was cut+pasting code.

There are still several modules that haven't had types added yet since this PR got quite large.

Note that there are no type errors, unit tests did not change, and test coverage is still 100%.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -245,10 +245,10 @@ If you run ``nox --list``, you'll see that this generates the following set of s
 
 .. code-block:: console
 
-    * tests(django='1.9', database='postgres')
-    * tests(django='2.0', database='mysql')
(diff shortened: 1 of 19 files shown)


theacodes:
honestly we can just disable flake8's import order checking.

cs01:
Merged master into this PR. Now the diff is much smaller and only affects types in tasks.py.

theacodes:
Rad, thanks!

PR #242, opened 2019-08-27
replace flake8 sorter with isort

cs01:
There is no easy way to automatically satisfy flake8's linting requirements for sort order. Using isort there is.

I ran 
```
isort --recursive nox
```
to automatically sort imports. Then I updated noxfile to test with isort rather than flake8.


theacodes:
I'm in favor of removing the isort call from the lint session.

theacodes:
tbh we should pin all of the lint/format tools so that contributors don't
get tangled in unrelated style updates.

PR #249, opened 2019-09-06
add types to _option_set.py

cs01:
Ran `monkeytype apply nox._option_set` then added additional types as needed.

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -130,7 +130,7 @@ def _session_completer(
     module = load_nox_module(global_config)
     manifest = discover_manifest(module, global_config)
     filtered_manifest = filter_manifest(manifest, global_config)
-    if isinstance(filtered_manifest, int):  # pragma: no cover
+    if isinstance(filtered_manifest, int):
         return []
     return [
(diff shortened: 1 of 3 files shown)


nox/_options.py
```diff
@@ -122,11 +123,15 @@ def _posargs_finalizer(value, args):
...
 
-def _session_completer(prefix, parsed_args, **kwargs):
+def _session_completer(
+    prefix: str, parsed_args: argparse.Namespace, **kwargs
+) -> List[str]:
     global_config = parsed_args
     module = load_nox_module(global_config)
     manifest = discover_manifest(module, global_config)
     filtered_manifest = filter_manifest(manifest, global_config)
+    if isinstance(filtered_manifest, int):  # pragma: no cover
```
theacodes: If this is never hit, we should just remove the branch.

cs01: mypy complains that it might be an int, so it's there as a type refinement since `filter_manifest` can return an `int`. I can look into adding a test case to cover this line.

theacodes:
SG, thanks for doing this!

theacodes:
Thank you, @cs01!

PR #258, opened 2019-10-15
appveyor.yml: Order Pythons in descending version

jayvdb:
This ordering helps identify when a problem occurs only
in older versions of Python, as AppVeyor fail-fast means
failures in the early jobs prevent seeing if latter jobs
would have succeeded.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,7 +65,7 @@ install:
   # compiled extensions and are not provided as pre-built wheel packages,
   # pip will build them from source using the MSVC compiler matching the
   # target Python version and architecture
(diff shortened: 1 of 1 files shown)


jayvdb:
New release of pip looks like it might be related https://github.com/pypa/pip/issues/6993

jayvdb:
The second commit appears to be solving the current CI problems.  I can split it off to a separate PR if the first commit isnt also seen as insta-mergable.

jayvdb:
Upstream issue is https://github.com/pypa/pip/issues/5599

PR #265, opened 2019-11-22
install cache to avoid frequent pip installs...

toppk:
this creates a feature to avoid running pip all the time.  This does it by creating a state file inside the root of the venv to cache when the last good run of the pip command (with those arguments) was run.

For example a cache file will look like this:
```
$ more .nox/test-3-6/.nox_install_cache 
{"[\"--upgrade\",\"-r\",\"test-requirements.txt\"]":1574446974}
```
There are three ways to set this 

- globally on the command line: `    nox -install-cache 10`
- globally inside the noxfile.py:  `  nox.options.install_cache="10"`
- per individual session invocations:      ` session.install("--upgrade", "-r", "test-requirements.txt", nox_install_cache="10")`

I'm sure there's some tidying to do before this is accepted, my list:

1. Where should the code for the InstallCache object reside (currently in sessions.py)?
2. How to preserve the InstallCache instance  (currently creating it from scrach every `Session.install()` invocation)
3. surely there are more error case to code for (corrupt cache files, invalid setting types, etc).

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -209,6 +209,13 @@ options.add_options(
         help="Re-use existing virtualenvs instead of recreating them.",
     ),
     _option_set.Option(
+        "install_cache",
+        "--install-cache",
+        noxfile=True,
+        group="secondary",
+        help="How often to skip succesful install calls (in minutes)",
+    ),
+    _option_set.Option(
         "noxfile",
         "-f",
         "--noxfile",
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import defaultdict
 import enum
 import hashlib
+import json
 import os
 import re
 import sys
+import time
 import unicodedata
 
 import nox.command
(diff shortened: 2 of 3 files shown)


tswast:
What about having a flag to put pip in offline mode https://stackoverflow.com/a/51646354/101923 This could accomplish a similar goal of speeding things up without introducing a cache mechanism to nox.

toppk:
Perhaps cache mechanism is not the best term for the feature.  It's more of a timeout option to avoid calling the same pip commands in the same virtualenvs over and over.   I suppose setting up a auto caching pypi proxy would be a option here rather than offline mode, balancing performance and hands off use. 

As it is, this patch brings me from 25 seconds nox run time down to 6 seconds.  Which helps me a great deal.

toppk:
On the otherhand, if upstream isn't interested in this, I could rework this to be a "plugin" of sorts that someone could put into their noxfile.py.

cs01:
I am a +1 on this. 

One of the biggest pain points for me with nox is when I'm iterating on something and want to continuously run pytest or something similar. My velocity gets reduced by nox's repeated pip installations and I will often end up side-stepping it by running the command from the virtual environment directly. I'll do something like `.nox/tests-3-7/bin/pytest`. 

Even if I run nox with the `-r` flag to reuse and existing virtual environment, the slowdown time is still noticeable to the point where it pushes me to invoke commands directly.

theacodes:
I'm going to close this PR due to inactivity. That's not to say I'm not interested in merging it! If you're still up for it just reply and I'll re-open the PR. :)

PR #267, opened 2019-12-06
Use stdout.isatty to finalize color

### LINKED ISSUE #266: Colorized output is not disabled when output is redirected
**Describe the bug**
Redirecting nox output does not correctly determine that stdout is not a tty and adds color control codes to the output.

**How to reproduce**
Given a noxfile that defines some targets, e.g. unit tests, run
`cd path/to/repo && python -m nox -s &> output-log.txt`

**Expected behavior**
The contents of `output-log.txt` should be plain ASCII (or UTF-8) and not contain cursor control codes.

software-dov:
Currently checks stdin
Fixes #266

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -96,7 +96,7 @@ def _color_finalizer(value, args):
     if args.nocolor is True:
         return False
 
-    return sys.stdin.isatty()
+    return sys.stdout.isatty()
 
 
 def _posargs_finalizer(value, args):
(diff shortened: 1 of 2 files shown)


theacodes:
Thank you! This looks reasonable, but you probably need to update the associated test. Let me know if you need any help with that.

software-dov:
Fixed the test, had some trouble running the full unit test suite locally because of an issue with venv.

theacodes:
Thank you!

PR #270, opened 2019-12-31
Support for Python 3.8

lewoudar:
Hi everyone,
Since I work on my projects using nox and python3.8, I think it's wise to point out that nox also works with python3.8, hence my pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 matrix:
   include:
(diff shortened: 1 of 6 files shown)


theacodes:
Thank you!

lewoudar:
You are welcome :)
Just for my culture, why coverage failed on my laptop but not on CI providers? and how conda tests are checked because it is not obvious to me where they are tested in CI configuration files

PR #272, opened 2020-01-16
Use a more accurate error message if loading of noxfile runs into error

jaikiran:
I had a `noxfile.py` which looked something like:

```
import os
os.makedirs(some_already_existing_path)
....
# all other nox sessions
```
While running this with `nox`, I kept running into an odd:

```
nox > Noxfile /home/me/noxfile.py not found
```
Pretty sure the file was there and I also checked the file permissions and other usual suspects. Couldn't make much sense out of it so looked into the source code of `nox` and realized that this error message is inaccurate. In fact, this has been reported as an issue at https://github.com/theacodes/nox/issues/248.

The commit in this PR, changes the log message to more accurately describe what went wrong and also include a trace back of the actual exception to help diagonize the issue. Any reviews/suggestions?

P.S: I took the liberty to open this PR since it's been a while since that issue was reported, but if the volunteer who volunteered to fix this already had done some work on it that can be merged soon, I will gladly close this one.

### DIFF
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -60,7 +60,7 @@ def load_nox_module(global_config):
         ).load_module()
 
     except (IOError, OSError):
-        logger.error("Noxfile {} not found.".format(global_config.noxfile))
+        logger.exception("Failed to load Noxfile {}".format(global_config.noxfile))
         return 2
 
 


theacodes:
Thank you for doing this!

PR #277, opened 2020-01-29
Prefer importlib.metadata for metadata loading.

### LINKED ISSUE #271: Include `setuptools` as an explicit project requirement due to `import pkg_resources`
**Describe the bug**
`nox` depends on `setuptools` but does not specify it explicitly in `setup.py`. This may not be a common issue, but in a minimal CI environment, I ran into this issue.

**How to reproduce**
If `setuptools` is not present, `nox` cannot import `pkg_resources`.

See the abbreviated console session below for an example:

```
> nox --help
  File "...\python38\lib\runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "...\python38\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "...\AppData\Roaming\Python\Python38\Scripts\nox.exe\__main__.py", line 5, in <module>
  File "...\AppData\Roaming\Python\Python38\site-packages\nox\__main__.py", line 24, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
> pip install setuptools
Collecting setuptools
  ...
Successfully installed setuptools-44.0.0
> nox --help
usage: nox [-h] [--version] [-l] [-s [SESSIONS [SESSIONS ...]]] [-k KEYWORDS] [-v] [-r] [--no-reuse-existing-virtualenvs] [-f NOXFILE] [--envdir ENVDIR]
           [-x] [--no-stop-on-first-error] [--error-on-missing-interpreters] [--no-error-on-missing-interpreters] [--error-on-external-run]
           [--no-error-on-external-run] [--install-only] [--report REPORT] [--non-interactive] [--nocolor] [--forcecolor]
           ...

Nox is a Python automation toolkit.
```

**Expected behavior**
The `nox` package should pull `setuptools` as an explicit dependency in `setup.py`.

jaraco:
Fixes #271.

### DIFF
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -22,9 +22,9 @@ control to :meth:``nox.workflow.execute``.
 import sys
 
 try:
-    import importlib.metadata as metadata
+    import importlib.metadata as metadata  # type: ignore
 except ImportError:
-    import importlib_metadata as metadata
+    import importlib_metadata as metadata  # type: ignore
 
 from nox import _options, tasks, workflow
 from nox.logger import setup_logging
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -23,7 +23,7 @@ import sys
 
 try:
     import importlib.metadata as metadata  # type: ignore
-except ImportError:
+except ImportError:  # pragma: no cover
     import importlib_metadata as metadata  # type: ignore
 
 from nox import _options, tasks, workflow
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -21,7 +21,10 @@ control to :meth:``nox.workflow.execute``.
 
 import sys
 
-import pkg_resources
+try:
+    import importlib.metadata as metadata
+except ImportError:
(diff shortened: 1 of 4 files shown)


nox/__main__.py
```diff
@@ -22,9 +22,9 @@
 import sys
 
 try:
-    import importlib.metadata as metadata
-except ImportError:
-    import importlib_metadata as metadata
+    import importlib.metadata as metadata  # type: ignore
+except ImportError:  # pragma: no cover
```
theacodes: Out of curiosity, when would this branch get hit?

theacodes:
Thanks for doing this @jaraco, I had one question before merging.

nox/__main__.py
jaraco: On Python 3.7 or earlier.

theacodes:
Thanks so much for doing this

PR #280, opened 2020-02-01
Added venv params

mostealth:
Session parameter venv_params has been added as an optional list for virtualenv, venv and conda backends


mostealth:
Do you want it at the end of that section in that same section or in a new section?

theacodes:
Whichever you think works the best.

Issue #276, opened 2020-01-29
Are more type hints desired?


Peilonrayz:
I was interested in adding a different virtual environment / package manager to nox, independently. And noticed you don't have many type hints. The [`noxfile.py` runs mypy](https://github.com/theacodes/nox/blob/8151acdf0a8206559a0c42e50b0b199c5592da8e/noxfile.py#L70), so I'm thinking the answer is yes.

Are more type hints / static typing desired?

---

If you do:

-   Is there anything you don't want to be typed?
-   If required, are you ok with adding `typing_extensions` as an additional requirement?  
    Should it be avoided at all costs?
-   If a situation arises where mypy errors unless a substantial refactor occurs, what is the desired solution?

    If it's not deemed possible to resolve the situation with type hints or variable annotations alone. Should the error be silenced, `# type: ignore`, and a todo, `# TODO: ...`, be added?

cs01:
I am not a maintainer but I added the mypy  checks and type hints in https://github.com/theacodes/nox/pull/230, and expanded coverage in a few subsequent PRs. 

Yes, they would like more type coverage. I don’t think there is anything that shouldn’t be typed. I know there is push for not adding new end-user dependencies to this project, so if you need that package, it would be preferable to only install it in the nox lint session where mypy is run. 

I have also run into cases where a substantial refactor is needed. In my experience, the maintainers prefer you make a PR with just the minimal refactor (to keep the diff reviewable), then add types in another PR. 

I have been adding “type: ignore” where tricky things have arisen. I think that’s fine going forward. If we ever want to go back and fix them we can do a search for “type: ignore”. Right now nothing is typed so typing 99% of the code with a few ignores is much better than keeping none of it typed.

Peilonrayz:
Ok, that all sounds reasonable. Adding `typing_extensions` to just the lint session should be possible, if not a handy `# type:ignore` should suffice.

I'll start work on this soon, but I don't think I'll have any PRs for at least a week.

theacodes:
I couldn't have said it any better than @cs01. I'd love more type hints. :)

Peilonrayz:
With the merger of #282 100% of the code has been typed. I'm going to close this as further issues should probably be their own issues.

PR #282, opened 2020-02-04
Upgrade typehints to support disallow-untyped-defs flag

Peilonrayz:
# Core Changes

Since more typehints are desired, #276, I have added typehints to support the `--disallow-untyped-defs` flag. The change set is large, as I've added annotations so all functions have them. For the most part the change set only includes:

-   Add the required `typing` imports.
-   Add [function annotations](https://www.python.org/dev/peps/pep-3107/) to all functions.
-   Add and update some [type comments](https://www.python.org/dev/peps/pep-0484/#type-comments).
-   Ignore typehints, "type: ignore", which can't be solved by the above solutions.

Even though all functions have been hinted, this is by no means a complete solution. There are lots of errors being ignored, with a significant amount being around how you bind things like `python` to a `Callable`. There's also a lack of generics, and there's a lot of `Any`.

# Errors

-   When running `nox --session tests-3.6`, with 3.6.0, one of the `test_command.py` tests fails. With the final line of the original traceback being:

    ```text
    sqlite3.OperationalError: Safety level may not be changed inside a transaction
    ```
 
    This only happens when using nox. Manually changing into the venv and running pytest doesn't cause that error.

-   For me the coverage test fails. I'm unsure if this is due to not using Conda, or because I'm using x.y.0 for all Python versions.
-   The `docs` session doesn't work for me on Windows. The command `rm` can't be found.
-   Black and isort can't agree on how the imports should be formatted in `_parametrize.py`, `manifest.py` and `sessions.py`. I have manually ran Black to get the `lint` session to not error locally.

These errors are making it hard to see if anything basic is blocking the PR.

# Additional Changes

I've tried to keep the changes strictly to the core changes. However, I have deviated from that plan slightly. And so below are all the changes that I have made that are to achieve the goal, but aren't addressed above.

## Non-typehint

-   `.gitignore`

    -   Mypy creates a cache under `.mypy_cache/`.
    -   Visual Studio Code puts configurations under `.vscode/`.

-   `noxfile.py`  
    I have added the `--disallow-untyped-defs` flag when mypy is run.

### Imports

-   `nox/manifest.py`  
    Import `argparse` and `nox.sessions.Session`.
    
-   `nox/sessions.py`  
    Import `argparse` and `nox.manifest.Manifest`.

-   `nox/tasks.py`  
    Import `importlib.machiniery` and `types`.

-   `nox/workflow.py`  
    Import `argparse`.

### [Type Aliases](https://www.python.org/dev/peps/pep-0484/#type-aliases)

These are to reduce duplicate definitons, and to make the types easier to read.

-   `nox/_parameterize.py`  
    Add `ArgValue`.

-   `nox/registry.py`  
    Add `Python`.

### Other

-   `nox/logger.py`  
    Use `typing.cast` to ensure `logger` gets the correct type.

-   `nox/virtualenv.py`  
    Add an explicit `return None` to `locate_via_py`.

    The added return can be replaced with `--no-warn-no-return` flag to mypy instead.

-   `nox/_typing.py`  
    I have added this file as a small central interface to `typing` and `typing_extensions`. This is as `typing_extensions` supplies the backport and experimental typing features. Currently it, `_typing.py`, only contains code to interface with Python 3.5.0-3 and Python 3.6.0-1.

    I have used `__all__` to ignore flake8 as `NoReturn` is being used, but is being used in a string annotation - as it will not be defined in Python 3.5.0 if the user doesn't have `typing_extensions` installed.

---

This is a fairly big PR. I'm happy to split it up if desired.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,9 @@
 [run]
 branch = True
+omit =
+    nox/_typing.py
 
 [report]
 exclude_lines =
     pragma: no cover
+    if _typing.TYPE_CHECKING:
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+.mypy_cache/
+.vscode/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,6 @@
+[settings]
+multi_line_output=3
+include_trailing_comma=True
+force_grid_wrap=0
+use_parentheses=True
+line_length=88
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -21,13 +21,13 @@ control to :meth:``nox.workflow.execute``.
 
 import sys
 
+from nox import _options, tasks, workflow
+from nox.logger import setup_logging
+
 try:
     import importlib.metadata as metadata  # type: ignore
 except ImportError:  # pragma: no cover
     import importlib_metadata as metadata  # type: ignore
-
-from nox import _options, tasks, workflow
-from nox.logger import setup_logging
 
 
 def main() -> None:
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -27,7 +27,7 @@ from nox import _options, tasks, workflow
 from nox.logger import setup_logging
 
 
-def main():
+def main() -> None:
     args = _options.options.parse_args()
 
     if args.help:
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -21,7 +21,7 @@ import argparse
 import collections
 import functools
 from argparse import ArgumentError, ArgumentParser, Namespace, _ArgumentGroup
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import argcomplete  # type: ignore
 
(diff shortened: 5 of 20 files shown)


theacodes:
Thank you so much for doing this!

> When running nox --session tests-3.6, with 3.6.0, one of the test_command.py tests fails. With the final line of the original traceback being

Not sure what's going on here. I'll try to research when I'm home next week, but if you solve it in the meantime that's awesome!

> For me the coverage test fails. I'm unsure if this is due to not using Conda, or because I'm using x.y.0 for all Python versions.

Yeah, coverage won't report everything locally if you don't have conda. This is expected. You can push here and see if CI passes.

> The docs session doesn't work for me on Windows. The command rm can't be found.

I wouldn't sweat it. CI will test docs.

> Black and isort can't agree on how the imports should be formatted in _parametrize.py, manifest.py and sessions.py. I have manually ran Black to get the lint session to not error locally.

Maybe update the isort to this config:

```
[settings]
multi_line_output=3
include_trailing_comma=True
force_grid_wrap=0
use_parentheses=True
line_length=88
```

That seems to be what the black documentation recommends.

Peilonrayz:
Ok, I'll try get everything green here. I was hoping Travis would run, but it seems I needed to resolve the conflicts first. So I'll start trying to fix any issues it reports.

I don't think I'll be able to solve the issue with 3.6.0, as I barely know how I'd debug it - and my first step made the issue disappear. I can give it a go once I get the PR to light up green. But I don't think I'll be of any use here.

Peilonrayz:
# Further changes
    
-   Merge changes. I made a couple of blunders so it's split over two commits.
-   Update isort so it's not fighting black. Thanks @theacodes your solution worked straight away.
-   Ignore coverage on typing only code. This comes in two parts:

    -   I've added `if _typing.TYPE_CHECKING:` as an exclude line. This is because any code that follows it will never run.
    -   I have flat out ignored `nox/_typing.py`.

        I think this is the best solution, as if anyone is running the tests on 3.5.0 than they are likely to get different coverage results than 3.5.latest. Additionally I don't think coverage is a good metric for the code in the file.

---

CI seems to be happy with the above changes. I'm still getting the error with 3.6.0, however it was erroring before I started work on this PR, so I hope it's thought of a separate issue.

I won't be able to fix anything blocking this for the next day or so. But I should be able to after that.

theacodes:
Thank you so much. Just a heads up that I'm at PyCascades this weekend and
likely won't be able to take a look until next week.

Peilonrayz:
No problem, there's no rush. Have a good weekend!

theacodes:
Thank you, @Peilonrayz!

Peilonrayz:
Thanks for the merge 😁

I looked into the 3.6.0 issue today and found it to not be a problem with nox. It looks like either [pytest-cov](https://pypi.org/project/pytest-cov/) or [coverage](https://pypi.org/project/coverage/) is the one with the issue. Sorry to mess you about like this 😅

Issue #292, opened 2020-03-06
Handle the last of the type ignore


Peilonrayz:
# 1. Simple changes

With the exception of a couple of `# type: ignore`s that can't really be removed. I would like to remove all that are left. I have had a quick glance at the ones that remain after #290, and to remove them would require:

1.  Use [`--ignore-missing-imports`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-ignore-missing-imports) to remove all ignores around imports.
2.  Change [`_parametrize.parametrize_decorator`](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/_parametrize.py#L123) to delegate the transformation of `arg_values_list` to a function so we don't reuse the same variable name.  
    Alternately we could create a new variable and change the [two `if`s](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/_parametrize.py#L125) to an `if` `elif` `else`.

3.  Change [`sessions.Session.virtualenv`](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/sessions.py#L121) to treat `None` as an `AttributeError`, changing the return type to `ProcessEnv` rather than `Optional[ProcessEnv]`.

    Additionally change all instances of `self._runner.venv` to `self.virtualenv`. This would require some changes to the tests.

4.  Change [`sessions.Session.error`](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/sessions.py#L346) and [`skip`](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/sessions.py#L350) to not allow keyword arguments. As they just produce the following `TypeError`:

    > TypeError: Exception() takes no keyword arguments

5.  Add a `create` method to [`virtualenv.ProcessEnv`](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/virtualenv.py#L40) that just raises `NotImplementedException`. This is because we're using `PrecessEnv` like an interface, as the two subtypes `CondaEnv` and `VirtualEnv` both define this method.
[...]

PR #294, opened 2020-03-06
Remove some type ignores

Peilonrayz:
Hey, I have started work on #292. In this PR I have updated the code to be able to remove the `# type: ignore` from 1.1, 1.2, 1.4, 1.5 and I have fixed an incorrect typing for [`ProcessEnv.allowed_globals`](https://github.com/theacodes/nox/blob/8869bf2a20d51343e76459ca50ae2dc36d2b46e2/nox/virtualenv.py#L160).

Since 1.1 touches a lot of files and since 1.2 was a bit more complex than I originally thought, I've stopped here. The other changes should be fairly trivial.

This leaves 1.3, 1.6 and all of 2 to be done.

### DIFF
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -25,9 +25,9 @@ from nox import _options, tasks, workflow
 from nox.logger import setup_logging
 
 try:
-    import importlib.metadata as metadata  # type: ignore
+    import importlib.metadata as metadata
 except ImportError:  # pragma: no cover
-    import importlib_metadata as metadata  # type: ignore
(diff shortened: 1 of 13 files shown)


Peilonrayz:
I think it's time I learnt how to install all the Conda versions on Windows.

Peilonrayz:
I just noticed this doesn't work on 3.5.0. I will fix this.

Peilonrayz:
Fixed the issue with 3.5.0

Issue #298, opened 2020-03-15
Update noxfile.py to have better support for Windows


Peilonrayz:
When developing I run `nox -r` before finishing a commit so I know that I can push without problems. Whilst I find and fix most errors, docs and cover never complete successfully. This is causing fractured PRs, as I don't know coverage has fallen - as it always errors - until it's too late.

# Coverage

Coverage fails to run on the following lines.

-   nox/virtualenv.py

    -   [189](https://github.com/theacodes/nox/blob/ca6f2c288e7a858b623a6213a39b0d78c702a64c/nox/virtualenv.py#L189) (186->189)  
        Changing this [to a ternary](https://stackoverflow.com/q/47958166) allows for 100% coverage.
    -   [212](https://github.com/theacodes/nox/blob/ca6f2c288e7a858b623a6213a39b0d78c702a64c/nox/virtualenv.py#L212) (211->212)
    -   [352](https://github.com/theacodes/nox/blob/ca6f2c288e7a858b623a6213a39b0d78c702a64c/nox/virtualenv.py#L352) (351->352)

-   nox/workflow.py

    -   [57](https://github.com/theacodes/nox/blob/ca6f2c288e7a858b623a6213a39b0d78c702a64c/nox/workflow.py#L57) (44->57)

# Docs

1.  Change `rm -rf` to [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree).
2.  Change the Python version to 3.7 because [sphinx-autobuild](https://github.com/GaretJax/sphinx-autobuild/issues/77), which relies on [tornado](https://github.com/tornadoweb/tornado/issues/2608), no longer works on Windows on 3.8.
3.  Sphinx fails to follow the reference to changelog and contributing.  
    Since the files don't have a heading this causes some warnings. Since the `-W` flag is passed this elevates them to errors.

    I have found two solutions:

    -   Converting everything to rst, and [use `include`](https://stackoverflow.com/a/17217041).
    -   Don't use the `-W` flag on Windows.

---

Most of the fixes are rather straight forward. However Docs (3) seems controversial, how would you like me to approach it?

PR #300, opened 2020-03-19
Improve automated tests support on Windows

Peilonrayz:
This fixes the majority of #298. Most of the changes should be self-explanatory.  
However, I have changed coverage to combine the output as [a 3.8 specific coverage bug](https://github.com/nedbat/coveragepy/issues/959) was causing `cover` to fail.

I have not fixed Docs.3 as I'm unsure how I should fix it, this means `docs` is still failing.


tests/test_virtualenv.py
```diff
@@ -296,12 +301,16 @@ def test_create_venv_backend(make_one):
     venv.create()
 
 
-@pytest.mark.skipif(IS_WINDOWS, reason="Not testing multiple interpreters on Windows.")
 def test_create_interpreter(make_one):
-    venv, dir_ = make_one(interpreter="python3")
+    interpreter = "3.7" if IS_WINDOWS else "python3"
```
theacodes: Can we make this work regardless of which Python version we're testing with?

Peilonrayz:
I'm confused. As far as I know neither the Windows or non-Windows are
testing all Python versions.

I changed python3 to 3.7 as the former can't resolve to an interpreter. Not
to version pin the tests.

How would you like to proceed?

theacodes:
I would prefer a single session (like tests-3.6) be able to pass if *only* that interpreter is installed. Does that make sense?

Peilonrayz:
I don't understand what you mean. Are you saying if a user only has Python
3.6 installed, you want all the tests to pass?

When making the change I looked at the rest of the file and see lots of 3.x
specifiers. I don't understand why this one is important and the others are
not. Is this also a problem with `test_condaenv_create_interpreter`?

theacodes:
Yes, the test should either pass or skip. I can't see why this test can't be changed to work with whatever python version the tests are currently running under.

Peilonrayz:
OK, is not passing an interpreter, on just Windows, an acceptable
solution?
That way the interpreter is guaranteed to exist.

theacodes:
sure

theacodes:
Howdy, @Peilonrayz. It looks like this PR has gone stale. If you're still interested in getting it merged, let me know and we can get this started again.

Peilonrayz:
Sorry life got busy. I've looked into the problem again and `make_one(interpreter="python3")` fails to resolve to an interpreter for me. Maybe I have a bug with my setup or nox has a bug making the test fail. I have gotten `make_one(interpreter=None)` to work on Windows but I'd just be adding a potential bug on Linux since the `assert dir_.join("bin", "python3").check()` could fail if Python 2 is used. I could use `make_one(interpreter=None if IS_WINDOWS else "python3")`, which smells to me.

Bottom line; I don't know how to nicely solve the issue we have here.

Would the best course of action be to just remove the changes to `test_create_interpreter` and let the rest of the PR through?

theacodes:
> Would the best course of action be to just remove the changes to
test_create_interpreter and let the rest of the PR through?

i'm fine with that. :)

Peilonrayz:
Ok! I've undone the changes to `test_create_interpreter`, but the commit isn't showing up here for me. This is my first time working on a closed PR; do I need to create a new PR?

theacodes:
I re-opened it. Looks like you may need to `git push` though?

Peilonrayz:
Thank you for reopening my PR. FWIW I edited the file on GitHub and when I got to the issue today the commit had appeared here.

I had to merge some changes, for the most part I took your changes. Where I kept `coverage combine` to fix a bug in Python that affects coverage on Windows. Since I can't reproduce the issue - you may have changed the one bit of code which had the bug. I'm happy to remove the call to coverage combine if you'd like.

theacodes:
Thank you!!

PR #310, opened 2020-04-20
Fixed the default paths for conda on windows where the `python.exe` found was not the correct one

### LINKED ISSUE #256: Session.run finds incorrect Python
**Describe the bug**
`session.run('python', ..., ....)` can not find the *virtual* environments python. 

**How to reproduce**
Here is my noxfile.py. The last line is the where the issue occurs.  There are **no** issues in creating a conda environment, installing conda packages, installing pip packages, and performing the pytest. I have verified it all and it completely works as expected (installing and using binaries and scripts in the virtual environment). The issue is only the last command which could only find a default python on the system (external). I have diagnosed the issue and it is explained below.

```python
import nox
import os
VENV = 'conda' if os.name == 'nt' else None
NAME = 'win' if os.name == 'nt' else 'linux'

@nox.session(
    python=['3.6', '3.7'],
    reuse_venv=True,
    venv_backend=VENV,
    name='test_' + NAME)
def test(session):
    if VENV == 'conda':
        session.conda_install('--channel=conda-forge', 'shapely')
    session.install('-e', '.[dev]')
    session.run('pytest')
    session.run('python', 'setup.py', 'sdist', 'bdist_wheel')
```

**Fix**

The issue is here:  https://github.com/theacodes/nox/blob/ac0992332a266f85152db8a5252b1fdf9818eb75/nox/command.py#L82

where `cmd` is `python` and  `path` will be to the virtual environments `Script` directory. Here is a picture of what `path` is:
scripts

As you can see there is no python here. The correct python is here:

python

The change I performed to fix this issue is highlighted in the images. I changed this:
https://github.com/theacodes/nox/blob/ac0992332a266f85152db8a5252b1fdf9818eb75/nox/command.py#L36

to this:

```
if path:
        full_path = py.path.local.sysfind(program, paths=[path, os.path.dirname(path)])
```

Now everything works as expected. I am not sure if this is the best way to fix but it works well for me. If this is acceptable I can do a pull request if you would like.

Thanks for your great work!

smarie:
Changelog (should I commit it in the changelog file ?)

 - Virtual environments now have a `bin_paths` attribute instead of a single `bin`. `command.run` and `which` both have their `path` argument renamed `paths` and supporting a list of paths, to handle it. 

 - Fixed the default paths for conda on windows where the `python.exe` found was not the correct one. Fixes #256

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -131,9 +131,9 @@ class Session:
         return self._runner.func.python
 
     @property
-    def bin(self) -> Optional[str]:
-        """The bin directory for the virtualenv."""
-        return self.virtualenv.bin
+    def bin_paths(self) -> Optional[List[str]]:
+        """The bin directories for the virtualenv."""
+        return self.virtualenv.bin_paths
 
     @property
     def interactive(self) -> bool:
@@ -243,7 +243,7 @@ class Session:
             kwargs["external"] = True
 
         # Run a shell command.
-        return nox.command.run(args, env=env, path=self.bin, **kwargs)
+        return nox.command.run(args, env=env, paths=self.bin_paths, **kwargs)
 
     def conda_install(self, *args: str, **kwargs: Any) -> None:
         """Install invokes `conda install`_ to install packages inside of the
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -139,10 +139,8 @@ class Session:
     @property
     def bin(self) -> Optional[str]:
         """The first bin directory for the virtualenv."""
-        try:
-            return self.bin_paths[0]
-        except (TypeError, KeyError):
-            return None
+        paths = self.bin_paths
+        return paths[0] if paths is not None else None
 
     def create_tmp(self) -> str:
         """Create, and return, a temporary directory."""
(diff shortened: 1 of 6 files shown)


nox/sessions.py
```diff
@@ -131,9 +131,9 @@ def python(self) -> Optional[Union[str, Sequence[str], bool]]:
         return self._runner.func.python
 
     @property
-    def bin(self) -> Optional[str]:
-        """The bin directory for the virtualenv."""
-        return self.virtualenv.bin
+    def bin_paths(self) -> Optional[List[str]]:
```
theacodes, line 134: Nox has avoided breaking API changes for a long time. Can you add a `bin` alias that returns the first entry in the list to maintain backwards compatibility?

smarie: Sure. Shall I do it for ProcessEnv , CondaEnv and VirtualEnv too ?

theacodes: Yes.

smarie: Done.

smarie:
thanks!

PR #312, opened 2020-04-20
`conda_install` and `install` args are now automatically double-quoted when needed.

### LINKED ISSUE #311: `conda_install` should automatically escape args when they contain version constraints
In `Session.conda_install` one reads this example:

```python
session.conda_install('--channel=conda-forge', 'dask==2.1.0')
```

Now consider that we tweak it very lightly: 

```python
session.conda_install('--channel=conda-forge', 'dask>2.0.0')
```

Running it, at least on windows, will 

 - not take into account the constraint ">2.0.0"
 - write a nice but completely useless file named "2.0.0" in the working dir, containing the installation log.

Even worse: reverting the sign (`'dask<2.0.0'`) yields an error:

```shell
Command conda install --yes --prefix C:\......\.nox\tests-2-7 --channel=conda-forge dask<2.0.0 failed with exit code 1:
The system cannot find the file specified.
  ...
  File "C:\_dev\python_ws\_Libs_OpenSource\nox\nox\sessions.py", line 293, in conda_install
    **kwargs
  File "C:\_dev\python_ws\_Libs_OpenSource\nox\nox\sessions.py", line 246, in _run
    return nox.command.run(args, env=env, paths=self.bin_paths, **kwargs)
  File "C:\_dev\python_ws\_Libs_OpenSource\nox\nox\command.py", line 125, in run
    raise CommandFailed("Returned code {}".format(return_code))
nox.command.CommandFailed: Returned code 1
```

This is of course because the `>` and `<` operators are handled by the shell and should be escaped. I'll try to propose some fix.

smarie:
Changelog (let me know if I should commit it somewhere):

 - `conda_install` ~~and `install`~~ args are now automatically double-quoted when they contain a `<` or a `>`. Fixes #311

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -76,7 +76,9 @@ def _dblquote_pkg_install_args(args):
     def _dblquote_pkg_install_arg(pkg_req_str):
         # sanity check: we need an even number of double-quotes
         if pkg_req_str.count('"') % 2 != 0:
-            raise ValueError("ill-formated argument with odd number of quotes: %s" % pkg_req_str)
+            raise ValueError(
+                "ill-formated argument with odd number of quotes: %s" % pkg_req_str
+            )
 
         if "<" in pkg_req_str or ">" in pkg_req_str:
             if pkg_req_str[0] == '"' and pkg_req_str[-1] == '"':
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -359,8 +359,8 @@ class Session:
         if not args:
             raise ValueError("At least one argument required to install().")
 
-        # Escape args that should be
-        args = _dblquote_pkg_install_args(args)
+        # Escape args that should be: not needed and would make pip raise an InvalidRequirement
+        # args = _dblquote_pkg_install_args(args)
 
         if "silent" not in kwargs:
             kwargs["silent"] = True
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -74,14 +74,16 @@ def _dblquote_pkg_install_args(args):
 
     # routine used to handle a single arg
     def _dblquote_pkg_install_arg(pkg_req_str):
-        if '<' in pkg_req_str or '>' in pkg_req_str:
+        if "<" in pkg_req_str or ">" in pkg_req_str:
             if pkg_req_str[0] == '"':
                 # already double-quoted string
                 return pkg_req_str
             else:
                 # need to double-quote string
                 if '"' in pkg_req_str:
-                    raise ValueError("Cannot escape requirement string: %s" % pkg_req_str)
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


smarie:
Note: although the tests pass, it seems that this fix is only needed for `conda` and makes `pip` crash with

```bash
Invalid requirement: '"setuptools"'
  File "C:\...\.nox\tests-2-7\lib\site-packages\pip\req\req_install.py", line 82, in __init__
    req = Requirement(req)
  File "C:\...\.nox\tests-2-7\lib\site-packages\pip\_vendor\packaging\requirements.py", line 96, in __init__
    requirement_string[e.loc:e.loc + 8]))
InvalidRequirement: Invalid requirement, parse error at "'"setupto'"
```

theacodes:
This looks fine, but can you show me some invocations that cause this behavior?

smarie:
For example this:

`session.conda_install("pytest>7")`

On windows, it runs with success ! Indeed the `>7` is understood by `cmd` as "pipe the output to a new file named 7".

smarie:
Note that there is a workaround today: the user has to quote the dependency, e.g. `session.conda_install('"pytest>7"')`

nox/sessions.py
```diff
@@ -361,6 +395,9 @@ def install(self, *args: str, **kwargs: Any) -> None:
         if not args:
             raise ValueError("At least one argument required to install().")
 
+        # Escape args that should be: not needed and would make pip raise an InvalidRequirement
```
theacodes, line 398: You can remove this instead of leaving it commented out.

smarie: Done. This was for maintainability - but instead of writing this here, I edited the comment in the conda part so that future readers are not tempted to apply the same escaping procedure to the pip section.

tswast:
I'm a bit confused as to why this is necessary. When I run `conda install "urllib3<1.25"`, the extra quotes around the package shouldn't actually be sent to conda itself. Adding extra quotes doesn't seem to hurt (`conda install "'urllib3<1.25'"` does work), but why are they necessary?

tswast:
It appears that this is a windows-only issue (https://docs.python.org/3/library/subprocess.html#subprocess.Popen)

> On Windows, if args is a sequence, it will be converted to a string in a manner described in Converting an argument sequence to a string on Windows. This is because the underlying CreateProcess() operates on strings.

tswast:
LGTM, since conda is able to handle the extra quotes on all platforms.

smarie:
Great, thanks for the review @tswast ! I updated the branch, it should be ready for merging

smarie:
(Appveyor build did succeed, but apparently the status never returned correctly)

theacodes:
Thanks all!

PR #316, opened 2020-04-21
`venv_backend` new options and choices

### LINKED ISSUE #301: Add a command line option for setting python=False
If you have a `noxfile.py` that you use for development but also use `Docker` as a part of your CI you can skip having to create a virtualenv and reinstall dependencies that may have been installed as a part of your Docker image build step. For instance if I have:

```dockerfile
FROM python:3

COPY dev-requirements.txt .
RUN python -m pip install -r dev-requirements.txt

COPY . .
RUN nox -s test --use-current-env
```

then my `session.install("-r", "dev-requirements.txt")` step within my `noxfile.py` would only require checking cache, much faster to run. 

Using `--use-current-env` (or alternative name, I'm fine with anything!) would override any `python=[]` within `noxfile.py` to `python=False` and use interpreter at `sys.executable`.

### LINKED ISSUE #315: New global option `nox.options.venv_backend` to set the default backend
As you've understood from my recent activity on this repo, I am a conda user ;)

I like my `noxfile.py` but it could be much lighter if I did not have to write `venv_backend="conda"` everywhere.

Therefore I propose to define a new global option `nox.options.venv_backend` (or whatever alternate name you'll prefer of course)

### LINKED ISSUE #317: Be able to force the backend temporarily
We can imagine that a user working on some opensource project and cloning it will maybe have different settings than the project's continuous integration engine.

It could therefore be quite interesting to allow users to override the venv backend through a commandline or a global option.

This would be even interesting if combined to a "no_venv" choice of backend, in order to directly work on a passthrough (see #301 )

### LINKED ISSUE #318: Allow `session.install` and `session.conda_install` to run when no venv backend is set
As of now, when no venv backend is set (either through `python=False` or by setting the venv to `'none'`), users get a `ValueError` when trying to `install` or `conda_install`.

In some cases (See also #301 ) this would be actually useful.

smarie:
Changelog (let me know if I should commit it in the changelog file):

 - New global option `nox.options.default_venv_backend` to set the default backend. Fixes #315 
 - New global option `nox.options.force_venv_backend` to temporarily force the backend. The commandline version of this option will probably be used more frequently than the noxfile one, but thanks to the great generic option mechanism built in nox, I got both for the same price :) . Fixes #317
 - Added possibility to declare a "no backend / passthrough" venv with `venv_backend='none'` (or to force it temporarily with the `--no-venv` cli option, equivalent to `--force-venv-backend none`, see above). Fixes #301
 - Now `install` and `conda_install` work when there is no venv backend (or python=False). Previously a `ValueError` was raised. Fixes #318

I updated the doc accordingly.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -132,7 +132,7 @@ Will produce these sessions:
 
 Note that this expansion happens *before* parameterization occurs, so you can still parametrize sessions with multiple interpreters.
 
-If you want to disable virtualenv creation altogether, you can set ``python`` to ``False``, or set `venv_backend` to ``"none"``, both are equivalent.
+If you want to disable virtualenv creation altogether, you can set ``python`` to ``False``, or set ``venv_backend`` to ``"none"``, both are equivalent. Note that this can be done temporarily through the :ref:`--no-venv <opt-force-venv-backend>` commandline flag, too.
 
 .. code-block:: python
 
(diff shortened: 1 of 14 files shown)


smarie:
Finally all set and ready for review :) @theacodes @sethmlarson

smarie:
Thanks @theacodes for the review, and thanks for merging !

Issue #319, opened 2020-05-01
create (and pass to sessions) a "temporary directory" they can use for output files


moshez:
**How would this feature be useful?**

When testing documentation builds, it is nice to have the output be under the `nox` temporary directory instead of in an OS-specific temporary directory. It makes cleanup easier, and it means there is a consistent URL to be loaded in the browser.

**Describe the solution you'd like**

A session should have access to a variable "environment temporary directory", and pass it on to commands as appropriate (for example, as the `-d` option to sphinx).

**Describe alternatives you've considered**

In theory, session could grab the virtual environment directory and create a `tmp` subdirectory there themselves, but this is a little ugly.

PR #320, opened 2020-05-03
Address #319: Give a create_tmp API

moshez:
This doesn't have docs or tests! Hooray :)

It does however show one way to implement it (kinda hacky) and a working use case for further discussion.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -135,6 +135,16 @@ class Session:
         """The bin directory for the virtualenv."""
         return self.virtualenv.bin
 
+    def create_tmp(self) -> Optional[str]:
+        bin = self.virtualenv.bin
+        if bin is None:
+            raise ValueError("Cannot have a tmp if no bin")
+        updir = os.path.dirname(bin)
+        tmpdir = os.path.join(updir, 'tmp')
+        if not os.path.exists(tmpdir):
+            os.mkdir(tmpdir)
+        return tmpdir
+
     @property
     def interactive(self) -> bool:
         """Returns True if Nox is being run in an interactive session or False otherwise."""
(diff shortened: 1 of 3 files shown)


moshez:
Hi @theacodes ! This is a rough draft, but I want some feedback on how much in the wrong direction am I going. Is it reasonable to assume the bin directory is writable? Should we push this API to the specific virtualenv? Is it OK to fail if bin is None? Etc.

theacodes:
This looks like a great start.

> Is it reasonable to assume the bin directory is writable? 

Yup.

> Should we push this API to the specific virtualenv?

Nope, I think even "virtualenvless" sessions will have use cases for the tmpdir.

> Is it OK to fail if bin is None?

I think this should work even if there's no virtualenv for the session.

moshez:
Hi @theacodes ! I added tests and documentation. I'm not sure if I understood your response to the third question -- do you mean I should have a tmpdir even without bin? In that case, I'm not sure the location would still make sense.

theacodes:
>  do you mean I should have a tmpdir even without bin? In that case, I'm not sure the location would still make sense.

Yup.

I think it makes sense for it to go in `.nox/{session name}/tmp`.

moshez:
OK, let me know what you think now. It should always work now.

theacodes:
Looks good. 👌

dhermes:
@moshez I'd love to chat about this feature and how it relates to my use / abuse of `.nox/libbezier-debug` and `.nox/libbezier-release` to cache and re-use debug and release builds of `libbezier` to speed up local dev for the Python parts of `bezier`: https://github.com/dhermes/bezier

moshez:
@dhermes sent you an e-mail re: chatting about it. let me know!

PR #323, opened 2020-05-09
draft: --add-timestamp

omry:
introducing an --add_timestamp flag and support for in in the logging system.

#289

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -271,7 +271,8 @@ options.add_options(
         group=options.groups["secondary"],
         action="store_true",
         help="Adds a timestamp to logged output.",
-        noxfile=True),
+        noxfile=True,
+    ),
     _option_set.Option(
         "default_venv_backend",
         "-db",
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -265,6 +265,14 @@ options.add_options(
         noxfile=True,
     ),
     _option_set.Option(
+        "add_timestamp",
+        "-ts",
+        "--add-timestamp",
+        group=options.groups["secondary"],
+        action="store_true",
+        help="Adds a timestamp to logged output.",
+        noxfile=True),
+    _option_set.Option(
         "default_venv_backend",
         "-db",
         "--default-venv-backend",
--- a/nox/__main__.py
+++ b/nox/__main__.py
@@ -41,7 +41,9 @@ def main() -> None:
         print(metadata.version("nox"), file=sys.stderr)
         return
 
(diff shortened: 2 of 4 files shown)


omry:
1. Looking for overall feedback.
2. colorlog test is failing even though things are working. looks like some kind of interaction between colorlog and logcap.

omry:
Screen shots:

image
image

omry:
@theacodes ping

nox/_options.py
```diff
@@ -221,6 +221,15 @@ def _session_completer(
         help="Logs the output of all commands run including commands marked silent.",
         noxfile=True,
     ),
+    _option_set.Option(
+        "add_timestamp",
+        "-ts",
+        "--add_timestamp",
```
theacodes, line 227: should be `--add-timestamp`.

theacodes:
> Looking for overall feedback (Is it okay to add the flag? is the log format okay? [I noticed it's missing nox> )

This is a great idea. I would definitely add the `nox >`, as it's important to be able to distinguish Nox's output from the output that programs called via `run` produce.

> colorlog test is failing even though things are working. looks like some kind of interaction between colorlog and logcap.

We'll definitely wanna get that working, but I'm not a fan of using subprocess in tests unless its absolutely necessary.

omry:
> We'll definitely wanna get that working, but I'm not a fan of using subprocess in tests unless its absolutely necessary.

The problem I ran into was that for an unknown reason, asctime was not populated in the caplog record when using colorlog. but we are talking about an interaction between two third party libs (caplog and colorlib), through the somewhat complicated Python logging subsystem.

I already spent about an hour or two digging into it. I will give it another shot before resorting to process testing.

omry:
responded to comments, except the test.
Upon further digging, I suspect custom log formatter are [not fully supported](https://github.com/pytest-dev/pytest/issues/2987) by pytest caplog.

I see two options:
1. Drop the testing timestamp override testing for the colorlog. This is very visible so if it breaks we will hear about it and keep the current working testing for non-colorlog mode.
2. Switch both test modes to process based testing.

What do you think?

theacodes:
Sorry for taking so long to get to this. Let's go with (1).

omry:
### Regular run
image

### Run with timestamps
image

### Runt with timestamp and no color.
image

omry:
I noticed that there is some cleanup to do:
1. blacken is not running and seems to be producing results different than `black .`. it should probably be removed from the noxfile.
2. docs/conf.py needs black formatting.

theacodes:
Thanks!

PR #344, opened 2020-09-03
Export `Session` in `__init__`

### LINKED ISSUE #343: Export Session in __init__
Currently, if you'd like to annotate nox callbacks, for e.g. type checking, code completion or docstring lookup, you have to import `Session` from `nox.sessions`:

```py
import nox
from nox.sessions import Session


@nox.session
def foo(session: Session):
    ...
```

Re-exporting `Session` in `__init__` would 'jibe' with nox's global object approach and would perhaps make typing in nox more accessible.

For the `Session` class to be imported/re-exported in `nox.__init__` so as to not require a separate import.

Importing `Session` from `nox.sessions`, as above.

layday:
Closes #343.

### DIFF
--- a/nox/__init__.py
+++ b/nox/__init__.py
@@ -16,5 +16,6 @@ from nox._options import noxfile_options as options
 from nox._parametrize import Param as param
 from nox._parametrize import parametrize_decorator as parametrize
 from nox.registry import session_decorator as session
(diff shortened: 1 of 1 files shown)


theacodes:
thanks!

PR #360, opened 2020-11-20
Add py.typed to manifest.in

### LINKED ISSUE #336: PEP 561 marker file py.typed missing from releases
While there is a [PEP 561](https://www.python.org/dev/peps/pep-0561/) maker file in `nox/py.typed` it is not included in the published releases. To reproduce, extract the published [nox-2020.5.24.tar.gz](https://files.pythonhosted.org/packages/6b/a6/e3f1bdadf1d1eeed17a5cb431796711ae06f8a78107f5f60040fa0d5ac07/nox-2020.5.24.tar.gz) and see that the file is missing.

Mypy therefore does not know that this package includes type annotations and will display errors to users.

I'm actually not sure why the marker file is not included in the releases. The `setup.py` specifically includes it in the package data:

https://github.com/theacodes/nox/blob/00775b470129787b14c45f2853441f9f1cad006d/setup.py#L49

However, this setting might be ignored and perhaps adding the following line to the `MANIFEST.in` is the correct solution:

```
included src/py.typed
```

I would appreciate it, if the correct setting can be found and this published as a bugfix release.

V1NAY8:
Closes #336 
@theacodes Please review :)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include LICENSE
 recursive-include nox *.jinja2
+include nox/py.typed


theacodes:
Thank you!

PR #364, opened 2020-11-26
Sort imports using isort

cjolowicz:
Run the blacken session on the codebase.

PR #366, opened 2020-11-28
Add isort check to lint session

cjolowicz:
This came up in the context of #364:

> Thanks! I think we can add this check in the lint step (`isort --check`)
> 
> https://github.com/theacodes/nox/blob/495b23ff690049e8f3d25f3d29c06458da2ee274/noxfile.py#L82
> 
> Compatible settings are in https://github.com/psf/black/blob/master/docs/compatible_configs.md#isort
> 
> That can be done in another PR of course, CI failure looks unrelated

@stsewd Nox's `.isort.cfg` contains all of the compatible configuration settings you linked to, with the exception of `ensure_newline_before_comments = True`. I did not bother to add that setting to the configuration file because isort 5 has a black profile, rendering manual configuration unnecessary once we upgrade.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -80,7 +80,7 @@ def blacken(session):
 
 @nox.session(python="3.8")
 def lint(session):
-    session.install("flake8==3.7.8", "black==19.3b0", "mypy==0.720")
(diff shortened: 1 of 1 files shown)


stsewd (approved):
Thanks! The new isort black compatibility looks great https://pycqa.github.io/isort/docs/configuration/compatibility_black/

PR #376, opened 2021-01-01
Allow passing PathLike types to Session.chdir()

jdufresne:
os.chdir() has supported PathLib since Python 3.6.

Discovered while adding types pip's noxfile.py:
https://github.com/pypa/pip/pull/9411


stsewd (approved):
Thanks!

PR #377, opened 2021-01-01
Define the location attribute on base class ProcessEnv

jdufresne:
Fixes errors when running mypy on a project's noxfile.py:

    noxfile.py:42: error: "ProcessEnv" has no attribute "location"

Discovered while adding types pip's noxfile.py:
pypa/pip#9411

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -43,6 +43,8 @@ class InterpreterNotFound(OSError):
 class ProcessEnv:
     """A environment with a 'bin' directory and a set of 'env' vars."""
 
+    location: str
+
     # Does this environment provide any process isolation?
     is_sandboxed = False
 


stsewd (commented):
I'm not that familiar with mypy, but location is a per instance attribute, not sure if there is a way to declare it like that.

jdufresne:
@stsewd The syntanx used here is an instance variable type declaration. It does not not assign the attribute to the class (or even the instance). A true assignment is still required for that. Observe:

```py
class MyClass:
    a: str

obj = MyClass()
obj.a
```

```
$ python blah.py
  File "/home/jon/blah.py", line 5, in <module>
    obj.a
AttributeError: 'MyClass' object has no attribute 'a'
```

The syntax is use for type declaration only so it makes mypy happy. Please let me know if I can provide additional detail.

theacodes (approved):

theacodes:
Looks good to me. Thanks, @jdufresne!

cjolowicz:
FTR this is incorrect: PassthroughEnv does not have a `location` attribute, only VirtualEnv and CondaEnv do. And for these, mypy should be able to work out from the type annotations on `__init__`.

The problem is that typical access is via `session.virtualenv.location`, where `session.virtualenv` is annotated as ProcessEnv. That's also the use case in pip's noxfile, AFAICS from the linked PR.

A simple fix would be the following:

```python
class ProcessEnv:
    location: Optional[str] = None
```

This would result in a runtime change though, adding a class attribute on ProcessEnv. I can't see a simple way to solve this purely on the typing level. :thinking: 

Another possibility, maybe preferable, would be to keep the type annotation from this PR, but make it raise on PassthroughEnv:

```python
class PassthroughEnv(ProcessEnv):
    @property
    def location(self) -> str:
        raise RuntimeError("PassthroughEnv has no location")
```

This would also have the advantage that client code does not need to check against None to pass mypy.

PR #378, opened 2021-01-01
Make ProcessEnv.bin always return a str

jdufresne:
The property now never returns None. This simplifies user code containing
types that already knows a bin directory must exist. It avoids the need
to pepper calling code with:

    assert session.bin is not None

Or

    # type: ignore

For example, in pip:
https://github.com/pypa/pip/blob/062f0e54d99f58e53be36be5a45adad89e2429fb/tools/automation/release/__init__.py#L29

A noxfile that tries to access a bin directory that doesn't exist will
now raise an exception.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -167,10 +167,12 @@ class Session:
         return self.virtualenv.bin_paths
 
     @property
-    def bin(self) -> Optional[str]:
+    def bin(self) -> str:
         """The first bin directory for the virtualenv."""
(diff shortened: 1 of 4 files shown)


crwilcox (approved):
This LGTM. I am not 100% on it, but I am not sure you could ever get `None` from `bin_paths`.

theacodes (approved):

theacodes:
Thank you!

PR #382, opened 2021-02-10
Docs: remove outdated notes on Windows compatibility 

### LINKED ISSUE #379: Current state of Windows support
Because tox has surprised me with some very confusing behavior lately, I was looking into converting my project to use Nox instead. Initial experiments have been very pleasant. However, one paragraph in the manual makes me wonder about the state of Windows support. It mentions _"Nox has provisional support for running on Windows."_, and then links to some tox, Python and virtualenv tickets.

Since that note was added to the manual 5 years ago, I wanted to check whether Windows support is still only _provisional_. Because of the links to non-tox issues, I'm assuming that Windows support in Nox will be roughly comparable to that in tox. Could that be correct?

brechtm:
Closes #379


theacodes (approved):
Looks wonderful. Thank you!

PR #393, opened 2021-03-02
Gracefully shutdown child processes

### LINKED ISSUE #139: CTRL-C always results in an error
If `nox -s foo` encounters a `KeyboardInterrupt` but the currently running script catches it and handles it gracefully, `nox` will still report a failure. For example:

```
$ nox -s app -r
nox > Running session app-3.7
nox > Re-using existing virtualenv at .../.nox/app-3-7.
nox > pip install --upgrade --requirement .../requirements/app.txt
nox > cd .../src
nox > python -m flask run
 * Serving Flask app "app.py" (lazy loading)
...
127.0.0.1 - - [20/Sep/2018 15:43:27] "GET /foo HTTP/1.1" 200 -
127.0.0.1 - - [20/Sep/2018 15:43:31] "GET /bar HTTP/1.1" 200 -
127.0.0.1 - - [20/Sep/2018 15:43:33] "GET /baz HTTP/1.1" 404 -
^Cnox > Interrupted...
nox > Session app-3.7 interrupted.
$ echo $?
130
```

However, if the same application runs outside of `nox` there is no error (i.e. `flask` exits gracefully):

```
$ cd src/
$ FLASK_APP=app.py ../.nox/app-3-7/bin/python -m flask run
 * Serving Flask app "app.py"
...
127.0.0.1 - - [20/Sep/2018 15:46:09] "GET /foo HTTP/1.1" 200 -
127.0.0.1 - - [20/Sep/2018 15:46:12] "GET /baz HTTP/1.1" 404 -
127.0.0.1 - - [20/Sep/2018 15:46:13] "GET /bar HTTP/1.1" 200 -
^C
$ echo $?
0
$ cd ..
```

What should the correct `nox` behavior be here?

cjolowicz:
Fixes #139

This PR changes Nox's behavior when receiving a keyboard interrupt (CTRL-C, `SIGINT`) while executing a command with `session.run` or `session.install`.

The current behavior is to terminate the child immediately (`SIGTERM` on Posix, `TerminateProcess` on Windows) and to reraise the `KeyboardInterrupt`, ultimately exiting with failure.

The new behavior is to shut down the child gracefully and only raise if the child did not exit with success. The shutdown sequence consists of these four steps:

1. Check if the child has terminated, with a 300ms timeout. If it did, skip to the last step. Note that the OS delivers the signal to the child as well.
2. Send `SIGTERM` and check again, with a 200ms timeout. If it did, skip to the last step.
3. Send `SIGKILL` and wait without timeout.
4. If the exit status of the child process is non-zero, reraise the `KeyboardInterrupt`. Otherwise, we consider the keyboard interrupt handled and return normally.
   
A few points of note:

- We should probably move `logger.error("Interrupted...")` from `command.run` to `popen.popen`, so the user gets immediate feedback before the shutdown sequence takes course. ~~We'd need to pass `logger` to `popen.popen`, though.~~ What do you think?

- The timeouts correspond to tox's defaults for its `interrupt_timeout` and `terminate_timeout` settings. For commands that need longer to shut down, it may be a good idea to make this configurable in Nox as well. These could be keyword arguments for `session.run`. I don't think this needs to be addressed by this PR though, right? (Interestingly, tox4 no longer seems to have these configurable.)

- tox does the shutdown slightly differently. It first checks if the child has terminated (with a timeout of zero), and sends a duplicate `SIGINT` if it hasn't. I believe the reason is that the shutdown protocol is also used in cases where there was no initial `SIGINT`. The rest of the shutdown sequence is the same. tox always reraises the interrupt, even if the child exited with success. (tox4 essentially behaves like tox3, but busy-waits and skips step 3 on Windows where `Popen.kill` is an alias for `Popen.terminate`.)

- The use case with `pytest --pdb` is not covered here; Nox still insists on terminating the child. When I tried this, I did get to see pytest's traceback before Nox killed it, which is an improvement. One way to support this use case would be a command-line option to disable sending additional signals to the child. I don't think this needs to be addressed by this PR either. tox also does not support this currently (https://github.com/tox-dev/tox/issues/1791).

- When output is captured and the child handles the interrupt, we silently discard anything `Popen.communicate` buffered for us before it was interrupted. However, the only *documented* way to capture output from `session.run` is the `silent` flag, and in that case it doesn't matter.

### DIFF
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -293,7 +293,7 @@ def run_pytest_in_new_console_session(test):
     )
 
     subprocess.run(
-        ["pytest", f"{__file__}::{test}"],
+        [sys.executable, "-m", "pytest", f"{__file__}::{test}"],
         env=env,
         check=True,
         stdout=subprocess.PIPE,
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ctypes
 import logging
 import os
 import platform
@@ -27,6 +28,15 @@ import nox.popen
 import pytest
 
 PYTHON = sys.executable
+
+skip_on_windows_console = pytest.mark.skipif(
+    platform.system() == "Windows" and "DETACHED_FROM_CONSOLE" not in os.environ,
+    reason="Do not run this test attached to a Windows console.",
+)
+
+only_on_windows = pytest.mark.skipif(
+    platform.system() != "Windows", reason="Only run this test on Windows."
+)
 
 
 def test_run_defaults(capsys):
@@ -207,32 +217,23 @@ def marker(tmp_path):
     return tmp_path / "marker"
 
 
+def enable_ctrl_c(enabled):
+    """Enable keyboard interrupts (CTRL-C) on Windows."""
+    kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)
+
+    if not kernel32.SetConsoleCtrlHandler(None, not enabled):
+        raise ctypes.WinError(ctypes.get_last_error())
+
+
 def interrupt_process(proc):
     """Send SIGINT or CTRL_C_EVENT to the process."""
-    if platform.system() == "Windows" and "CI" in os.environ:
-        # Do nothing. The interrupt would be received by every console process,
-        # including the parent process of pytest that is running the CI step.
-        return
-
     if platform.system() == "Windows":
-        # https://stackoverflow.com/a/35792192
-        import threading
+        # Disable Ctrl-C so we don't terminate ourselves.
+        enable_ctrl_c(False)
 
-        handler = signal.getsignal(signal.SIGINT)
-        event = threading.Event()
-
-        def handler_set_event(signum, frame):
-            event.set()
-            return handler(signum, frame)
-
-        signal.signal(signal.SIGINT, handler_set_event)
-
-        try:
-            os.kill(0, signal.CTRL_C_EVENT)
-            while not event.is_set():
-                pass
-        finally:
-            signal.signal(signal.SIGINT, handler)
+        # Send the keyboard interrupt to all processes attached to the current
+        # console session.
+        os.kill(0, signal.CTRL_C_EVENT)
     else:
         proc.send_signal(signal.SIGINT)
 
@@ -240,6 +241,10 @@ def interrupt_process(proc):
 @pytest.fixture
 def command_with_keyboard_interrupt(monkeypatch, marker):
     """Monkeypatch Popen.communicate to raise KeyboardInterrupt."""
+    if platform.system() == "Windows":
+        # Enable Ctrl-C because the child inherits the setting from us.
+        enable_ctrl_c(True)
+
     communicate = subprocess.Popen.communicate
 
     def wrapper(proc, *args, **kwargs):
@@ -278,6 +283,26 @@ def format_program(program, marker):
     return dedent(program).format(MAIN=dedent(main))
 
 
+def run_pytest_without_console_window(test):
+    """Run the given test in a subprocess detached from the console window."""
+    env = dict(os.environ, DETACHED_FROM_CONSOLE="")
+    creationflags = (
+        subprocess.CREATE_NO_WINDOW
+        if sys.version_info[:2] >= (3, 7)
+        else subprocess.CREATE_NEW_CONSOLE
+    )
+
+    subprocess.run(
+        ["pytest", f"{__file__}::{test}"],
+        env=env,
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        creationflags=creationflags,
+    )
+
+
+@skip_on_windows_console
 @pytest.mark.parametrize(
     "program",
     [
@@ -307,10 +332,13 @@ def test_interrupt_raises(command_with_keyboard_interrupt, program, marker):
         nox.command.run([PYTHON, "-c", format_program(program, marker)])
 
 
-@pytest.mark.skipif(
-    platform.system() == "Windows" and "CI" in os.environ,
-    reason="On Windows CI, the keyboard interrupt would terminate our parent process.",
-)
+@only_on_windows
+def test_interrupt_raises_on_windows():
+    """It kills the process and reraises the keyboard interrupt."""
+    run_pytest_without_console_window("test_interrupt_raises")
+
+
+@skip_on_windows_console
 def test_interrupt_handled(command_with_keyboard_interrupt, marker):
     """It does not raise if the child handles the keyboard interrupt."""
     program = """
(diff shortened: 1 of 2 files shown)


cjolowicz:
The Windows CI fails due to an unexpected `KeyboardInterrupt` in the test `test_interrupt_handled`, where the interrupt should have been handled by the child process. The exception was raised when invoking `WaitForSingleObject`, in the context of `Popen.communicate`.

I have been able to reproduce this on a Windows system. In fact, both of the `test_interrupt_*` tests fail *some of the time* due to `KeyboardInterrupt` being raised in the parent process at seemingly random places: apart from `subprocess.py` (when invoking `WaitForSingleObject`), also at various locations in `typing.py` and `pytest`.

A possible explanation for this is an [SO answer](https://stackoverflow.com/a/35792192) by @eryksun:

> Python's os.kill wraps two unrelated APIs on Windows. It calls GenerateConsoleCtrlEvent when the sig parameter is CTRL_C_EVENT or CTRL_BREAK_EVENT. In this case the pid parameter is a process group ID. [...]

The test calls `os.kill` indirectly via `Popen.send_signal` from the our monkeypatched `Popen.communicate`, passing `signal.CTRL_C_EVENT`.

So this means that we send the ^C to ourselves as well, because the parent process is in the same process group as the child. At first, this sounds like we might be able to fix this by passing `subprocess.CREATE_NEW_PROCESS_GROUP` when creating the child process on Windows. However, reading on in the SO answer, this would also require calling `SetConsoleCtrlHandler(NULL, FALSE)` in the child, which we can't. My conclusion so far is that we should skip these tests on Windows.

My next step was to paste the Python programs from the tests into a Noxfile, and trigger an actual ^C from the console. There was no spurious `KeyboardInterrupt` in any of these runs. The results were quite interesting, though:

- The session without signal handlers behaves as expected: The process immediately terminates, and the session fails.
- The session with a signal handler exiting with zero also behaves as expected: The process immediately terminates, and the session succeeds.
- The session ignoring SIGINT, on the other hand, ignores not only the ^C, but also the `TerminateProcess` sent from Nox. The process completes its sleep, and the session succeeds.
- The session ignoring both SIGINT and SIGTERM behaves the same.

This another loose end that I haven't had time yet to investigate.

cjolowicz:
833d44f fixes the spurious interrupts in the test on Windows, leveraging the SO post quoted above. Unfortunately, even though all Nox sessions succeed in CI now, Nox mysteriously exits with 1. Haven't been able to reproduce this locally, on my test machine Nox exits with success.

theacodes:
This looks good to me so far, let me know if you get stuck on anything or need some help deciding on something.

cjolowicz:
72f7cd222066c3a0b85c1b16107bb61ba6999f08 fixes the spurious interrupts in the Windows *CI*. While Nox itself exited with success, its parent process (the one running the CI step) exited with failure. This happened because as a console process, the parent also received the keyboard interrupt triggered by the test. The fix was to simply not send a real interrupt when we're on Windows CI. This means we have to skip the test where the child handles the interrupt (`test_interrupt_handled`). For the other test (`test_interrupt_raises`), it does not matter. Outside of CI, both tests pass on Windows.

So this finally ready for review.

eryksun:
I'm not familiar with your specific needs, but here's some general advice.

If you need to test the effect of sending `CTRL_C_EVENT` (i.e. cancel) to a target process that doesn't manually enable Ctrl+C, then you can't spawn it in a new process group as the target of the console control event. You need to send the event to group 0, i.e. all processes that are attached to the current console session. However, doing that is unreliable for an arbitrary console session, since you don't know what other processes are attached and what will be the effect of sending them Ctrl+C. 

You can spawn the test in a new console session, allocated with the process creation flag `CREATE_NEW_CONSOLE`. Report the result back to the main test process over stdout or some other IPC channel. The test process in the new session can implement the test as follows:

* enable Ctrl+C to ensure it's inherited as such
* spawn the target process to be tested, *without* any of the following creation flags:
    * `CREATE_NEW_PROCESS_GROUP`
    * `CREATE_NEW_CONSOLE`
    * `CREATE_NO_WINDOW`
    * `DETACHED_PROCESS`
* wait for the target process to be ready
* disable Ctrl+C
* call `os.kill(0, CTRL_C_EVENT)` or `GenerateConsoleCtrlEvent(ctrl_event, pgid)`
* wait for the target process to exit

ctypes wrapper functions:

```python
import signal
import ctypes

kernel32 = ctypes.WinDLL('kernel32', use_last_error=True)

def enable_ctrl_c():
    if not kernel32.SetConsoleCtrlHandler(None, False):
        raise ctypes.WinError(ctypes.get_last_error())

def disable_ctrl_c():
    if not kernel32.SetConsoleCtrlHandler(None, True):
        raise ctypes.WinError(ctypes.get_last_error())

def send_ctrl_event(pgid=0, ctrl_event=signal.CTRL_C_EVENT):
    if not kernel32.GenerateConsoleCtrlEvent(ctrl_event, pgid):
        raise ctypes.WinError(ctypes.get_last_error())
```

cjolowicz:
Thank you @eryksun for taking the time to comment here. Your comment, as well as your posts on SO and bpo, were tremendously helpful!

dd9f3c2e22a83b7b5b04963e63f269775488edb6 implements the solution proposed above: On Windows, the keyboard interrupt tests are run in a subprocess that is not attached to the current console window. This means that all tests now run on Windows as well, both locally and in CI.

theacodes (approved):
This LGTM!

tests/test_command.py
```diff
@@ -23,6 +29,15 @@
 
 PYTHON = sys.executable
 
+skip_on_windows_console = pytest.mark.skipif(
+    platform.system() == "Windows" and "DETACHED_FROM_CONSOLE" not in os.environ,
+    reason="Do not run this test attached to a Windows console.",
+)
```
eryksun, line 35: The wording here is possibly misleading. `CREATE_NO_WINDOW` and `CREATE_NEW_CONSOLE` run the process attached to a new console session. The creation flag `DETACHED_PROCESS`, on the other hand, tells the system to spawn a process detached from any console session. A truly detached process is definitely not wanted here since the test needs to generate and receive a console control event.

FYI, in Windows 8.1+, a console session is a set of processes that are connected to an instance of conhost.exe by way of the ConDrv kernel device. The session window is optional and may be excluded because of `CREATE_NO_WINDOW` or, in Windows 10, because the console host is running in headless conpty mode (pseudoconsole). For example, each tab in Windows Terminal uses a headless conpty session, with an open-source build of "conhost.exe" that's named "openconsole.exe".

cjolowicz: Thank you for pointing that out. d6b0e59 changes the naming, hopefully for the better :wink:

cjolowicz:
@theacodes Thanks for the approval! I pushed two more commits:

- d6b0e59 improves naming in the tests, following @eryksun 's comment above
- 3df8312 ensures that the subprocess always uses the same `pytest` as the main process (even when it's not on `PATH`)

theacodes:
Cool, still looks good to me. Feel free to merge when you and @eryksun are happy. :)

theacodes:
Oh, I noticed you don't have merge permissions. Feel free to ping me when you're ready to merge (and if you want merge rights, let me know!)

cjolowicz:
> Feel free to ping me when you're ready to merge

This is good to go from my side.

>  (and if you want merge rights, let me know!)

Thanks for offering this, I'd be honored :)

theacodes:
> Thanks for offering this, I'd be honored :)

Done. I'll let you merge this :) 🎆 

This PR is eligible for $50 from our [open collective](https://opencollective.com/python-nox). Feel free to submit an expense for that if you'd like.

theacodes:
Huzzah!

PR #397, opened 2021-03-07
set local session posargs via notify

rmorshea:
Closes: https://github.com/theacodes/nox/issues/391

Once the basic API has been reviewed and accepted I'll add tests for this feature.

### DIFF
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -16,7 +16,18 @@ import argparse
 import collections.abc
 import itertools
 from collections import OrderedDict
-from typing import Any, Iterable, Iterator, List, Mapping, Sequence, Set, Tuple, Union
+from typing import (
+    Any,
+    Iterable,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Union,
+)
 
 from nox._decorators import Call, Func
 from nox.sessions import Session, SessionRunner
@@ -257,7 +268,9 @@ class Manifest:
     def next(self) -> SessionRunner:
         return self.__next__()
 
-    def notify(self, session: Union[str, SessionRunner]) -> bool:
+    def notify(
+        self, session: Union[str, SessionRunner], posargs: Optional[List[str]] = None
+    ) -> bool:
         """Enqueue the specified session in the queue.
 
         If the session is already in the queue, or has been run already,
@@ -282,6 +295,8 @@ class Manifest:
         # the end of the queue.
         for s in self._all_sessions:
             if s == session or s.name == session or session in s.signatures:
+                if posargs is not None:
+                    s.posargs = posargs
                 self._queue.append(s)
                 return True
 
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -279,6 +279,10 @@ class Manifest:
         Args:
             session (Union[str, ~nox.session.Session]): The session to be
                 enqueued.
+            posargs (Optional[List[str]]): If given, sets the positional
+                arguments *only* for the queued session. Otherwise, the
+                standard globally available positional arguments will be
+                used instead.
 
         Returns:
             bool: Whether the session was added to the queue.
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -433,7 +433,9 @@ class Session:
         self._run("python", "-m", "pip", "install", *args, external="error", **kwargs)
 
     def notify(
-        self, target: "Union[str, SessionRunner]", posargs: Optional[List[str]] = None
+        self,
+        target: "Union[str, SessionRunner]",
+        posargs: Optional[Iterable[str]] = None,
     ) -> None:
         """Place the given session at the end of the queue.
 
@@ -444,7 +446,13 @@ class Session:
             target (Union[str, Callable]): The session to be notified. This
                 may be specified as the appropriate string (same as used for
                 ``nox -s``) or using the function object.
+            posargs (Optional[Iterable[str]]): If given, sets the positional
+                arguments *only* for the queued session. Otherwise, the
+                standard globally available positional arguments will be
+                used instead.
         """
+        if posargs is not None:
+            posargs = list(posargs)
         self._runner.manifest.notify(target, posargs)
 
     def log(self, *args: Any, **kwargs: Any) -> None:
@@ -468,26 +476,14 @@ class SessionRunner:
         func: Func,
         global_config: argparse.Namespace,
         manifest: "Manifest",
-        posargs: Optional[List[str]] = None,
     ) -> None:
         self.name = name
         self.signatures = signatures
         self.func = func
         self.global_config = global_config
         self.manifest = manifest
-        self.venv = None  # type: Optional[ProcessEnv]
-        self._posargs = posargs
-
-    @property
-    def posargs(self) -> List[str]:
-        if self._posargs is not None:
-            return self._posargs
-        else:
-            return self.global_config.posargs
-
-    @posargs.setter
-    def posargs(self, value: Optional[List[str]]) -> None:
-        self._posargs = value
+        self.venv: Optional[ProcessEnv] = None
+        self.posargs: List[str] = global_config.posargs
 
     @property
     def description(self) -> Optional[str]:
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -149,8 +149,7 @@ class Session:
 
     @property
     def posargs(self) -> List[str]:
-        """This is set to any extra arguments
-        passed to ``nox`` on the commandline."""
+        """Any extra arguments from the ``nox`` commandline or :class:`Session.notify`"""
         return self._runner.posargs
 
     @property
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -151,7 +151,7 @@ class Session:
     def posargs(self) -> List[str]:
         """This is set to any extra arguments
         passed to ``nox`` on the commandline."""
-        return self._runner.global_config.posargs
+        return self._runner.posargs
 
     @property
     def virtualenv(self) -> ProcessEnv:
@@ -432,7 +432,9 @@ class Session:
 
         self._run("python", "-m", "pip", "install", *args, external="error", **kwargs)
 
-    def notify(self, target: "Union[str, SessionRunner]") -> None:
+    def notify(
+        self, target: "Union[str, SessionRunner]", posargs: Optional[List[str]] = None
+    ) -> None:
         """Place the given session at the end of the queue.
 
         This method is idempotent; multiple notifications to the same session
@@ -443,7 +445,7 @@ class Session:
                 may be specified as the appropriate string (same as used for
                 ``nox -s``) or using the function object.
         """
-        self._runner.manifest.notify(target)
+        self._runner.manifest.notify(target, posargs)
 
     def log(self, *args: Any, **kwargs: Any) -> None:
         """Outputs a log during the session."""
@@ -466,6 +468,7 @@ class SessionRunner:
         func: Func,
         global_config: argparse.Namespace,
         manifest: "Manifest",
+        posargs: Optional[List[str]] = None,
     ) -> None:
         self.name = name
         self.signatures = signatures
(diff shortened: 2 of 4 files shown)


nox/sessions.py
```diff
@@ -432,7 +432,9 @@ def install(self, *args: str, **kwargs: Any) -> None:
 
         self._run("python", "-m", "pip", "install", *args, external="error", **kwargs)
 
-    def notify(self, target: "Union[str, SessionRunner]") -> None:
+    def notify(
+        self, target: "Union[str, SessionRunner]", posargs: Optional[List[str]] = None
```
cjolowicz, line 436: Maybe choose a generic argument type for the public API?

```suggestion
        self,
        target: "Union[str, SessionRunner]",
        posargs: Optional[Iterable[str]] = None,
```

cjolowicz (commented):
Looks great, I like the approach with the property setter :) Public API stays read-only, which is good. I left some minor suggestions inline, what do you think? Otherwise, I think this just needs  tests and an update to the docstring for `session.notify`.

nox/manifest.py
```diff
@@ -257,7 +268,9 @@ def make_session(
     def next(self) -> SessionRunner:
         return self.__next__()
 
-    def notify(self, session: Union[str, SessionRunner]) -> bool:
+    def notify(
+        self, session: Union[str, SessionRunner], posargs: Optional[List[str]] = None
```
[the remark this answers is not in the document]
rmorshea: Good point, I'll make this change.

nox/manifest.py
```diff
@@ -282,6 +295,8 @@ def notify(self, session: Union[str, SessionRunner]) -> bool:
         # the end of the queue.
         for s in self._all_sessions:
             if s == session or s.name == session or session in s.signatures:
+                if posargs is not None:
+                    s.posargs = posargs
```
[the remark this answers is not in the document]
rmorshea: I think I'll probably just cast it to a list in `Session.notify` since it'll make the internal API a little simpler if you can make some more strict assumptions about the type.

rmorshea:
@cjolowicz assuming tests pass I think this should be good now.

nox/sessions.py
```diff
@@ -472,7 +481,8 @@ def __init__(
         self.func = func
         self.global_config = global_config
         self.manifest = manifest
-        self.venv = None  # type: Optional[ProcessEnv]
+        self.venv: Optional[ProcessEnv] = None
+        self.posargs: List[str] = global_config.posargs
```
cjolowicz, line 485: Should we inoculate `global_config.posargs` against mutation here?

```suggestion
        self.posargs: List[str] = global_config.posargs[:]
```

Now that every session conceptually has their own `posargs`, I think it makes sense to prevent one session from accidentally mutating the default `posargs` of other sessions.

cjolowicz (commented):
Thanks @rmorshea this is pretty much good to go :) I just noticed a few more small things, could you take another look? Comments are inline.

nox/sessions.py
cjolowicz: Oh dear, this breaks a whole lot of tests :sweat_smile: 

Essentially, there are three groups of tests that break:

- Tests that compare posargs using `is` instead of `==`
- Tests that construct a namespace manually without including posargs
- Tests that construct a namespace manually and set posargs to a mock sentinel

<details>
<summary>Here's a patch that would make this green again.</summary>

(Some of these could be turned into fixtures.)

```diff
diff --git a/tests/test__option_set.py b/tests/test__option_set.py
index 3ad8b59..d05bd7f 100644
--- a/tests/test__option_set.py
+++ b/tests/test__option_set.py
@@ -55,7 +55,7 @@ class TestOptionSet:
             optionset.namespace(non_existant_option="meep")
 
     def test_session_completer(self):
-        parsed_args = _options.options.namespace(sessions=(), keywords=())
+        parsed_args = _options.options.namespace(sessions=(), keywords=(), posargs=[])
         all_nox_sessions = _options._session_completer(
             prefix=None, parsed_args=parsed_args
         )
@@ -65,7 +65,9 @@ class TestOptionSet:
         assert len(set(some_expected_sessions) - set(all_nox_sessions)) == 0
 
     def test_session_completer_invalid_sessions(self):
-        parsed_args = _options.options.namespace(sessions=("baz",), keywords=())
+        parsed_args = _options.options.namespace(
+            sessions=("baz",), keywords=(), posargs=[]
+        )
         all_nox_sessions = _options._session_completer(
             prefix=None, parsed_args=parsed_args
         )
diff --git a/tests/test_manifest.py b/tests/test_manifest.py
index 35eb775..b0c3ac9 100644
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -354,7 +354,7 @@ def test_notify_with_posargs():
     # delete my_session from the queue
     manifest.filter_by_name(())
 
-    assert session.posargs is cfg.posargs
+    assert session.posargs == cfg.posargs
     assert manifest.notify("my_session", posargs=["--an-arg"])
     assert session.posargs == ["--an-arg"]
 
diff --git a/tests/test_sessions.py b/tests/test_sessions.py
index 150c15e..518ab0c 100644
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -66,9 +66,7 @@ class TestSession:
             signatures=["test"],
             func=func,
             global_config=_options.options.namespace(
-                posargs=mock.sentinel.posargs,
-                error_on_external_run=False,
-                install_only=False,
+                posargs=[], error_on_external_run=False, install_only=False
             ),
             manifest=mock.create_autospec(nox.manifest.Manifest),
         )
@@ -100,7 +98,7 @@ class TestSession:
 
         assert session.name is runner.friendly_name
         assert session.env is runner.venv.env
-        assert session.posargs is runner.global_config.posargs
+        assert session.posargs == runner.global_config.posargs
         assert session.virtualenv is runner.venv
         assert session.bin_paths is runner.venv.bin_paths
         assert session.bin is runner.venv.bin_paths[0]
@@ -351,7 +349,7 @@ class TestSession:
             name="test",
             signatures=["test"],
             func=mock.sentinel.func,
-            global_config=_options.options.namespace(posargs=mock.sentinel.posargs),
+            global_config=_options.options.namespace(posargs=[]),
             manifest=mock.create_autospec(nox.manifest.Manifest),
         )
         runner.venv = mock.create_autospec(nox.virtualenv.CondaEnv)
@@ -390,7 +388,7 @@ class TestSession:
             name="test",
             signatures=["test"],
             func=mock.sentinel.func,
-            global_config=_options.options.namespace(posargs=mock.sentinel.posargs),
+            global_config=_options.options.namespace(posargs=[]),
             manifest=mock.create_autospec(nox.manifest.Manifest),
         )
         runner.venv = mock.create_autospec(nox.virtualenv.CondaEnv)
@@ -447,7 +445,7 @@ class TestSession:
             name="test",
             signatures=["test"],
             func=mock.sentinel.func,
-            global_config=_options.options.namespace(posargs=mock.sentinel.posargs),
+            global_config=_options.options.namespace(posargs=[]),
             manifest=mock.create_autospec(nox.manifest.Manifest),
         )
         runner.venv = mock.create_autospec(nox.virtualenv.VirtualEnv)
@@ -476,7 +474,7 @@ class TestSession:
             name="test",
             signatures=["test"],
             func=mock.sentinel.func,
-            global_config=_options.options.namespace(posargs=mock.sentinel.posargs),
+            global_config=_options.options.namespace(posargs=[]),
             manifest=mock.create_autospec(nox.manifest.Manifest),
         )
         runner.venv = mock.create_autospec(nox.virtualenv.VirtualEnv)
@@ -564,7 +562,7 @@ class TestSessionRunner:
             global_config=_options.options.namespace(
                 noxfile=os.path.join(os.getcwd(), "noxfile.py"),
                 envdir="envdir",
-                posargs=mock.sentinel.posargs,
+                posargs=[],
                 reuse_existing_virtualenvs=False,
                 error_on_missing_interpreters=False,
             ),
@@ -580,7 +578,7 @@ class TestSessionRunner:
         assert runner.func is not None
         assert callable(runner.func)
         assert isinstance(runner.description, str)
-        assert runner.global_config.posargs == mock.sentinel.posargs
+        assert runner.global_config.posargs == []
         assert isinstance(runner.manifest, nox.manifest.Manifest)
 
     def test_str_and_friendly_name(self):
diff --git a/tests/test_tasks.py b/tests/test_tasks.py
index 65f3112..ca477ff 100644
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -139,7 +139,7 @@ def test_discover_session_functions_decorator():
     mock_module = argparse.Namespace(
         __name__=foo.__module__, foo=foo, bar=bar, notasession=notasession
     )
-    config = _options.options.namespace(sessions=(), keywords=())
+    config = _options.options.namespace(sessions=(), keywords=(), posargs=())
 
     # Get the manifest and establish that it looks like what we expect.
     manifest = tasks.discover_manifest(mock_module, config)
@@ -149,7 +149,9 @@ def test_discover_session_functions_decorator():
 
 
 def test_filter_manifest():
-    config = _options.options.namespace(sessions=(), pythons=(), keywords=())
+    config = _options.options.namespace(
+        sessions=(), pythons=(), keywords=(), posargs=()
+    )
     manifest = Manifest({"foo": session_func, "bar": session_func}, config)
     return_value = tasks.filter_manifest(manifest, config)
     assert return_value is manifest
@@ -157,14 +159,18 @@ def test_filter_manifest():
 
 
 def test_filter_manifest_not_found():
-    config = _options.options.namespace(sessions=("baz",), pythons=(), keywords=())
+    config = _options.options.namespace(
+        sessions=("baz",), pythons=(), keywords=(), posargs=()
+    )
     manifest = Manifest({"foo": session_func, "bar": session_func}, config)
     return_value = tasks.filter_manifest(manifest, config)
     assert return_value == 3
 
 
 def test_filter_manifest_pythons():
-    config = _options.options.namespace(sessions=(), pythons=("3.8",), keywords=())
+    config = _options.options.namespace(
+        sessions=(), pythons=("3.8",), keywords=(), posargs=()
+    )
     manifest = Manifest(
         {"foo": session_func_with_python, "bar": session_func, "baz": session_func},
         config,
@@ -175,7 +181,9 @@ def test_filter_manifest_pythons():
 
 
 def test_filter_manifest_keywords():
-    config = _options.options.namespace(sessions=(), pythons=(), keywords="foo or bar")
+    config = _options.options.namespace(
+        sessions=(), pythons=(), keywords="foo or bar", posargs=()
+    )
     manifest = Manifest(
         {"foo": session_func, "bar": session_func, "baz": session_func}, config
     )
@@ -230,7 +238,7 @@ def test_verify_manifest_empty():
 
 
 def test_verify_manifest_nonempty():
-    config = _options.options.namespace(sessions=(), keywords=())
+    config = _options.options.namespace(sessions=(), keywords=(), posargs=())
     manifest = Manifest({"session": session_func}, config)
     return_value = tasks.verify_manifest_nonempty(manifest, global_config=config)
     assert return_value == manifest
```
</details>

I think we can also choose to be pragmatic here, and keep this as it is. After all, the mutability of `session.posargs` is undocumented and an implementation detail, and should not be used in Noxfiles. What do you think?

@theacodes What's your take on this?

rmorshea: While we're thinking about this breaking change we should also consider making `posargs` a tuple. Modifying `posargs` at all just seems ill-advised.

cjolowicz (approved):
LGTM

I'd be happy for this to be merged without making `session.posargs` immutable. (To be clear it's already mutable on the main branch, this PR does not change that.) We can deal with that in a separate PR if required.

nox/sessions.py
cjolowicz: I agree that a tuple would be preferable. (Changing the type might not even constitute a breaking change; it's not documented and most Noxfiles probably just unpack into `session.run`). But I think we can merge this without addressing the issue here.

rmorshea: I'll create an issue to capture this.

rmorshea: See here: https://github.com/theacodes/nox/issues/400

rmorshea:
@cjolowicz I think this is good to merge.

rmorshea:
ping: @cjolowicz - I think this is good to merge

theacodes:
Looks great, thanks, y'all!

PR #402, opened 2021-03-11
Ignore mypy configuration file

paw-lu:
Nox does not have a `mypy.ini` or `setup.cfg`. This will cause mypy to look for configuration files [in other directories](https://mypy.readthedocs.io/en/stable/config_file.html), which can cause failures when testing locally. This PR adds the `--config-file=` argument, which tells mypy to ignore config files.

---

Thanks!

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -90,6 +90,7 @@ def lint(session):
     session.install("flake8==3.7.8", "black==19.3b0", "isort==4.3.21", "mypy==0.720")
     session.run(
         "mypy",
+        "--config-file=",
         "--disallow-untyped-defs",
         "--warn-unused-ignores",
         "--ignore-missing-imports",


cjolowicz:
Thanks for catching this @paw-lu! This LGTM, I'm just wondering if we shouldn't move the mypy command-line options from the Noxfile to our own mypy.ini.

cjolowicz (commented):
Thanks for catching this @paw-lu! This LGTM, I'm just wondering if we shouldn't move the mypy command-line options from the Noxfile to our own mypy.ini.

theacodes:
>  Thanks for catching this @paw-lu! This LGTM, I'm just wondering if we shouldn't move the mypy command-line options from the Noxfile to our own mypy.ini.

Maybe, but I think for now we're fine. It'd be a nice follow-up if someone wants to tackle that.

PR #404, opened 2021-03-14
Allow nox.parametrize to select the session Python

### LINKED ISSUE #392: Exclude combinations from parametrized matrix builds
I'm using the `@parametrize` decorator to create a build matrix, but i've got certain combinations I wish to exclude.

right now i'm using a little escape hatch

```python
if valid_combination(*deps):
    run_tests(session, *deps)
else:
    session.skip("unsupported combination")
```

would be nice if Nox could provide an API for adding or removing 'cells' from the matrix (similarly to what you can do with github actions)

cjolowicz:
Fixes #392

Here's a first go at allowing `@nox.parametrize` to select the Python interpreter for a session.

### DIFF
--- a/nox/_decorators.py
+++ b/nox/_decorators.py
@@ -1,7 +1,7 @@
 import copy
 import functools
 import types
-from typing import Any, Callable, Dict, Iterable, List, Optional, cast
+from typing import AbstractSet, Any, Callable, Dict, Iterable, List, Optional, cast
 
 from . import _typing
 
@@ -65,10 +65,20 @@ class Func(FunctionDecorator):
 
 
 class Call(Func):
-    def __init__(self, func: Func, param_spec: "Param") -> None:
+    def __init__(
+        self, func: Func, param_spec: "Param", sessionparams: AbstractSet[str]
+    ) -> None:
+        call_spec = param_spec.call_spec
+
+        for sessionparam in sessionparams & call_spec.keys() & {"python", "py"}:
+            python = call_spec[sessionparam]
+            break
+        else:
+            python = func.python
+
         super().__init__(
             func,
-            func.python,
+            python,
             func.reuse_venv,
             None,
             func.venv_backend,
@@ -76,12 +86,19 @@ class Call(Func):
             func.should_warn,
         )
         self.param_spec = param_spec
+        self.sessionparams = sessionparams
         self.session_signature = "({})".format(param_spec)
 
     def __call__(self, *args: Any, **kwargs: Any) -> Any:
-        kwargs.update(self.param_spec.call_spec)
+        kwargs.update(
+            (key, value)
+            for key, value in self.param_spec.call_spec.items()
+            if key not in self.sessionparams
+        )
         return super().__call__(*args, **kwargs)
 
     @classmethod
-    def generate_calls(cls, func: Func, param_specs: "Iterable[Param]") -> "List[Call]":
-        return [cls(func, param_spec) for param_spec in param_specs]
+    def generate_calls(
+        cls, func: Func, param_specs: "Iterable[Param]", sessionparams: AbstractSet[str]
+    ) -> "List[Call]":
+        return [cls(func, param_spec, sessionparams) for param_spec in param_specs]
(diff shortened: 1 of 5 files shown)


nox/_decorators.py
```diff
@@ -1,7 +1,7 @@
 import copy
 import functools
 import types
-from typing import Any, Callable, Dict, Iterable, List, Optional, cast
+from typing import AbstractSet, Any, Callable, Dict, Iterable, List, Optional, cast
```
smarie, line 4: Very minor nitpick: note that `AbstractSet` is deprecated [in 3.9](https://docs.python.org/3/library/typing.html#typing.AbstractSet)

nox/_parametrize.py
```diff
@@ -78,10 +91,14 @@ def __eq__(self, other: object) -> bool:
         raise NotImplementedError
 
 
-def _apply_param_specs(param_specs: List[Param], f: Any) -> Any:
+def _apply_param_specs(
+    param_specs: List[Param], sessionparams: AbstractSet[str], f: Any
+) -> Any:
     previous_param_specs = getattr(f, "parametrize", None)
     new_param_specs = update_param_specs(previous_param_specs, param_specs)
+    sessionparams |= getattr(f, "sessionparams", frozenset())
```
smarie, line 99: Just curious: is this equivalent to 

```suggestion
    sessionparams.update(getattr(f, "sessionparams", frozenset()))
```

? If so readability might suggest the latter but this is maybe because I am an old pythonista already :)

smarie:
Only a few nitpicks on my side, otherwise it looks great ! Thanks @cjolowicz

nox/_decorators.py
cjolowicz: Thank you for taking the time to review this.

`typing.AbstractSet` is superceded by `collections.abc.Set` in 3.9, but this won't work in earlier Python versions where the abstract collections are not yet generics, i.e. you can't write `Set[str]`. The same deprecation affects Callable and Iterable; Dict and List are also deprecated in favor of `dict` and `list`.

I'm not sure how to address this, do you have an idea? It would be interesting to see how other projects deal with the deprecation. I'm hesitant to introduce a large amount of compatibility boilerplate.

nox/_parametrize.py
cjolowicz: Join the club :wink: Yes, the two are almost synonymous, except that `update` accepts any iterable; both have been around since [PEP 218] introduced sets. I'm happy to change this if it's confusing. Personally, I find that using set union and assignment instead of a single mutating function call makes the code more declarative and idiomatic, and avoiding parentheses improves readability.

[PEP 218]: https://www.python.org/dev/peps/pep-0218/

nox/_decorators.py
smarie: For what's worth, in `pytest` the contract when `ids` are explicitly provided is to receive an `Iterable[str]` (and to transform it into a list or set internally so as to consume possible generators). Maybe this could be sufficient ?

nox/_parametrize.py
smarie: I learnt something today thanks ! :) Let's let the core nox team decide then.

nox/_decorators.py
cjolowicz: The public API does use `Iterable[str]`:

```python
# _parametrize.py
def parametrize_decorator(
    arg_names: Union[str, List[str], Tuple[str]],
    arg_values_list: Union[Iterable[ArgValue], ArgValue],
    ids: Optional[Iterable[Optional[str]]] = None,
    sessionparams: Optional[Iterable[str]] = None,
) -> Callable[[Any], Any]:
    ...
    sessionparams = frozenset(sessionparams if sessionparams is not None else [])
    ...
```

We only use `AbstractSet` internally.

smarie: oh sorry I read too fast ! perfect then.

cjolowicz:
Superceded by #413 

Happy to reopen this if people prefer this approach to the one in #413.

PR #405, opened 2021-03-14
Adds option to skip Session dependency installs. Default is False. (#…

jamandbees:
* skips pip installs
  * skips conda installs
  * skips run_always if no-install is set

-----
* I had problems figuring out how to actually run my local codebase, so I didn't get to do a sanity check on the command line. I'm still working on this...
* I wasn't entirely sure my error message in run_always was a good one


jamandbees:
Closing as the PR advice was simpler to implement on a new branch.

PR #411, opened 2021-03-27
394 integrates no install flag

### LINKED ISSUE #394: Add an option to skip install commands
Frequently you just want to re-run a particular session in a reused virtual environment as quickly as possible (e.g for unit tests, docs or linting). This is possible with `--reuse-existing-virtualenvs`, but `session.install()` still takes a second or two, which is not a lot, but still slower than just running `pytest` directly. Therefore, it would be useful to have an option for skipping `session.install()`, especially when executing sessions locally in a development loop.

Add an inverse of the `--install-only` command.

* Use [_invoke_](https://www.pyinvoke.org/) within a manually created virtual environment (this is what I've been doing).
* Just use a corresponding Python interpreter directly, e.g. `.nox/<session_name>/bin/python -m pytest`.
* Accept the 1-2 second overhead, and take your time to sip on your favourite beverage :)

jamandbees:
* Adds a new option, `--no-install`. This is used to skip installing packages if the packages have been installed once.
* Adds a boolean to the virtualenv base class, `venv_created`, which tracks whether or not the virtual environment has already been created. The `--no-install` flag behaves differently dependent upon whether a venv is already created or not.
  * If no virtualenv is present, a message is logged that `no-install` is being ignored and normal installations proceed.
  * If a virtualenv is already present, `--no-install` prevents the installation of packages.
* Adds tests to validate the above 
* Also refactors `SessionNoSlots` into an extracted class `MockableSession`. This followed from scratching my head trying to understand what SessionNoSlots provided, and then heading to the PR which implemented it (https://github.com/theacodes/nox/pull/128) to see what the class was buying the tests. I'm happy to put this back how I found it 😅

Closes #394

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -424,6 +424,7 @@ The following options can be specified in the Noxfile:
 * ``nox.options.default_venv_backend`` is equivalent to specifying :ref:`-db or --default-venv-backend <opt-default-venv-backend>`.
 * ``nox.options.force_venv_backend`` is equivalent to specifying :ref:`-fb or --force-venv-backend <opt-force-venv-backend>`.
 * ``nox.options.reuse_existing_virtualenvs`` is equivalent to specifying :ref:`--reuse-existing-virtualenvs <opt-reuse-existing-virtualenvs>`. You can force this off by specifying ``--no-reuse-existing-virtualenvs`` during invocation.
+* ``nox.options.no_install`` is equivalent to specifying :ref:`--no-install <opt-no-install>`.
 * ``nox.options.stop_on_first_error`` is equivalent to specifying :ref:`--stop-on-first-error <opt-stop-on-first-error>`. You can force this off by specifying ``--no-stop-on-first-error`` during invocation.
 * ``nox.options.error_on_missing_interpreters`` is equivalent to specifying :ref:`--error-on-missing-interpreters <opt-error-on-missing-interpreters>`. You can force this off by specifying ``--no-error-on-missing-interpreters`` during invocation.
 * ``nox.options.error_on_external_run`` is equivalent to specifying :ref:`--error-on-external-run <opt-error-on-external-run>`. You can force this off by specifying ``--no-error-on-external-run`` during invocation.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -424,7 +424,6 @@ The following options can be specified in the Noxfile:
 * ``nox.options.default_venv_backend`` is equivalent to specifying :ref:`-db or --default-venv-backend <opt-default-venv-backend>`.
 * ``nox.options.force_venv_backend`` is equivalent to specifying :ref:`-fb or --force-venv-backend <opt-force-venv-backend>`.
 * ``nox.options.reuse_existing_virtualenvs`` is equivalent to specifying :ref:`--reuse-existing-virtualenvs <opt-reuse-existing-virtualenvs>`. You can force this off by specifying ``--no-reuse-existing-virtualenvs`` during invocation.
-* ``nox.options.no_install`` is equivalent to specifying :ref:`--no-install <opt-no-install>`.
 * ``nox.options.stop_on_first_error`` is equivalent to specifying :ref:`--stop-on-first-error <opt-stop-on-first-error>`. You can force this off by specifying ``--no-stop-on-first-error`` during invocation.
 * ``nox.options.error_on_missing_interpreters`` is equivalent to specifying :ref:`--error-on-missing-interpreters <opt-error-on-missing-interpreters>`. You can force this off by specifying ``--no-error-on-missing-interpreters`` during invocation.
 * ``nox.options.error_on_external_run`` is equivalent to specifying :ref:`--error-on-external-run <opt-error-on-external-run>`. You can force this off by specifying ``--no-error-on-external-run`` during invocation.
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -291,7 +291,6 @@ class Session:
         :type external: bool
         """
         if self._runner.global_config.no_install:
-            logger.info("Skipping run_always, as --no-install is set.")
             return None
 
         if not args:
@@ -373,10 +372,6 @@ class Session:
             raise ValueError("At least one argument required to install().")
 
         if self._runner.global_config.no_install and venv._reused:
-            logger.info(
-                "Venv exists: skipping conda installation, as --no-install is set."
-            )
-
             return None
 
         # Escape args that should be (conda-specific; pip install does not need this)
@@ -440,7 +435,6 @@ class Session:
             raise ValueError("At least one argument required to install().")
 
         if self._runner.global_config.no_install and venv._reused:
-            logger.info("Venv exists: skipping installation, as --no-install is set.")
             return None
 
         if "silent" not in kwargs:
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -374,11 +374,13 @@ class Session:
 
         if self._runner.global_config.no_install:
             if venv._reused:
-                logger.info("Skipping conda installation, as --no-install is set.")
+                logger.info(
+                    "Venv exists: skipping conda installation, as --no-install is set."
+                )
                 return None
             else:
                 logger.info(
-                    "Venv not created yet. Ignoring --no-install and installing from conda."
+                    "Venv not created yet: ignoring --no-install and installing from conda."
                 )
 
         # Escape args that should be (conda-specific; pip install does not need this)
@@ -443,11 +445,13 @@ class Session:
 
         if self._runner.global_config.no_install:
             if venv._reused:
-                logger.info("Skipping installation, as --no-install is set.")
+                logger.info(
+                    "Venv exists: skipping installation, as --no-install is set."
+                )
                 return None
             else:
                 logger.info(
-                    "Venv not created yet. Ignoring --no-install and installing."
+                    "Venv not created yet: ignoring --no-install and installing."
                 )
 
         if "silent" not in kwargs:
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -536,7 +536,7 @@ class SessionRunner:
             return
 
         reuse_existing = (
-            self.func.reuse_venv or self.global_config.reuse_existing_virtualenvs
+            self.func.reuse_venv or self.global_config.reuse_existing_virtualenvs or self.global_config.no_install
         )
 
         if backend is None or backend == "virtualenv":
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -375,7 +375,7 @@ class Session:
             raise ValueError("At least one argument required to install().")
 
         if self._runner.global_config.no_install:
-            if (venv.venv_created):
+            if venv.venv_created:
                 logger.info(
                     "Skipping {} conda installation, as --no-install is set.".format(
                         args[0]
@@ -451,7 +451,7 @@ class Session:
             raise ValueError("At least one argument required to install().")
 
         if self._runner.global_config.no_install:
-            if(venv.venv_created):
+            if venv.venv_created:
                 logger.info(
                     "Skipping {} installation, as --no-install is set.".format(args[0])
                 )
@@ -554,7 +554,9 @@ class SessionRunner:
             return
 
         reuse_existing = (
-            self.func.reuse_venv or self.global_config.reuse_existing_virtualenvs or self.global_config.no_install
+            self.func.reuse_venv
+            or self.global_config.reuse_existing_virtualenvs
+            or self.global_config.no_install
         )
 
         if backend is None or backend == "virtualenv":
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -360,15 +360,22 @@ class Session:
         .. _conda install:
         """
 
-        if self._runner.global_config.no_install:
-            logger.info(
-                "Skipping {} conda installation, as --no-install is set.".format(
-                    args[0]
-                )
-            )
-            return None
-
         venv = self._runner.venv
+
+        if self._runner.global_config.no_install:
+            if (venv.venv_created):
+                logger.info(
+                    "Skipping {} conda installation, as --no-install is set.".format(
+                        args[0]
+                    )
+                )
+                return None
+            else:
+                logger.info(
+                    "Venv not created yet. Ignoring --no-install and installing {} via conda.".format(
+                        args[0]
+                    )
+                )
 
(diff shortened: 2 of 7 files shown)


nox/sessions.py
```diff
@@ -368,6 +374,21 @@ def conda_install(
         if not args:
             raise ValueError("At least one argument required to install().")
 
+        if self._runner.global_config.no_install:
+            if (venv.venv_created):
+                logger.info(
+                    "Skipping {} conda installation, as --no-install is set.".format(
+                        args[0]
+                    )
```
cjolowicz, line 382: Unlike `session.run`, the first argument is not special. It could be a package or an option. I'd suggest to omit this from the message:

```suggestion
                    "Skipping conda installation, as --no-install is set."
```

nox/sessions.py
```diff
@@ -368,6 +374,21 @@ def conda_install(
...
+                )
+                return None
+            else:
+                logger.info(
+                    "Venv not created yet. Ignoring --no-install and installing {} via conda.".format(
+                        args[0]
+                    )
```
cjolowicz, line 389: See above about omitting `args[0]`.

nox/sessions.py
```diff
@@ -417,8 +438,26 @@ def install(self, *args: str, **kwargs: Any) -> None:
...
+            if(venv.venv_created):
+                logger.info(
+                    "Skipping {} installation, as --no-install is set.".format(args[0])
+                )
+                return None
+            else:
+                logger.info(
+                    "Venv not created yet. Ignoring --no-install and installing {}.".format(
+                        args[0]
+                    )
```
cjolowicz, line 455: See above about omitting `args[0]`.

nox/virtualenv.py
```diff
@@ -249,6 +256,8 @@ def create(self) -> bool:
         )
         nox.command.run(cmd, silent=True, log=False)
 
+        self.venv_created = True
```
cjolowicz, line 259: Typo:

```suggestion
        self._venv_created = True
```

cjolowicz (commented):
Hi @jamandbees thanks for the new PR :+1:

Can you please run `nox -s blacken` in your branch? This should take care of the failed lint session in CI.

I left a few more quick comments inline.

nox/sessions.py
jamandbees: Thank you! I saw a package name in my tests and didn't reason much about what args[0] could actually be.

jamandbees:
> Hi @jamandbees thanks for the new PR 👍
> 
> Can you please run `nox -s blacken` in your branch? This should take care of the failed lint session in CI.
> 
> I left a few more quick comments inline.

I didn't make this clear, but linting etc are all passing now. 🙂

noxfile.py
```diff
@@ -97,7 +97,7 @@ def lint(session):
         "nox",
     )
     files = ["nox", "tests", "noxfile.py", "setup.py"]
-    session.run("black", "--check", *files)
+    session.run("black", *files)
```
cjolowicz, line 100: I think this was committed by mistake?

```suggestion
    session.run("black", "--check", *files)
```

nox/sessions.py
```diff
@@ -368,6 +374,15 @@ def conda_install(
...
             raise ValueError("At least one argument required to install().")
 
+        if self._runner.global_config.no_install:
+            if venv._venv_created:
+                logger.info("Skipping conda installation, as --no-install is set.")
+                return None
+            else:
+                logger.info(
+                    "Venv not created yet. Ignoring --no-install and installing from conda."
+                )
```
cjolowicz, line 384: I think the condition is inverted here. If Nox created the virtualenv afresh, then we should *not* skip installation.

```suggestion
            if venv._venv_created:
                logger.info(
                    "Venv not created yet. Ignoring --no-install and installing from conda."
                )
            else:
                logger.info("Skipping conda installation, as --no-install is set.")
                return None
```

nox/sessions.py
```diff
@@ -417,15 +432,27 @@ def install(self, *args: str, **kwargs: Any) -> None:
...
             raise ValueError("At least one argument required to install().")
 
+        if self._runner.global_config.no_install:
+            if venv._venv_created:
+                logger.info("Skipping installation, as --no-install is set.")
+                return None
+            else:
+                logger.info(
+                    "Venv not created yet. Ignoring --no-install and installing."
+                )
```
cjolowicz, line 454: See above about inverted condition.

nox/sessions.py
```diff
@@ -505,6 +532,7 @@ def envdir(self) -> str:
         return _normalize_path(self.global_config.envdir, self.friendly_name)
 
     def _create_venv(self) -> None:
+
```
cjolowicz, line 535: Stray blankline. I think there are a few more, can you take a look?

```suggestion
```

cjolowicz (commented):
Thanks, looking nice! Some more comments inline.

noxfile.py
jamandbees: How embarrassing...😅

Thanks for catching that!!

nox/sessions.py
jamandbees: Yes, I see what you're saying. 

I have been conceptualising the goal as using `ProcessEnv._venv_created` to determine whether to skip `session.install`. However, I had been misthinking of nox as a stateful app, and I had only been considering the _first_ run of a set of nox runs. Reading your suggestion, I think that it's runs 2...N that we're actually trying to optimise.

That's to say, what we're trying to do is figure out if the venvs already exist or not.  Now that I re-read section 2 of the comment at https://github.com/theacodes/nox/issues/394#issuecomment-799776880, you had said the problem this variable is addressing is "what we should do if an environment does not exist yet. ", so I think you had understood this earlier than I had.

What you're suggesting here makes sense, but I think it's showing that _both the variable name I had picked is a bad one, _and_ that the variable is tracking the wrong concept_ 🙂. I think we're not interested in whether a venv was created in this session, we're interested in whether a venv was created in _any_ run of nox.

The `--reuse-existing-virtualenvs` flag solves this by checking the file system to see whether the venv already exists before it attempts a delete. This is from virtualenv.py line 198 in my branch:

```
    def _clean_location(self) -> bool:
        """Deletes existing conda environment"""
        if os.path.exists(self.location): #<--- checking for state!
            if self.reuse_existing:
                return False
            else:
                cmd = ["conda", "remove", "--yes", "--prefix", self.location, "--all"]
```

Should `_venv_created` maybe have its implementation changed to something like:

```
self._venv_created = os.path.exists(self.location)
```

This way, it's checking the state generated by previous runs of nox, instead of just assuming that a venv was not created when it's initialised. With that change, I think my current implementation would make sense.

What do you think?

cjolowicz: Just in case it got lost in the to and fro of this review, I'll repeat my earlier suggestion from https://github.com/theacodes/nox/issues/394#issuecomment-801789356:

> The virtualenv needs to remember if it was reused. This could be a bool attribute `ProcessEnv._reused`, initially False. The VirtualEnv and CondaEnv implementations can set this to True in their `create` methods when the virtualenv gets reused. The logic is already there: These methods call `self._clean_location` which returns False if the virtualenv was reused.

(Edited to rename `reused` to `_reused` to avoid adding to the public API.)

Personally, I think the name `reused` avoids precisely the misunderstanding you described in your post above.

> Should `_venv_created` maybe have its implementation changed to something like:
>
> ```
> self._venv_created = os.path.exists(self.location)
> ```

Maybe I'm missing something here, but I don't see how this would work. I presume you want this code before we create the virtualenv, i.e. at the start of `VirtualEnv.create` and `CondaEnv.create`. But even if the virtualenv exists then, it may get removed and created from scratch. Instead you need the check that is already done inside `_clean_location`: `os.path.exists(self.location) and self.reuse_existing`. Instead of copying this check, I would recommend to use the return value of `_clean_location`, as described above.

jamandbees: I'm went ahead with the renaming from _venv_created to _reused because it was simple, but I'll come back to this later to finish up your other suggestions.

You were right, I had missed the advice around `self._clean_location`. Thanks for bringing it back up!

jamandbees: @cjolowicz Okay! I renamed `_venv_created` to `_reused` and then moved `_reused` to be behind the check of `_clean_location`. 

I simplified the tests a little as well, because I realised some of them were redundant.

When I read the code, it reads to me like the check for `--no-install` and `_reused` are doing the "right" things, but I made that mistake last time, so I may be mistaken 😞

I can see the comments in https://github.com/theacodes/nox/issues/394 about maybe renaming this from `--no-install` to `--reuse-install`. Should I close this PR while that conversation gets resolved? Or leave this open?

docs/config.rst
```diff
@@ -386,6 +386,7 @@ The following options can be specified in the Noxfile:
 * ``nox.options.default_venv_backend`` is equivalent to specifying :ref:`-db or --default-venv-backend <opt-default-venv-backend>`.
 * ``nox.options.force_venv_backend`` is equivalent to specifying :ref:`-fb or --force-venv-backend <opt-force-venv-backend>`.
 * ``nox.options.reuse_existing_virtualenvs`` is equivalent to specifying :ref:`--reuse-existing-virtualenvs <opt-reuse-existing-virtualenvs>`. You can force this off by specifying ``--no-reuse-existing-virtualenvs`` during invocation.
+* ``nox.options.no_install`` is equivalent to specifying :ref:`--no-install <opt-no-install>`.
```
cjolowicz, line 389:
```suggestion
```

This option cannot be specified in the Noxfile, you're not passing `noxfile=True` to the `Option` constructor. I'd suggest to omit this for now.

nox/sessions.py
```diff
@@ -368,6 +372,18 @@ def conda_install(
...
+            if venv._reused:
+                logger.info(
+                    "Venv exists: skipping conda installation, as --no-install is set."
+                )
+                return None
+            else:
+                if not self._runner.global_config.reuse_existing_virtualenvs:
+                    logger.info(
+                        "Venv not created yet: ignoring --no-install and installing from conda."
+                    )
```
cjolowicz, line 385: You asked about the coverage error (`382->388`). I think that happens because you have no test for when `--reuse-existing-virtualenvs --no-install` is passed and a new virtualenv was created. That's what the branch `382->388` is about, anyway.

I would suggest to remove this warning for now:

```suggestion
```

Actually, what I had in mind was not a warning in every session about ignoring the option, but a warning once at startup about using `--no-install` without `--reuse-existing-virtualenvs`. But I see now that we don't have a logger yet when the options are parsed, and there is no prior art for warning about command line options either.

In the interest of moving this forward, I think omitting the warning is a good idea. The warning does not seem important, and less code is good :wink:

nox/sessions.py
```diff
@@ -417,15 +433,29 @@ def install(self, *args: str, **kwargs: Any) -> None:
...
+            if venv._reused:
+                logger.info(
+                    "Venv exists: skipping installation, as --no-install is set."
+                )
+                return None
+            else:
+                if not self._runner.global_config.reuse_existing_virtualenvs:
+                    logger.info(
+                        "Venv not created yet: ignoring --no-install and installing."
+                    )
```
cjolowicz, line 457:
```suggestion
```

Same as above.

cjolowicz (commented):
Hey @jamandbees thanks for the update! Here are some suggestions. This should also take care of the coverage woes you mentioned.

nox/sessions.py
cjolowicz: Just a small clarification about Coverage output: `382->388` is *not* the region of lines 382 to 388. It is the jump from 382 to 388. That happens in the situation described above: `--no-install --reuse-existing-virtualenvs` plus no reuse i.e. new venv.

docs/config.rst
jamandbees: I was trying to be helpful 🙂

Thanks for the clarification! I didn't know that detail.

nox/sessions.py
jamandbees: Oh, that's super interesting, thanks for the extra detail! I'm not sure I would've figured out that the syntax was referring to a jump.

cjolowicz (commented):
Hey @jamandbees here's a fresh review, see the inline comments :+1: 

I think we're quite close to wrapping this up. The shorthand for `-r --no-install` and the documentation can also be left to a follow-up PR. But feel free to have a go at these if you want to. 

In case you do want to tackle the shorthand option: I liked the option name you proposed (the second one). Also, a short option `-R` would make sense to me (mirroring `-r` for plain reuse). You could probably implement this using a finalizer function, see `_color_finalizer` in `_options.py` for an example.

nox/sessions.py
```diff
@@ -417,15 +428,21 @@ def install(self, *args: str, **kwargs: Any) -> None:
...
+            venv, (CondaEnv, VirtualEnv, PassthroughEnv)
         ):  # pragma: no cover
             raise ValueError(
                 "A session without a virtualenv can not install dependencies."
             )
         if not args:
             raise ValueError("At least one argument required to install().")
 
+        if self._runner.global_config.no_install and venv._reused:
+            logger.info("Venv exists: skipping installation, as --no-install is set.")
```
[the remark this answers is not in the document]
jamandbees: I had added these warnings because I thought that an end-user may appreciate the heads-up that a non-default behaviour was occurring with the --no-install flags. However, I don't know that they're actually _necessary_, especially if the flags are well documented.

Let me remove them and see what we think 🙂

nox/virtualenv.py
```diff
@@ -399,6 +402,8 @@ def create(self) -> bool:
                     self.location_name
                 )
             )
+            self._reused = False
```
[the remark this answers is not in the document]
jamandbees: I've added a couple of assertions for _reused, one in the tests/test_virtualenv.py in `test_condaenv_create` and the other in `test_create`.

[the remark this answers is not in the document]
jamandbees: > I think we're quite close to wrapping this up. The shorthand for -r --no-install and the documentation can also be left to a follow-up PR. But feel free to have a go at these if you want to.

I think it makes sense to relegate those to a follow-up PR. I think I've had this one hanging out for long enough 🙂

Thanks for your help and patience so far!

cjolowicz:
@jamandbees Thanks for the update! I enabled your CI run, and it looks like some tests need adapting to your latest changes. Could you take a look?

jamandbees:
@cjolowicz I'm honestly not sure how to adapt them; when I check git, I get the impression that this is from a recent merge, but I'm not confident that I understand what's going on.

cjolowicz:
Hey @jamandbees , no problem and thanks for your work so far! I picked this up in #432 and credited you as co-author.

PR #413, opened 2021-03-31
Allow nox.parametrize to select the session Python

### LINKED ISSUE #392: Exclude combinations from parametrized matrix builds

cjolowicz:
Fixes #392
Supercedes #404

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -339,6 +339,44 @@ Produces these sessions when running ``nox --list``:
     * tests(mysql, new)
 
 
+Parametrizing the session Python
+--------------------------------
+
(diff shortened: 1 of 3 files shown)


theacodes (approved):
Lovely! Thank you. :)

Issue #396, opened 2021-03-05
Double-quotation marks aren't accepted for parametrised sessions


smithfred:
Given the following:

```
@nox.parametrize("foo", ["bar", "baz"])
def test(...
```

I'd expect to be able to run the test with a given value by calling either `nox -s "test(foo='bar')"` or `nox -s 'test(foo="bar")'`, since single and double-quotation marks are equivalent in Python generally.

However, the latter form (or similar variations) fails.

Being able to use either is useful in terms of managing shell variable expansion etc.

theacodes:
This should be pretty straightforward if we normalize the quotes for the `--session` arg.

t-eckert:
@theacodes, I will take this one, but I will need some help. I have a fork and a branch. I started digging into the code for the first time. I _think_ what I need to do is modify code in the `tasks` module, but I could use some guidance to get on the right track.

Thank you!

theacodes:
Yeah, I'm thinking that you'll need to normalize the user-provided session names either [here](https://github.com/theacodes/nox/blob/main/nox/tasks.py#L129) or [here](https://github.com/theacodes/nox/blob/main/nox/manifest.py#L110). I feel like the latter is probably the more logical location.

t-eckert:
Ok, I think I will squeeze it into the line above where the session is appended to the `queue`. For "normalization"... I could write a function that checks if the string begins with a single quote then just swap every single to double or vice versa. I worry I would miss an edge case or change something unintentionally (a contraction perhaps (though unlikely)). 

Is there a more elegant way to do this normalization that you can think of?

If I have to write a helper function, where do I put it in the codebase?

cjolowicz:
How would we handle the case where the string itself contains quotes?

TBH I think we're better off documenting that `--session` must be used with the session name as displayed by `nox --list`. `nox.parametrize` can take any Python value, so fundamentally Nox would have to do what Black does to normalize these.

t-eckert:
I'll leave that discussion up to the maintainers. But since you mentioned Black, I wonder if there is a solution where Black's functionality could be leveraged to normalize the string.

t-eckert:
I will hang in a holding pattern on this until I get some direction from the maintainers. Still willing to do the work.

cjolowicz:
I'll provide some additional info, hoping that it helps us decide what to do here.

Here are some examples of valid session names:

```
* test(arg='jane')
* test(arg="jane's")
* test(arg='"hello world"')
[...]

PR #417, opened 2021-04-10
Add normalized comparison to check sessions (#396)

dpalmasan:
First of all, thanks for this amazing tool, I started to use it recently (I come from the `tox` world)

This is a possible fix for (#396):

- Attempt to do the previous checks
- On fail, it will normalize arguments using ast (based on discussion)
- The comparison will be on ast representation of the strings
- Add unit tests

### DIFF
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import argparse
+import ast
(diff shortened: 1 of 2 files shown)


theacodes (approved):
I like this approach a lot! Thank you!

PR #418, opened 2021-04-10
Check whether interpreter changed and type (#123)

### LINKED ISSUE #123: reuse_existing should check the interpreter

dpalmasan:
This is an attempt to fix #123 

The steps are the following
- Check virtualenv interpreter
- Check current interpreter
- If there is a match, do nothing. If they are different, recreate
- Check if the type of interpreter remains the same.

Not sure, but for the last case I made an assertion.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -313,11 +313,18 @@ class VirtualEnv(ProcessEnv):
         """Deletes any existing virtual environment"""
         if os.path.exists(self.location):
             if self.reuse_existing:
+                self._check_reused_environment()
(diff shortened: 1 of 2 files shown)


theacodes (approved):

theacodes:
Thank you!

cjolowicz:
This caused a regression, see #424.

dpalmasan:
Thanks, gonna check

PR #420, opened 2021-04-15
fix position of venv_params within conda create

### LINKED ISSUE #419: creation of conda environment fails when venv_params is not None
**Describe the bug**
If venv_params is used when creating a conda environment, then the value of venv_params gets inserted into the conda command line argument list at the wrong location. The current code attempts to add pip to the list of packages to be installed and places 'pip' before venv_params. However, the `conda create ...` command expects all options to be listed before package names, so if venv_params contains an optional switch, then an invalid command is generated.  This causes the `conda create ...` command to fail and not create the desired environment.

**How to reproduce**
```
$ uname -a
Darwin myhostname 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64
$ python --version
Python 3.8.5
$ nox --version
2020.12.31
$ cat noxfile.py 
import nox

@nox.session(venv_backend='conda', venv_params=['--channel=conda-forge'])
def should_work(session):
    session.conda_install('flake8')
$ nox
nox > Running session should_work
nox > Creating conda env in .nox/should_work with python
nox > Command conda create --yes --prefix /Users/myuserid/myproject/.nox/should_work pip --channel=conda-forge python failed with exit code 2:
usage: conda [-h] [-V] command ...
conda: error: unrecognized arguments: python
nox > Session should_work failed.
$ 
```

**Expected behavior**
The command executed should be:
```
conda create --yes --prefix /Users/myuserid/myproject/.nox/should_work --channel=conda-forge pip python
```
And then the conda environment is successfully created.

wholtz:
Fixes #419 by moving 'pip' to after venv_params

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -226,17 +226,12 @@ class CondaEnv(ProcessEnv):
             )
             return False
 
-        cmd = [
-            "conda",
-            "create",
-            "--yes",
-            "--prefix",
-            self.location,
-            # Ensure the pip package is installed.
-            "pip",
-        ]
+        cmd = ["conda", "create", "--yes", "--prefix", self.location]
 
         cmd.extend(self.venv_params)
+
+        # Ensure the pip package is installed.
+        cmd.append("pip")
 
         if self.interpreter:
             python_dep = "python={}".format(self.interpreter)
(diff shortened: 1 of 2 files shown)


theacodes:
Nice, thank you!

jcollado:
@theacodes I need this PR to fix a problem I have with conda dependencies (I need to pass `venv_params=["-c", "conda-forge"]`. Since it's been more than four months since last release, would it be possible to create a new one that includes this fix?

PR #421, opened 2021-05-02
use standatd contextlib instead outdates contexter

kloczek:
contexter module is not maintained since 2017.

Tested using pytest
```console
$ /usr/bin/python3 -Bm pytest -ra tests/test_sessions.py
===================== test session starts ===============================
platform linux -- Python 3.8.9, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
rootdir: /home/tkloczko/rpmbuild/BUILD/nox-2020.12.31
plugins: forked-1.3.0, shutil-1.7.0, virtualenv-1.7.0, asyncio-0.14.0, expect-1.1.0, cov-2.11.1, mock-3.5.1, httpbin-1.0.0, xdist-2.2.1, flake8-1.0.7, timeout-1.4.2, betamax-0.8.1, pyfakefs-4.4.0, freezegun-0.4.2, flaky-3.7.0, cases-3.4.6, hypothesis-6.10.1, case-1.5.3, isort-1.3.0, aspectlib-1.5.2
collected 77 items

tests/test_sessions.py .............................................................................[100%]

====================== 77 passed in 1.53s ================================
```

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -332,7 +332,7 @@ incompatible versions of packages already installed with conda.
 
 .. code-block:: python
 
-    session.install("contexter", "--no-deps")
+    session.install("contextlib", "--no-deps")
     session.install("-e", ".", "--no-deps")
 
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -58,7 +58,7 @@ def conda_tests(session):
     session.conda_install(
         "--file", "requirements-conda-test.txt", "--channel", "conda-forge"
     )
-    session.install("contexter", "--no-deps")
+    session.install("contextlib", "--no-deps")
     session.install("-e", ".", "--no-deps")
     tests = session.posargs or ["tests/"]
     session.run("pytest", *tests)
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,7 +1,7 @@
 flask
 pytest
 pytest-cov
-contexter
+contextlib
 sphinx
 sphinx-autobuild
 recommonmark
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -17,7 +17,7 @@ import sys
 from pathlib import Path
 from unittest import mock
 
-import contexter
+import contextlib
 import nox
 import nox.__main__
 import nox._options
@@ -296,7 +296,7 @@ def test_main_positional_flag_like_with_double_hyphen(monkeypatch):
 def test_main_version(capsys, monkeypatch):
     monkeypatch.setattr(sys, "argv", [sys.executable, "--version"])
 
-    with contexter.ExitStack() as stack:
+    with contextlib.ExitStack() as stack:
         execute = stack.enter_context(mock.patch("nox.workflow.execute"))
         exit_mock = stack.enter_context(mock.patch("sys.exit"))
         nox.__main__.main()
@@ -309,7 +309,7 @@ def test_main_version(capsys, monkeypatch):
 def test_main_help(capsys, monkeypatch):
     monkeypatch.setattr(sys, "argv", [sys.executable, "--help"])
 
-    with contexter.ExitStack() as stack:
+    with contextlib.ExitStack() as stack:
         execute = stack.enter_context(mock.patch("nox.workflow.execute"))
         exit_mock = stack.enter_context(mock.patch("sys.exit"))
         nox.__main__.main()


docs/tutorial.rst
```diff
@@ -332,7 +332,7 @@ incompatible versions of packages already installed with conda.
 
 .. code-block:: python
 
-    session.install("contexter", "--no-deps")
+    session.install("contextlib", "--no-deps")
```
cjolowicz, line 335: I'd suggest to leave this unchanged. This is an example in the documentation for an arbitrary third-party library.

Also this does not work, contextlib cannot be installed from PyPI. Nor does it need to be, it's part of the standard library.

noxfile.py
```diff
@@ -58,7 +58,7 @@ def conda_tests(session):
     session.conda_install(
         "--file", "requirements-conda-test.txt", "--channel", "conda-forge"
     )
-    session.install("contexter", "--no-deps")
+    session.install("contextlib", "--no-deps")
```
cjolowicz, line 61: This line should be removed. There's no need to install contextlib, because it's part of the standard library.

```suggestion
```

requirements-test.txt
```diff
@@ -1,7 +1,7 @@
 flask
 pytest
 pytest-cov
-contexter
+contextlib
```
cjolowicz, line 4: This requirement should be removed, see above.

```suggestion
```

tests/test_main.py
```diff
@@ -17,7 +17,7 @@
 from pathlib import Path
 from unittest import mock
 
-import contexter
+import contextlib
```
cjolowicz, line 20: This import needs to be moved up to the start of the imports section, because it is a standard library import, not a third-party import.

This is what caused the lint check to fail in CI.

cjolowicz (changes requested):
Hi @kloczek thank you for reporting this, as well as providing a PR with a fix!

AFAICS we were only using contexter for its Python 2 support, not for any special features. So your approach of using contextlib instead should indeed work fine.

Please see my inline comments with suggestions how to fix the CI checks. As contextlib is not a PyPI package but a standard library, replacing all occurrences of contexter by contextlib does not work.

kloczek:
> AFAICS we were only using contexter for its Python 2 support, not for any special features. So your approach of using contextlib instead should indeed work fine.

Python 2 have been long time ago EOSed so IMO it is time to remove on the master all python 2.x suppoert :)

henryiii:
contextlib2 supports Python 2 and has a final stable Python 2 version (2019). I'd import from contextlib2 `if sys.version_info < (3,)`

kloczek:
As python 2.x is now EOSed keeping still support for that version of python today is IMO a bit odd.

henryiii:
That's up to the authors. Packages designed to run many versions of Python (virtualenv, tox, and of course nox) may want to keep Python 2 longer than normal packages. This could also be seen as a reason to use tox over nox.

Not saying Python 2 _should_ be kept here. And using an older nox is always a possibility for Python 2 users. But am saying it is the authors decision.

henryiii:
Also, and, personally, dropping something because of a _test_ is _ really irritating. Maybe it's just me. But I changed 5K LOC to go from GTest to Catch2 gust because GTest was going to force me to drop old compilers. Now for a feature, like full static typing, etc, or simpler code, sure.

kloczek:
> Also, and, personally, dropping something because of a test is _ really irritating. Maybe it's just me. But I changed 5K LOC to go from GTest to Catch2 gust because GTest was going to force me to drop old compilers. Now for a feature, like full static typing, etc, or simpler code, sure.

Don't get me wrong but it inot about test. As at the moment 2.x is EOSed do remove python 2.x code on the master is kind of against KISS principle.
All what is needed IMO is just document which one last version is for python 2.x and remove all python 2.x traces on master. Eventually if you still want to provide python 2.x support source code could be always branched at exact point list of changes and some fixes with updated release can be done.

Simple it is IMO about keep code on th emaster as clean as it is only possible.
Only this and nothing more😊

cjolowicz:
@kloczek Are you planning to address the points raised [above](https://github.com/theacodes/nox/pull/421#pullrequestreview-649948035)?

@henryiii Nox has dropped Python 2 support a while ago, so I don't see a reason to pull in a third-party dependency here. To be clear, we're talking about the Python version that Nox itself runs under.

henryiii:
I saw it in a nox file and assumed it was for testing a python 2 session. I think I was wrong, though.

kloczek:
Sorry have no time now :(
Feel free to close this PR and implement it how you like.

PR #423, opened 2021-05-19
Allow arbitrary kwargs to subprocess.Popen

rmorshea:
In my particular case I need to set the `cwd` param of the subprocess.


rmorshea:
nevermind there's a `session.chdir`

PR #426, opened 2021-05-23
Drop contexter from test requirements

cjolowicz:
The contexter package is a third-party backport of contextlib (with some
extensions). It is no longer required because Nox does not support Python 2
anymore, and we only ever used the standard ExitStack interface.

The contexter package was used only in the test suite.

Supercedes #421


theacodes (approved):

cjolowicz:
thanks @theacodes  :)

PR #427, opened 2021-05-24
Add --force-python option as shorthand for --python and --extra-python

cjolowicz:
This PR adds an option `--force-python` / `--force-pythons` which functions as a shorthand for when the same interpreter spec should be passed to both `--python` and `--extra-python`. This has the effect of running sessions on the specified interpreter *instead* of those listed for each session.

In other words, these two are equivalent:

- `nox --python=X.Y --extra-python=X.Y`
- `nox --force-python=X.Y`

By comparison, `--extra-python` on its own adds an interpreter to those listed for a session. The sessions will be run on the original Python versions *as well as* the extra version.

This came up a few times in the past:

- https://github.com/theacodes/nox/pull/361#issuecomment-748300867
- https://github.com/theacodes/nox/issues/412#issuecomment-810164777
- https://github.com/theacodes/nox/issues/233#issuecomment-846614741

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -163,22 +163,27 @@ If the Noxfile sets ``nox.options.reuse_existing_virtualenvs``, you can override
 
 Running additional Python versions
 ----------------------------------
-In addition to Nox supporting executing single sessions, it also supports runnings python versions that aren't specified using ``--extra-pythons``.
+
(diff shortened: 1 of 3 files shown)


theacodes (approved):
This looks great! Thank you. :)

PR #432, opened 2021-05-28
Add option --no-install to skip install commands in reused environments

### LINKED ISSUE #394: Add an option to skip install commands

cjolowicz:
Add option `--no-install` to skip the re-installation of packages when a virtualenv is reused.

Use either of these to reuse a virtualenv without re-installing packages:

```
nox -R
nox --reuse-existing-virtualenvs --no-install
```

The `--no-install` option causes the following session methods to return early:

- `session.install`
- `session.conda_install`
- `session.run_always`

This option has no effect if the virtualenv is not being reused.

Closes #394
Supercedes #411 

Co-authored-by: @jamandbees

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -157,6 +157,13 @@ def _force_pythons_finalizer(
     return value
 
 
+def _R_finalizer(value: bool, args: argparse.Namespace) -> bool:
+    """Propagate -R to --reuse-existing-virtualenvs and --no-install."""
+    if value:
+        args.reuse_existing_virtualenvs = args.no_install = value
+    return value
+
+
 def _posargs_finalizer(
     value: Sequence[Any], args: argparse.Namespace
 ) -> Union[Sequence[Any], List[Any]]:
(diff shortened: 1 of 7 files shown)


emilmelnikov:
LGTM, thanks!

nox/_options.py
```diff
@@ -320,6 +327,18 @@ def _session_completer(
         group=options.groups["secondary"],
         help="Re-use existing virtualenvs instead of recreating them.",
     ),
+    _option_set.Option(
+        "R",
```
theacodes, line 331: This should be lowercase, right?

cjolowicz: The option is uppercase (`-R`), so using `r` as the option name would lead to an error:

```python
  File ".../nox/venv/bin/nox", line 33, in <module>
    sys.exit(load_entry_point('nox', 'console_scripts', 'nox')())
  File ".../nox/nox/__main__.py", line 30, in main
    args = _options.options.parse_args()
  File ".../nox/nox/_option_set.py", line 267, in parse_args
    self._finalize_args(args)
  File ".../nox/nox/_option_set.py", line 255, in _finalize_args
    value = getattr(args, option.name)
AttributeError: 'Namespace' object has no attribute 'r'
```

Our `option.name` must match the name that argparse derives for the option. In this case, the option is uppercase `-R` with no equivalent long option, so argparse derives `R` as the option name.

theacodes: Gotcha

theacodes (approved):

PR #438, opened 2021-06-03
Upgrade linters to the latest version

cjolowicz:
Bump the dependencies of Nox's lint session:

- Bump black from 19.3b0 to 21.5b2
- Bump flake8 from 3.7.8 to 3.9.2
- Bump isort from 4.3.21 to 5.8.0
- Bump mypy from 0.720 to 0.812

Fix new warnings and type errors.

Apply changes in code style and import ordering.

Use the new profile feature of isort 5 to achieve compatibility with black.

Supercedes #435
Supercedes #436
Supercedes #437

PR #431, opened 2021-05-26
Allow colorlog <7.0.0

theacodes:
5 is the latest release, and 6.x just drops support for Python 2, so we should be good.

Fixes #430

PR #425, opened 2021-05-23
Fix regression in check for stale environments

cjolowicz:
Fixes #424

PR #403, opened 2021-03-11
Remove Travis references from docs

paw-lu:
#389 replaces Travis with GitHub actions. This PR removes Travis references from the docs.

---

Thanks!

PR #440, opened 2021-06-06
Release 2021.6.6

cjolowicz:
What are everybody's thoughts on cutting a release?

## 2021.6.6

- Add option `--no-install` to skip install commands in reused environments. (#432)
- Add option `--force-python` as shorthand for `--python` and `--extra-python`. (#427)
- Do not reuse environments if the interpreter or the environment type has changed. (#418, #425, #428)
- Allow common variations in session names with parameters, such as double quotes instead of single quotes. Session names are considered equal if they produce the same Python AST. (#417, #434)
- Preserve the order of parameters in session names. (#401)
- Allow `@nox.parametrize` to select the session Python. (#413)
- Allow passing `posargs` when scheduling another session via `session.notify`. (#397)
- Add `nox.needs_version` to specify Nox version requirements. (#388)
- Add `session.name` to get the session name. (#386)
- Gracefully shutdown child processes. (#393)
- Decode command output using the system locale if UTF-8 decoding fails. (#380)
- Fix creation of Conda environments when `venv_params` is used. (#420)
- Various improvements to Nox's type annotations. (#376, #377, #378)
- Remove outdated notes on Windows compatibility from the documentation. (#382)
- Increase Nox's test coverage on Windows. (#300)
- Avoid mypy searching for configuration files in other directories. (#402)
- Replace AppVeyor and Travis CI by GitHub Actions. (#389, #390, #403)
- Allow colorlog <7.0.0. (#431)
- Drop contexter from test requirements. (#426)
- Upgrade linters to the latest version. (#438)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,32 @@
 # Changelog
 
+## 2021.6.6
+
+- Add option `--no-install` to skip install commands in reused environments. (#432)
+- Add option `--force-python` as shorthand for `--python` and `--extra-python`. (#427)
+- Do not reuse environments if the interpreter or the environment type has changed. (#418, #425, #428)
+- Allow common variations in session names with parameters, such as double quotes instead of single quotes. Session names are considered equal if they produce the same Python AST. (#417, #434)
+- Preserve the order of parameters in session names. (#401)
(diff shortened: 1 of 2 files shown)


theacodes (approved):

cjolowicz:
Thanks!

PR #448, opened 2021-06-11
feat: add mamba environment

### LINKED ISSUE #444: conda: support for Mamba

It would be great to support `mamba`, which is a conda replacement solver that is dramatically faster. A solve of a complex package like ROOT (which is a reason to use conda in the first place) takes several minutes with conda, and a few seconds with mamba.

Option 1: Provide a new backend: `"mamba"`. This would simply make `conda_install(...)` use mamba instead of conda. This is the best option I could think of.

Option 2: Automatically use mamba if available over conda. This is slightly too magical, and there are _minute_ difference in the solve (that shouldn't matter).

Option 3: Add a setting to select the command used for conda. I don't think there is a good reason to generalize too much - mamba is a drop-in replacement for part of conda (other parts can just use conda, as it's always available with mamba).

This was mentioned in https://github.com/theacodes/nox/issues/346#issue-695644289 as a side note, but never attempted.

henryiii:
Closes #444.

Initial prototype, no docs or tests yet. Does this look like a good way to do it?

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -375,8 +375,9 @@ incompatible versions of packages already installed with conda.
     session.install("contexter", "--no-deps")
     session.install("-e", ".", "--no-deps")
 
-``"mamba"`` is also allowed as a choice, which will use/require 
-`mamba <https://github.com/mamba-org/mamba>`_ instead of conda.
+``"mamba"`` is also allowed as a choice for ``venv_backend``, which will
+use/require `mamba <https://github.com/mamba-org/mamba>`_ instead of conda.
+
 
 Parametrization
 ---------------
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -375,6 +375,8 @@ incompatible versions of packages already installed with conda.
     session.install("contexter", "--no-deps")
     session.install("-e", ".", "--no-deps")
 
+``"mamba"`` is also allowed as a choice, which will use/require 
+`mamba <https://github.com/mamba-org/mamba>`_ instead of conda.
 
 Parametrization
 ---------------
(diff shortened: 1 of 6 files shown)


henryiii:
Does this look like a good direction? Happy to polish it off if so. Also, shouldn't mamba fall back to conda if mamba is not available? I rather think so?

henryiii:
@theacodes any direction here on the way you'd like it to go?

theacodes:
Ah, I'm sorry this kept flying under my radar. This approach is great.

henryiii:
This implementation is just a simple "mamba" can be chosen implementation - it does not auto-switch to conda if mamba is not available, or install mamba for you. A user _could_ detect this when setting the backend in their `noxfile.py`, though, if that was important to them, and one of those two things could be added later if commonly used.

docs/tutorial.rst
```diff
@@ -375,6 +375,8 @@ incompatible versions of packages already installed with conda.
     session.install("contexter", "--no-deps")
     session.install("-e", ".", "--no-deps")
 
+``"mamba"`` is also allowed as a choice, which will use/require 
+`mamba <https://github.com/mamba-org/mamba>`_ instead of conda.
```
cjolowicz, line 379:
```suggestion
``"mamba"`` is also allowed as a choice for ``venv_backend``, which will
use/require `mamba <https://github.com/mamba-org/mamba>`_ instead of conda.

```

We should mention `venv_backend` as it's not clear from the immediate context. The code example with `venv_backend="conda"` is quite a bit further up.

cjolowicz (commented):
Thank you @henryiii

Minor nit for the docs, otherwise this looks good to me! Good call about `PassthroughEnv` also needing `conda_cmd`.

cjolowicz (approved):
Thanks 🚀

DiddiLeija (approved):

PR #445, opened 2021-06-10
Avoid polluting tests with a .nox directory

cjolowicz:
Fixes #443 
Follow-up to #434 

The test `test_main_with_normalized_session_names` from #434 creates a `.nox` directory under `tests/resources`. This breaks the linting session. Avoid this by passing `venv_backend="none"` to the session decorator.

PR #442, opened 2021-06-10
Group CLI arguments in functional groups

DrGFreeman:
Fixes #305 

Replace the primary/secondary argparse argument groups with functional groups:

```
$ nox -h
usage: nox [-h] [--version] [-l] [-s [SESSIONS [SESSIONS ...]]]
           [-p [PYTHONS [PYTHONS ...]]] [-k KEYWORDS] [-v] [-ts]
           [-db {none,virtualenv,conda,venv}]
           [-fb {none,virtualenv,conda,venv}] [--no-venv] [-r]
           [--no-reuse-existing-virtualenvs] [-R] [-f NOXFILE]
           [--envdir ENVDIR]
           [--extra-pythons [EXTRA_PYTHONS [EXTRA_PYTHONS ...]]]
           [--force-pythons [FORCE_PYTHONS [FORCE_PYTHONS ...]]] [-x]
           [--no-stop-on-first-error] [--error-on-missing-interpreters]
           [--no-error-on-missing-interpreters] [--error-on-external-run]
           [--no-error-on-external-run] [--install-only] [--no-install]
           [--report REPORT] [--non-interactive] [--nocolor] [--forcecolor]
           ...

Nox is a Python automation toolkit.

General arguments:
  These are general arguments used when invoking Nox.

  -h, --help            Show this help message and exit.
  --version             Show the Nox version and exit.
  posargs               Arguments following ``--`` that are passed through to
                        the session(s).
  -f NOXFILE, --noxfile NOXFILE
                        Location of the Python file containing nox sessions.

Sessions arguments & flags:
  These arguments are used to control which Nox session(s) to execute.

  -l, --list-sessions, --list
                        List all available sessions and exit.
  -s [SESSIONS [SESSIONS ...]], -e [SESSIONS [SESSIONS ...]], --sessions [SESSIONS [SESSIONS ...]], --session [SESSIONS [SESSIONS ...]]
                        Which sessions to run. By default, all sessions will
                        run.
  -k KEYWORDS, --keywords KEYWORDS
                        Only run sessions that match the given expression.

Python arguments & flags:
  These arguments are used to control which Python version(s) to use.

  -p [PYTHONS [PYTHONS ...]], --pythons [PYTHONS [PYTHONS ...]], --python [PYTHONS [PYTHONS ...]]
                        Only run sessions that use the given python
                        interpreter versions.
[...]

PR #452, opened 2021-06-12
Release 2021.6.12

cjolowicz:
- Fix crash on Python 2 when reusing environments. (#450)
- Hide staleness check behind a feature flag. (#451)
- Group command-line options in `--help` message by function. (#442)
- Avoid polluting tests with a .nox directory. (#445)


theacodes:
Thanks, @cjolowicz  :)

PR #455, opened 2021-06-15
chore: upgrade to mypy 0.902

henryiii:
Uses the new `pyproject.toml` MyPy configuration and much tighter checking, nearly `--strict`. I only left about two strict flags turned off.


nox/command.py
```diff
@@ -70,7 +75,7 @@ def run(
     paths: Optional[List[str]] = None,
     success_codes: Optional[Iterable[int]] = None,
     log: bool = True,
-    external: bool = False,
+    external: Union[Literal["error"], bool] = False,
```
henryiii, line 78: This was the confusing type that I noticed in the code that prompted this PR. The old type was incorrect, and should (and now is) have been caught by mypy.

noxfile.py
```diff
@@ -86,15 +86,16 @@ def blacken(session):
...
-        "--disallow-untyped-defs",
-        "--warn-unused-ignores",
-        "--ignore-missing-imports",
-        "nox",
+    session.install(
+        "flake8==3.9.2",
+        "black==21.6b0",
+        "isort==5.8.0",
+        "mypy==0.902",
+        "types-jinja2",
```
henryiii, line 94: In mypy 0.900+ the non-std-lib types have been pulled out, so these are now required for you will end up with much less type checking.

pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel",
+]
+build-backend = "setuptools.build_meta"
```
henryiii, line 6: This does activate the PEP 517 build system - this shouldn't affect anything, is the future of Python packaging, and may become the default after pip 21.3. It also works correctly in an environment without wheel installed, while the old build-system does not. But wanted to point out that this is a related change.

nox/command.py
```diff
@@ -21,11 +21,16 @@
 from nox.logger import logger
 from nox.popen import popen
 
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal
+
```
cjolowicz, line 28: Unfortunately, this will need `# pragma: no cover` on both branches, since we don't combine coverage data from the various Python sessions.

nox/command.py
cjolowicz, line 78: Thanks for catching this!

How about making this `Union[str, bool]` though? Being so specific here means tighter coupling, and forces us to pull in the dependency on `typing_extensions` for older Pythons.

To be clear, I'm very much in favor of using modern typing constructs, and I agree that we might pull in that dependency anyway in the future. I'm just not convinced we should be using `Literal` here specifically.

pyproject.toml
cjolowicz: I'm all for it, but IMHO this should happen in a separate PR. We might also consider migrating to `setup.cfg`.

setup.py
```diff
@@ -54,6 +54,7 @@
         "colorlog>=2.6.1,<7.0.0",
         "packaging>=20.9",
         "py>=1.4.0,<2.0.0",
+        "typing_extensions>=3.7.4; python_version < '3.8'",
```
cjolowicz, line 57: See above.

cjolowicz (commented):
Lovely, thank you!

I left a comment about the coverage failure in CI (we should just disable it for these lines). Also, I'd prefer to separate the switch to PEP-517 from this PR. And having some small doubts about the use of `Literal`.

Overall, this is great, thanks again!

nox/command.py
cjolowicz: On second thoughts, I'm fine with using `Literal` here.

It's nice to have mypy detect typos in string literals, and it's definitely an idiomatic use of `Literal` as the parameter can only ever be `False`, `True`, or `"error"`.

pyproject.toml
henryiii: Okay, that has to be a pre-req for this one, though. Simply adding this file changes the build process to PEP 517, and the default for this is `"build-backend = "setuptools.build_meta.__legacy__"` (and it doesn't force setuptools >= 42 - 40 and 41 have bad bugs in pyproject.toml support).

henryiii: Pulled it out into #456, allow addressed the setup.cfg point in #457.

henryiii: (Simply adding a blank pyproject.toml file causes PEP 517 to activate, as the build-system selection is dependent on the presence of files, so I can't use the nice mypy config without #456 ad a pre-req)

cjolowicz: TIL :) Somehow I was under the impression that was a bug fixed long ago, but it turns out to be intended behavior, https://github.com/pypa/pip/issues/9738#issuecomment-808052363 and https://github.com/pypa/pip/issues/8437#issuecomment-805313362.

I see you just submitted a PR for this, thanks a lot!

henryiii: I think I finally understand the reasoning based on the filesystems (due to long discussions about whether a setup.cfg-only package with no pyproject.toml is a valid package under PEP 517), though from a user standpoint I still think it would have been much better to have required pyproject.toml _and_ a `project.build-backend` entry to activate.

PS: It was a very fine detail, but technically it is not valid to have a setup.cfg without either a pyproject.toml or setup.py, since PEP 517 allows a builder to directly call setup.py instead of running the setuptools entrypoint (though most/all builders choose the latter), so a valid PEP 517 builder could be written that wouldn't handle a project.toml file without pyprojec.toml, so `build` and `pip` have static checks requiring one of the two files).

cjolowicz:
@theacodes @crwilcox 
I'll merge #455, #456, and #457 on the weekend if there are no concerns from anybody.

layday:
Would need to rebase this on top of #457 or vice versa to handle the added dependency.

PR #457, opened 2021-06-15
chore: use setup.cfg

henryiii:
Since this was mentioned in #455 and I happen to have a tool that does 90% of the conversion automatically, here's the setup.cfg conversion. Shouldn't change anything. I left a token setup.py, but it's not even needed anymore even for editable installs on the latest version of pip. You can't remove the file completely though until you have a pyproject.toml file. But once you do, you can remove it if you want.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,60 @@
+[metadata]
+name = nox
+version = 2021.6.12
+description = Flexible test automation.
+long_description = file: README.rst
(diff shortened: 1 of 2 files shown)


henryiii:
PS: I used a one-time run `pipx run setup-cfg-format setup.cfg` to have a standard setup.cfg style. I usually add that via pre-commit & pre-commit.ci - see https://scikit-hep.org/developer/style.

PR #458, opened 2021-06-20
Remove setup.py

cjolowicz:
This PR removes the setup.py shim left after migrating to setup.cfg. It also updates the `deploy` job in CI to use `python -m build` to build the sdist and wheel for PyPI.

Note that with pip >= 21.1 (2021-04-24), editable installs for projects with a setup.cfg no longer require a setup.py shim.

Follow-up to #457


cjolowicz:
We'll also need to update the `deploy` job in CI to use `python -m build`.

.github/workflows/ci.yml
```diff
@@ -59,10 +59,10 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.9
-      - name: Install setuptools and wheel
-        run: python -m pip install --upgrade --user setuptools wheel
+      - name: Install build
+        run: python -m pip install --user build
       - name: Build sdist and wheel
-        run: python setup.py sdist bdist_wheel
+        run: python -m build
```
henryiii, line 65: _All_ of these lines can be replaced with:

```
- name: Build sdist and wheel
  run: pipx run build
```

GHA supports pipx without setup-python, and build supports pipx too.

.github/workflows/ci.yml
```diff
@@ -59,10 +59,10 @@ jobs:
...
       - name: Publish distribution PyPI
         uses: pypa/gh-action-pypi-publish@master
```
henryiii, line 67: I'd recommend pinning this and then using dependabot to update it every so often (it doesn't change that often).

DiddiLeija:
Hello @cjolowicz! Seems like you are trying to remove the `setup.py` in favor of `setup.cfg`. Just remember that pip will not longer support `setup.cfg`-only projects, see pypa/pip#9945 for the related discussion. 

If you are going to remove `setup.py`, just keep the `pyproject.toml`.

henryiii:
This is already a pyproject.toml based build, so it's safe to remove the setup.py - Pip is not "removing" support for anything except for a mistake that was introduced in a patch release of 21.1. That discussion was that _only_ setup.cfg, with no pyproject.toml, should not be supported, because that's not technically valid for all valid PEP 517 backends, as a valid backend _could_ call setup.py if pyproject.toml is missing (though most do not, most use the valid alternate method, which is directly calling a setuptools hook).

The downsides to removing setup.py is that older versions of pip (<21.1 IIRC) can't do an editable mode install, and you need a version of pip, 10+ or so, to install from SDist. Wheels are unaffected, so for a pure Python project, most users will not be affected at all. I do expect you will get questions or complaints from package managers who have workflows that call setup.py commands. But really, if they want to keep doing things the same (poor) way, they can just `echo "import setuptools; setuptools.setup()" > setup.py` to "fix" it. And it's better not to depend on `setup.py`.

DiddiLeija:
Of course. I just wanted to say that you must not remove `pyproject.toml` since you delete `setup.py`.

DiddiLeija:
> But really, if they want to keep doing things the same (poor) way, they can just echo "import setuptools; setuptools.setup()" > setup.py to "fix" it.

Oh, thanks. I also wanted to ask that.

.github/workflows/ci.yml
cjolowicz: Interesting, thanks! I'd like to keep the scope of this PR restricted to removing setup.py, but feel free to open an issue.

.github/workflows/ci.yml
cjolowicz: Using Dependabot is a great idea, I think there are more places where we could benefit from it. As with your other suggestion, I'd prefer to handle this in a separate PR.

henryiii: Happy to make those in a followup if (since) you like the ideas.

cjolowicz:
@henryiii @DiddiLeija Thanks for your input!

@theacodes I'll merge this on the weekend if you don't shout "stop" :)

Issue #454, opened 2021-06-14
cli: add header or footer to nox -l


henryiii:
**How would this feature be useful?**
`nox -l` is great; one of the best features is the fact that it shows the help text from the session's docstrings. It makes it a great entrypoint (and if there was a way to make the "default" be to show the sessions, such as `nox.sessions = "-l"`, I'd probably tend to use that in most of my noxfiles). However, it would be very nice if there was a way to add general information at the top or bottom (don't care which) of that printout. For example, I often would like to add a little help snippet like this:

```
Run on all versions of Python with:
  docker run --rm -itv $PWD:/src -w /src quay.io/pypa/manylinux_2_24_x86_64:latest pipx run nox -s tests
```

since we've added support for running nox directly from pipx in the manylinux images, this is a handy note to add.

**Describe the solution you'd like**
This could either be something explicit like `nox.header='...'`, or it could take the module docstring for the noxfile.

cjolowicz:
I like this idea. The module docstring seems a good fit for this IMO, as we already use docstrings of session functions.

Would this be done unconditionally, or require some specific markup?

Looking at pip, pipx, urllib3, and Nox itself, only pip uses a module docstring, which is simply "Automation using nox". From this partial evidence, I'd say probably no need for specific markup.

FollowTheProcess:
I use nox for all my python projects and I rate this idea!

Out of curiosity I forked and made a (very) naive implementation:

<img width="834" alt="image" src="https://user-images.githubusercontent.com/62767721/124123062-ba318f80-da6e-11eb-9444-5235b5dfb55d.png">


Which when you add the docstring at the top of `noxfile.py`:

<img width="681" alt="image" src="https://user-images.githubusercontent.com/62767721/124122593-295ab400-da6e-11eb-8c11-ee80a2efd298.png">



Exhibits the desired behaviour:

<img width="495" alt="image" src="https://user-images.githubusercontent.com/62767721/124122503-10ea9980-da6e-11eb-8e53-aa0636aba068.png">


I just took the functionality of `load_nox_module` and applied it here (hence naive). I'm assuming though that loading the module twice (even if just to get the `__doc__`) is not a good idea but I'm not really familiar enough with the rest of the `nox` machinery to know how best to handle this.

Do either of you have any pointers?

I'd be more than happy to implement this if so, I use nox a lot and would love the opportunity to contribute back (albeit a small amount).

cjolowicz:
Hi @FollowTheProcess 

[...]

PR #459, opened 2021-07-01
Add header from noxfile.py docstring #454

FollowTheProcess:
This commit adds the functionality for using the `noxfile.py` module
docstring as a header description for the `nox -l` option. #454 

The module docstring is now an attribute in `Manifest` which is
now populated in `discover_manifest` which is in-turn passed
to `honor_list_request` which will print the docstring if it is
present, and do nothing if it is not present.

I've also added two tests which cover these conditions (existent and
non-existent module docstring) and added to an existing
parametrized test.


nox/tasks.py
```diff
@@ -166,7 +169,9 @@ def honor_list_request(
         return manifest
 
     # If the user just asked for a list of sessions, print that
-    # and be done.
+    # and any docstring specified in noxfile.py and be done.
+    if manifest.module_docstring:
+        print(manifest.module_docstring)
```
cjolowicz, line 174: We need to deal with whitespace here, which is different depending on the various docstring styles:

```suggestion
        print(manifest.module_docstring.strip(), end="\n\n")
```

This should handle the following styles:

```python
# we need to ensure a blank line after the docstring
"""Automation using Nox."""

# ...while avoiding a double blank line here:
"""Automation using Nox.
"""

# and this should not start with an initial blank line:
"""
Automation using Nox.
"""
```

cjolowicz (changes requested):
Thank you :tada:

Please see the inline comment about handling whitespace in different docstring styles.

nox/tasks.py
FollowTheProcess: Good spot! I only ever use the third docstring style so this didn't occur to me.

I've incorporated your suggested changes now.

FollowTheProcess:
By the way, sorry for the commit spam. I tried to be clever and merge in the upstream changes and got myself confused. Still fairly new to this!

cjolowicz (approved):
LGTM

Issue #462, opened 2021-07-05
Have friendlier message if no noxfile.py found.


FollowTheProcess:
**How would this feature be useful?**

If you run nox in a directory with no `noxfile.py` you get a scary looking exception:

<img width="847" alt="image" src="https://user-images.githubusercontent.com/62767721/124506519-b92c9500-ddc3-11eb-8761-80dbe2b13440.png">

This is clearly the intended behaviour as using nox without a noxfile is a bit of a non-starter to say the least, and most people could easily figure out what it's trying to tell you, but this traceback is somewhat distracting and slightly off-putting (especially to newer python developers).

A friendlier message could just provide a nicer experience for people new to this whole thing.

**Describe the solution you'd like**

If nox hits a `FileNotFoundError` whilst trying to load `noxfile.py` it could instead return a friendlier log message without the traceback along the lines of "no noxfile.py found in current directory, create one and try again." with maybe a link to the docs or something like that?

It could perhaps even generate a "hello world" style placeholder noxfile along the lines of:

```python
import nox


@nox.session
def hello(session: nox.Session):
    """
    Your first nox session!
    """
    session.log("Hello world!")

```

But that might be unnecessary I don't know, keen for maintainers input here!

This seems pretty simple to implement and I'd be more than happy to submit a PR if this is something you'd like?

Thanks!

cjolowicz:
Thanks for raising this, PR welcome! I'd suggest to keep it simple, along the lines of "noxfile.py not found".

PR #463, opened 2021-07-05
Add friendlier message if no noxfile.py

FollowTheProcess:
#462 

Add a friendlier message on the specific case that user is calling
nox from within a directory with no noxfile.

Existing test for this case modified and two additional tests
added to ensure lower level errors are still handled further down.

### DIFF
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -52,6 +52,7 @@ def load_nox_module(global_config: Namespace) -> Union[types.ModuleType, int]:
             # Be sure to expand variables
             os.path.expandvars(global_config.noxfile)
         )
+        noxfile_parent_dir = os.path.realpath(os.path.dirname(global_config.noxfile))
 
         # Check ``nox.needs_version`` by parsing the AST.
         check_nox_version(global_config.noxfile)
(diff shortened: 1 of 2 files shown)


nox/tasks.py
```diff
@@ -60,14 +61,17 @@ def load_nox_module(global_config: Namespace) -> Union[types.ModuleType, int]:
...
         return importlib.machinery.SourceFileLoader(
             "user_nox_module", global_config.noxfile
         ).load_module()
 
     except (VersionCheckFailed, InvalidVersionSpecifier) as error:
         logger.error(str(error))
         return 2
+    except FileNotFoundError:
+        logger.error(f"noxfile.py not found in {noxfile_parent_dir!r}.")
+        return 2
```
cjolowicz, line 74: We need to generalize this error message for the case where the Noxfile is not named `noxfile.py`.

I'd suggest to use the exception handler for `IOError` below this one as a template:

```suggestion
    except FileNotFoundError:
        logger.error("Failed to load Noxfile {}: no such file".format(global_config.noxfile))
        return 2
```

Replacing `logger.exception` with `logger.error` avoids the traceback.

cjolowicz (commented):
Thank you for the PR!

Please see the inline comment.

nox/tasks.py
FollowTheProcess: Good spot!, I'll tweak that now.

cjolowicz (approved):

PR #464, opened 2021-07-10
Run Flynt to convert str.format to f-strings

FollowTheProcess:
I didn't file an Issue for this one because this is very much down to personal developer preference, so feel free to reject this if you don't want it I won't be offended at all! :)

Things like this would be very tedious to do by hand but I discovered an automated tool [Flynt](https://github.com/ikamensh/flynt) to convert the older style `"{}".format()` strings with newer `f"{}"` style strings and it seems to have worked great!

Since 3.5 is now EOL, it might make sense to convert to the new format!

### DIFF
--- a/nox/_decorators.py
+++ b/nox/_decorators.py
@@ -68,7 +68,7 @@ class Func(FunctionDecorator):
 class Call(Func):
     def __init__(self, func: Func, param_spec: "Param") -> None:
         call_spec = param_spec.call_spec
-        session_signature = "({})".format(param_spec)
+        session_signature = f"({param_spec})"
 
         # Determine the Python interpreter for the session using either @session
         # or @parametrize. For backwards compatibility, we only use a "python"
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -163,7 +163,7 @@ def make_flag_pair(
     The positive option is considered to be available to the noxfile, as
     there isn't much point in doing flag pairs without it.
     """
-    disable_name = "no_{}".format(name)
+    disable_name = f"no_{name}"
 
     kwargs["action"] = "store_true"
     enable_option = Option(
@@ -174,9 +174,7 @@ def make_flag_pair(
         **kwargs
     )
 
-    kwargs["help"] = "Disables {} if it is enabled in the Noxfile.".format(
-        enable_flags[-1]
-    )
+    kwargs["help"] = f"Disables {enable_flags[-1]} if it is enabled in the Noxfile."
     disable_option = Option(disable_name, *disable_flags, **kwargs)
 
     return enable_option, disable_option
@@ -285,7 +283,7 @@ class OptionSet:
         # used in tests.
         for key, value in kwargs.items():
             if key not in args:
-                raise KeyError("{} is not an option.".format(key))
+                raise KeyError(f"{key} is not an option.")
             args[key] = value
 
         return argparse.Namespace(**args)
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -85,7 +85,7 @@ class Option:
         default: Union[Any, Callable[[], Any]] = None,
         hidden: bool = False,
         completer: Optional[Callable[..., List[str]]] = None,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> None:
         self.name = name
         self.flags = flags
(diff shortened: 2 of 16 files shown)


cjolowicz (commented):
+1 on converting the code base to f-strings, thanks!

Some comments:

- Sometimes expressions in the converted f-string contain a function call, or involve quotes. In these situations, it would be better for readability to extract the expression into a variable beforehand.
- I'm wondering why some calls of `str.format` in the `nox` directory were not converted. Do you have any idea?
- Can you include `noxfile.py` and the files in the `tests` directory in the conversion?
- As the diff shows, the `blacken` session insists on inserting the commas in unrelated files, even though it doesn't do so on main. Didn't investigate as it does not matter that much anyway.

FollowTheProcess:
> +1 on converting the code base to f-strings, thanks!
> 
> Some comments:
> 
> * Sometimes expressions in the converted f-string contain a function call, or involve quotes. In these situations, it would be better for readability to extract the expression into a variable beforehand.

I'm inclined to agree with you, I was going to do this but figured I'd see if the conversion was something you were interested in first. I can certainly fix this.
> * I'm wondering why some calls of `str.format` in the `nox` directory were not converted. Do you have any idea?

I'm not too sure, I ran `Flynt` against the entire directory. My best guess is if it doesn't know what to do it just leaves them alone? I'll look into this and maybe just fix the remainders manually.

> * Can you include `noxfile.py` and the files in the `tests` directory in the conversion?

Yep can definitely do this!

> * As the diff shows, the `blacken` session insists on inserting the commas in unrelated files, even though it doesn't do so on main. Didn't investigate as it does not matter that much anyway.

The only thing I can think of is I just ran black directly from the command line (although it was the same version as the `blacken` session so it shouldn't make a difference?) as the lint session defaults to 3.8 which I don't have on my system (I run 3.9)

FollowTheProcess:
@cjolowicz I've been able to manually change most of the remaining `.format` calls but I'm slightly confused by [this one](https://github.com/FollowTheProcess/nox/blob/8ee3ebed052ed7eeef0151a424929039dd479b94/nox/command.py#L121-L126)

I'm not too familiar with this style of formatting so I'm not entirely sure how to port this to f-strings (or if it's possible):

``` python
if return_code not in success_codes:
    logger.error(
        "Command {} failed with exit code {}{}".format(
            full_cmd, return_code, ":" if silent else ""
        )
    )
```

There a few examples that follow the same pattern and a few in the tests that I could not convert e.g inside a parametrised test where replacing with an empty f-string caused a error (cannot have empty placeholder in f-string):

``` python
@pytest.mark.parametrize("offline", [False, True], ids="offline={}".format)
```

Aside from those I think I've got all of them.

cjolowicz:
How about this?

```python
if return_code not in success_codes:
    suffix = ":" if silent else ""
    logger.error(f"Command {full_cmd} failed with exit code {return_code}{suffix}")
```

The pytest example should be left unchanged, because `format` is not called there (only partially bound).

FollowTheProcess:
> How about this?
> 
> ```python
> if return_code not in success_codes:
>     suffix = ":" if silent else ""
>     logger.error(f"Command {full_cmd} failed with exit code {return_code}{suffix}")
> ```
> 
> The pytest example should be left unchanged, because `format` is not called there (only partially bound).

Of course! The `":" if silent else ""` is just a variable like the others 🤦🏻 That's worked perfectly, I'll push up now thanks!

henryiii:
Is there no pyupgrade check? That should convert some (maybe not as many?) of these. Thanks, by the way, I heard about Flynt on Python Bytes, but didn't know how to spell it. :)

FollowTheProcess:
> Is there no pyupgrade check? That should convert some (maybe not as many?) of these. Thanks, by the way, I heard about Flynt on Python Bytes, but didn't know how to spell it. :)

Yeah looks like pyupgrade does this [too](https://github.com/asottile/pyupgrade#f-strings)!

FollowTheProcess:
I could run `pyupgrade --py36-plus` on this too if we're interested in modernising any remaining older syntax? Don't just want to lump it on this PR without checking first as it might clutter the narrative?

theacodes (approved):

theacodes:
I'm good with this PR as-is, @FollowTheProcess & @cjolowicz. Either of y'all can merge it if you'd like.

cjolowicz (approved):

FollowTheProcess:
Thanks guys! 🎉

PR #466, opened 2021-07-12
Conda logs now respect nox.options.verbose.

### LINKED ISSUE #345: Some conda commands not logged.
When using `venv_backend="conda"`, the `conda create` command is not logged, even if `-v` is provided, since `log=False` is hard-coded in [`nox/virtualenv.py`](https://github.com/theacodes/nox/blob/master/nox/virtualenv.py#L243):

```python
        nox.command.run(cmd, silent=True, log=False)
```

From the docs, I expected that running `nox -v` would log all commands.

FollowTheProcess:
Fixes #345 

First pass at this, I basically just copied the implementation from #371.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -22,6 +22,7 @@ from typing import Any, List, Mapping, Optional, Tuple, Union
 
 import py
 
+import nox
 import nox.command
 from nox.logger import logger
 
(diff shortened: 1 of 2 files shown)


theacodes (approved):

theacodes:
Looks good. :)

PR #472, opened 2021-09-03
469 add initial cwd to session properties

### LINKED ISSUE #469: Working directory unexpectedly set to wherever noxfile.py is.
**Describe the bug**
Trying to invoke nox from a different folder using the --noxfile flag sets the current working directory to where the noxfile lives.  This makes reusing a noxfile across multiple projects in a monorepo very difficult, since I can't find a good way to set the current working directory to where nox was called from.  Even passing '.' as a posarg will end up being interpreted as the directory where the noxfile lives

**Expected behavior**
Either the cwd stays as wherever nox is called from, or we have some command line arg for what to treat as the cwd.  If this feature already exists and I've missed it, or there are some other recommendations for how to use a single nox file across multiple projects, please let me know!

franekmagiera:
Closes #469

Added a working directory nox was originaly invoked from to session propreties.

@DiddiLeija I started working on this issue some time ago, but then had a long break and couldn't find time to finalize it until today and now noticed that you already opened a PR for it. Hopefully we did not duplicate much work, sorry for not mentioning it earlier and please let me know if you are ok with me continuing on this one.

### DIFF
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -284,7 +284,7 @@ class OptionSet:
         # don't bother with coverage here, this is effectively only ever
         # used in tests.
         for key, value in kwargs.items():
-            if key not in args and key != "original_wd":
+            if key not in args and key != "invoked_from":
                 raise KeyError("{} is not an option.".format(key))
             args[key] = value
 
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -77,7 +77,7 @@ class Option:
         self,
         name: str,
         *flags: str,
-        group: OptionGroup,
+        group: Optional[OptionGroup],
         help: Optional[str] = None,
         noxfile: bool = False,
         merge_func: Optional[Callable[[Namespace, Namespace], Any]] = None,
@@ -230,7 +230,7 @@ class OptionSet:
         }
 
         for option in self.options.values():
-            if option.hidden:
+            if option.hidden or option.group is None:
                 continue
 
             argument = groups[option.group.name].add_argument(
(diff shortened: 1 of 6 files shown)


DiddiLeija:
Oh, it's ok. I'll work on another issue. Go ahead!

theacodes:
This looks great, and thanks for doing it @franekmagiera. :)

I don't love the name `original_wd` though. :/ I couldn't really find a lot of prior art for this term, but I think either `session.executed_from` or `session.execution_dir` might work a bit better. What do you think?

henryiii:
To me, `session.execution_dir` sounds like the place it is executing, not the place it was called from. How about `session.invocation_dir`?  I like it ending in `dir` vs. `wd`.

franekmagiera:
@theacodes Thanks!

Now that you mention it I agree `original_wd` does not make much sense without the context of #469.

I like your propositions, having the name end with `dir` is better I think. How about just `session.starting_dir`? It's a bit simpler, but I'm fine with `session.invocation_dir` as well.

theacodes:
invocation_dir sounds good to me.

FollowTheProcess:
How about `invoked_from`? I know it doesn't specify dir explicitly but I think it's nice and clear, especially when in the context of `session.invoked_from`

theacodes:
👍

henryiii:
I'm okay with that too. It's a tossup for me, I like that the type is in the name with `invocation_dir`, but `invoked_from` probably can't be viewed as anything but a directory, and is a bit easier to read.

franekmagiera:
Ok, `session.invoked_from` it is then! Thanks for the discussion, I will make necessary changes tomorrow or on Saturday

franekmagiera:
@theacodes @henryiii @FollowTheProcess Changed the name to `invoked_from` and merged with main.

FollowTheProcess:
Looks good to me! happy to merge in unless anyone has any other suggestions/comments? @theacodes @cjolowicz

theacodes:
I have one plumbing change I want to make here. Let me do that and I'll merge. :)

theacodes:
Thanks, everyone!

PR #475, opened 2021-09-09
Add python 3.10.0-rc2 to GitHub Actions

### LINKED ISSUE #474: CI: Add python 3.10.0-rc2 to GitHub Actions

With the release of python 3.10 around the corner, it's probably a good idea to run the release candidate against nox in CI to uncover any weirdness.

Add 3.10.0-rc2 to `matrix.python_versions` in `ci.yml`

At the moment, conda does not support 3.10 so this would just be for standard python for now. You can see this running (and passing) in my fork here: [https://github.com/FollowTheProcess/nox/runs/3559269771?check_suite_focus=true](https://github.com/FollowTheProcess/nox/runs/3559269771?check_suite_focus=true)

I guess wait until 3.10 is released officially but seems like a no brainer to get ahead of the game!

FollowTheProcess:
Closes #474 

This PR adds `python 3.10.0-rc2` to `ci.yml` so it will be run as part of the standard CI suite.

I also noticed that the docs action specified python 3.8 but the docs session in `noxfile.py` specified 3.7 causing the docs session to just skip and therefore pass CI without actually building the docs, which I assume is not what we want so I've changed the docs session to python 3.8 in ` noxfile.py` to bring it in line with the other sessions that specify a python.

This is all passing locally and on my forks CI. Let me know if anyone wants anything else as part of this 👍🏻

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, windows-2019]
-        python-version: [3.6, 3.7, 3.8, "3.10.0-rc.2"]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


theacodes:
Looks great, thanks for doing this!

PR #479, opened 2021-09-16
CI: Python 3.10.0-rc.2 coverage fix

### LINKED ISSUE #478: Coverage failure on 3.10.0-rc.2
**Describe the bug**

See #477 for context. But short summary is that the python 3.10 CI runs I introduced in #475 actually weren't doing anything due to the difference between the version identifier `3.10.0-rc.2` and the actual interpreter path on the GHA runner. So nox just ignored it citing a missing interpreter (correct behavior) which made it appear like it was passing.

This was an easy fix to the Github Actions workflow but now I have a weird issue with coverage.

When running both locally and on my fork here: https://github.com/FollowTheProcess/nox/runs/3613749834?check_suite_focus=true the coverage is failing as it's less than 100%, which is super weird.

This obviously doesn't happen on any other version.

**How to reproduce**

Run `nox -s test-3.10 cover` locally in a 3.10 environment or on Github Actions as shown in my fork above.

**Expected behavior**

Coverage should be identical between python versions

Now it looks to me like this could be an upstream thing, there a few issues on [coverage](https://github.com/nedbat/coveragepy) that reference 3.10 results being different (e.g. https://github.com/nedbat/coveragepy/issues/1106 and https://github.com/nedbat/coveragepy/issues/1187)

I also tried running it using the most up to date master branch from coverage but the results were the same

My question is what we should do with the 3.10.0-rc.2 runs. I see 3 potential ways forward:

- Temporarily drop 3.10 runs from Github Actions until a later date (perhaps when 3.10 is officially released)
- Add a condition in the `cover` nox session that if python is 3.10 just return (similar to the `ON_WINDOWS_CI` check, this way we're still testing against 3.10 just not testing coverage)
- Add a similar condition that lowers the coverage requirement for 3.10 to 99% (the value I'm getting when running on 3.10, this way we're at least measuring coverage)

I'm happy with any approach, just wanted to get some input on how we want to handle this 🙂

FollowTheProcess:
Closes #478 

After some playing around with my fork I've managed to get tests and coverage playing nicely on 3.10.

Summary of changes:

- Detect when we're running on 3.10 and reduce coverage fail under to 99 from 100
- Parametrise `cover` with the same pythons as `tests` so I could access `session.python` to do the above
- Explicitly call `cover-<version`> in the Github Actions workflow (`session.notify` didn't seem to work for me here, I tried both `"cover"` and `"cover-<version>"` as arguments with no luck, if anyone knows how to fix this it would clean this solution up a little bit)

I did also notice that one of the tests (`test_session_completer`) was coupled to the noxfile. I didn't look into it too much but I think we should look at decoupling if possible as then we're free to change the noxfile however we want without altering test behaviour

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         run: |
           python -m pip install --disable-pip-version-check .
       - name: Run tests on ${{ matrix.os }}
-        run: nox --non-interactive --session "tests-${{ matrix.python-version }}" -- --full-trace
+        run: nox --non-interactive --session "tests-${{ matrix.python-version }}" "cover-${{ matrix.python-version }}" -- --full-trace
 
   build-py310:
     name: Build python 3.10
@@ -42,7 +42,7 @@ jobs:
         run: |
           python -m pip install --disable-pip-version-check .
       - name: Run tests on ${{ matrix.os }}
-        run: nox --non-interactive --session "tests-3.10" -- --full-trace
+        run: nox --non-interactive --session "tests-3.10" "cover-3.10" -- --full-trace
 
   lint:
     runs-on: ubuntu-20.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
-          python-version: "3.10"
+          python-version: "3.10.0-rc.2"
       # Conda does not support 3.10 yet, hence why it's skipped here
       # TODO: Merge the two build jobs when 3.10 is released for conda
       - name: Install Nox-under-test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         run: |
           python -m pip install --disable-pip-version-check .
       - name: Run tests on ${{ matrix.os }}
-        run: nox --non-interactive --session "tests-${{ matrix.python-version }}" "cover" -- --full-trace
+        run: nox --non-interactive --session "tests-${{ matrix.python-version }}" -- --full-trace
 
   build-py310:
     name: Build python 3.10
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -66,15 +65,19 @@ def conda_tests(session):
...
-@nox.session
+@nox.session(python=["3.6", "3.7", "3.8", "3.9", "3.10"])
 def cover(session):
     """Coverage analysis."""
     if ON_WINDOWS_CI:
         return
 
+    # 3.10 produces different coverage results for some reason
+    # see https://github.com/theacodes/nox/issues/478
+    fail_under = 99 if is_python_version(session, "3.10.0") else 100
```
cjolowicz, line 76: Instead of parametrising this session, could we just check `sys.version_info` here?

FollowTheProcess: Yeah I like this idea, worked locally for me

.github/workflows/ci.yml
```diff
@@ -22,28 +22,27 @@ jobs:
         run: |
           python -m pip install --disable-pip-version-check .
       - name: Run tests on ${{ matrix.os }}
-        run: nox --non-interactive --session "tests-${{ matrix.python-version }}" -- --full-trace
+        run: nox --non-interactive --session "tests-${{ matrix.python-version }}" "cover" -- --full-trace
```
cjolowicz, line 25: Should no longer be necessary:

```suggestion
        run: nox --non-interactive --session "tests-${{ matrix.python-version }}"  -- --full-trace
```

Same below in the `build-py310` job.

FollowTheProcess: I thought this too but when running `nox -s tests` locally it wouldn't notify `cover`. I'll see if it works in GHA.

FollowTheProcess: Ahh brilliant, seems to have worked! 👍🏻

cjolowicz:
> I did also notice that one of the tests (test_session_completer) was coupled to the noxfile. I didn't look into it too much but I think we should look at decoupling if possible as then we're free to change the noxfile however we want without altering test behaviour

Agreed :+1:

cjolowicz (approved):
Thanks :rocket:

FollowTheProcess:
> Thanks 🚀

Thanks for the help as always!

PR #484, opened 2021-09-25
Move configs into pyproject.toml or setup.cfg(flake8)

FollowTheProcess:
This PR moves a few configs over to either `pyproject.toml` or `setup.cfg` in the case of flake8. Summary below:

* isort config moved from `.isort.cfg` to `pyproject.toml` (also added recommended multi line output setting for black compatibility see [here](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html))
* coverage config moved from `.coveragerc` to `pyproject.toml` (the nox cover session now installs `coverage[toml]` over just `coverage` to achieve this)
* flake8 config moved from `.flake8` to `setup.cfg`. I'm hoping they support `pyproject.toml` soon!

This is mainly me being a bit OCD and it doesn't actually alter any functionality so please feel free to reject if we want to keep separate files for clarity etc. 🙂


cjolowicz (approved):
LGTM

Is the multiline setting actually necessary? I understood the isort config example as saying "use black profile for compatibility, and you can apply other options like multiline on top of that".

FollowTheProcess:
> LGTM
> 
> Is the multiline setting actually necessary? I understood the isort config example as saying "use black profile for compatibility, and you can apply other options like multiline on top of that".

I'm actually not too sure it's just a habit for me at this point. It didn't seem to change anything, perhaps in some edge cases it would behave differently.

Happy to take it out if we don't want that setting?

cjolowicz:
I was curious and checked. isort has this definition for the black profile:

```python
black = {
    "multi_line_output": 3,
    "include_trailing_comma": True,
    "force_grid_wrap": 0,
    "use_parentheses": True,
    "ensure_newline_before_comments": True,
    "line_length": 88,
}
```

So the multiline setting actually duplicates the profile settings.

FollowTheProcess:
> I was curious and checked. isort has this definition for the black profile:
> 
> ```python
> black = {
>     "multi_line_output": 3,
>     "include_trailing_comma": True,
>     "force_grid_wrap": 0,
>     "use_parentheses": True,
>     "ensure_newline_before_comments": True,
>     "line_length": 88,
> }
> ```
> 
> So the multiline setting actually duplicates the profile settings.

Well TIL! Thanks for that! Guess I'd best go and drop that from all my other projects then 👀

PR #490, opened 2021-10-02
Using `shlex.join()` when logging a command

dhermes:


nox/command.py
```diff
@@ -84,7 +85,7 @@ def run(
         success_codes = [0]
 
     cmd, args = args[0], args[1:]
-    full_cmd = f"{cmd} {' '.join(args)}"
+    full_cmd = f"{cmd} {shlex.join(args)}"
```
dhermes, line 88: Aw man; added in 3.8

dhermes: Luckily the implementation is dead simple:

```python
def join(split_command):
    """Return a shell-escaped string from *split_command*."""
    return ' '.join(quote(arg) for arg in split_command)
```

nox/command.py
```diff
@@ -67,6 +68,11 @@ def _clean_env(env: Optional[dict]) -> Optional[dict]:
     return clean_env
 
 
+def _shlex_join(args: Sequence[str]) -> str:
+    # shlex.join() was added in Python 3.8
+    return " ".join(shlex.quote(arg) for arg in split_command)
```
dhermes, line 73: Whoops, copy-pasta life

nox/command.py
```diff
@@ -84,7 +90,7 @@ def run(
         success_codes = [0]
 
     cmd, args = args[0], args[1:]
-    full_cmd = f"{cmd} {' '.join(args)}"
+    full_cmd = f"{cmd} {_shlex_join(args)}"
```
dhermes, line 93: It occurs to me I should probably do a docs search for `python -m pip` and see what should be updated.

(As reviewers may already see from the branch name, I made this PR fully from the browser.)

theacodes (approved):

theacodes:
Thanks @dhermes!

Issue #492, opened 2021-10-06
Invalid `--keywords` expression raises a SyntaxError


nawatts:
**How would this feature be useful?**

Currently, providing a `--keywords` argument that is not a valid expression causes Nox to crash with a SyntaxError.

```
$ nox -k 'foo:bar'
Traceback (most recent call last):
  ...
  File "<string>", line 1
    foo:bar
       ^
SyntaxError: invalid syntax
```

The value of the `--keywords` argument should be a Python expression. That expression is evaluated to filter the list of sessions.

https://github.com/theacodes/nox/blob/a2bd140ed9702428d3fc93be9857879388f21ad0/nox/manifest.py#L347-L350

However, I don't think that's entirely obvious. The name of the argument is `--keywords` and [the docs](https://nox.thea.codes/en/stable/usage.html#specifying-one-or-more-sessions) also refer to "pytest-style keywords".

The issue arises if the user attempts to do a keyword match on the [session name](https://nox.thea.codes/en/stable/config.html#session-name), which may contain characters that are not valid in Python identifiers. Using one of those characters in the `--keywords` expression raises a SyntaxError when it is evaluated.

Nox could provide some guidance here on how to resolve the issue. With the current behavior (printing a SyntaxError and stack trace), it's not clear to the user whether the issue is an invalid argument or a bug in Nox.

**Describe the solution you'd like**

It would be nice if the keywords argument could support matching on session names, including characters that aren't allowed in Python identifiers.

**Describe alternatives you've considered**

Validating the keywords expression and printing a descriptive error message (#493) would help users resolve the error.

FollowTheProcess:
Closing due to fix in #493 Thanks 🎉

FollowTheProcess:
Sorry didn't fully read! Issue refers to a slightly different feature than the fix in #493. My bad!

nawatts:
Sorry, I did kind of conflate two separate, though related, issues here.

PR #493, opened 2021-10-06
Show error message when keywords expression contains a syntax error

nawatts:
Related to #492.

Currently, providing a `--keywords` argument that is not a valid expression causes Nox to crash with a SyntaxError.

```
$ nox -k 'foo:bar'
  ...
  File "<string>", line 1
    foo:bar
       ^
SyntaxError: invalid syntax
```

This change validates the keywords argument using [ast.parse](https://docs.python.org/3/library/ast.html#ast.parse) and prints an error message if it is not a valid expression.

```
$ nox -k 'foo:bar'
nox > Error while collecting sessions: keywords argument must be a Python expression.
```

### DIFF
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ast
 import importlib.machinery
 import io
 import json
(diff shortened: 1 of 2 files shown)


cjolowicz (approved):
Thanks @nawatts 

Just waiting for the CI to pass after the merge.

PR #501, opened 2021-10-16
Add Python 3.10 for conda

maciej-lech:


FollowTheProcess:
@maciej-lech Looks good to me! Thanks 🎉

PR #513, opened 2021-12-09
Add a 'cwd' keyword argument to the Session.run API

### LINKED ISSUE #512: Run commands from within a specific directory

Since many commands have specific usages of the current working directory, 
running commands from within a specific directory would smooth integrating nox
is existing projects.

```python
from pathlib import Path

import nox


@nox.session()
def test(session: nox.Session):
    session.install('pytest')
    python_package_root = Path("somewhere/deep/in/monorepo")
    session.run(sys.executable, "-m", "pytest", cwd=python_package_root)
```

```python
import os
from pathlib import Path

import nox


@nox.session()
def test(session: nox.Session):
    session.install('pytest')
    python_package_root = Path("somewhere/deep/in/monorepo")
    before_cwd = os.getcwd()
    os.chdir(python_package_root)
    session.run(sys.executable, "-m", "pytest")
    os.chdir(before_cwd)
```

cblegare:
Closes #512

Adds a `cwd` keyword argument for running commands. Uses directly the underlying `subprocess.Popen` API.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -175,7 +175,7 @@ If you want to pass more arguments to a program just add more arguments to ``run
         session.run("pytest", "-v", "tests")
 
 
-You can also pass environment variables:
+You can also pass environment variables or select a working directory:
 
 .. code-block:: python
 
@@ -187,6 +187,7 @@ You can also pass environment variables:
             env={
                 "FLASK_DEBUG": "1"
             }
+            cwd="some/directory"
         )
 
 See :func:`nox.sessions.Session.run` for more options and examples for running
--- a/nox/popen.py
+++ b/nox/popen.py
@@ -14,6 +14,7 @@
 
 import contextlib
 import locale
+import os
 import subprocess
 import sys
 from typing import IO, Mapping, Optional, Sequence, Tuple, Union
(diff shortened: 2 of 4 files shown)


theacodes:
I think I'd rather use make session.chdir a context manager.

DiddiLeija:
Yes, I would also prefer a context manager.

I just edited the first post with a "close" command to link with #512.

cblegare:
Thanks for your feedbacks.  While I disagree about what would feel and look nicer, since the `run` api already mimics some arguments to the already well known `subprocess.Popen`, that is your decision.

Also, that kind of context manager would not use anything from the session object, so I'd rather use a context manager that is independent from `nox`.  In other words, since there is no pros in being attached to the Session instance for this particular use-case, I'd rather not depend on it.

Farewell!

cblegare:
> I think I'd rather use make session.chdir a context manager.

I now better understand that sentence.  I wasn't aware of the `session.chdir` implementation.  I'd advise against making `chdir` a context manager since it mimics and share the same name as the standard `os.chdir` and having a different behavior could be suprising for users, and surprising is often not what you want  in an API.

I can't spend much more time on this particular issue at the moment, but again, thanks for the feedback.

cjolowicz:
We may want to reconsider this if parallel session execution ever becomes a thing.

cjolowicz:
See also #555

cblegare:
Well, in case you @theacodes change your mind about this PR, I'll gladly fix checks, rebase and re-submit it.

cblegare:
Well, in case you @theacodes change your mind about this PR, I'll gladly fix checks, rebase and re-submit it.

PR #521, opened 2021-12-22
Skip `conda_tests` session by default if user does not have conda

### LINKED ISSUE #520: Conda tests session should skip rather than fail if conda not installed
**Describe the bug**

While quickly doing #519, I noticed that running bare `nox` on my machine results in somewhat unattractive looking failures:

image

I think a better solution would be to first check if the user has `conda` and skip this session if not (much like the tests themselves have a `pytest.mark.skipif` condition.

**How to reproduce**

Run `nox` on this project without having conda installed.

**Expected behavior**

Should skip `conda_tests` if conda is not installed.

This is a super quick fix and I'll do it now, filing the issue for traceability 🙂

FollowTheProcess:
Fixes #520

Adds a simple session filter at the top of our `noxfile.py` that runs all sessions (other than `conda_tests`) by default, only adding `conda_tests` if the user has conda installed.

```python
# Skip 'conda_tests' if user doesn't have conda installed
if shutil.which("conda"):
```

I also spotted a stray conda-dependent test without a `pytest.mark.skipif(not HAS_CONDA)` guard on it so I quickly added that too:

```python
@pytest.mark.skipif(not HAS_CONDA, reason="Missing conda command.")
def test_condaenv_(make_conda):
    venv, dir_ = make_conda()
    assert not venv.is_offline()
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -23,6 +23,11 @@ import nox
 
 ON_WINDOWS_CI = "CI" in os.environ and platform.system() == "Windows"
 
+# Skip 'conda_tests' if user doesn't have conda installed
+nox.options.sessions = ["tests", "cover", "blacken", "lint", "docs"]
+if shutil.which("conda"):
+    nox.options.sessions.append("conda_tests")
+
 
 def is_python_version(session, version):
     if not version.startswith(session.python):
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -233,6 +233,7 @@ def test_condaenv_bin_windows(make_conda):
     assert [dir_.strpath, dir_.join("Scripts").strpath] == venv.bin_paths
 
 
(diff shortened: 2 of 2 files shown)


DiddiLeija (approved):
This works! Thank you @FollowTheProcess!

cjolowicz (commented):
Thanks for taking care of this!

How about using `session.skip`?

```python
def conda_tests(session):
    if not shutil.which("conda"):
        session.skip("conda is not installed")
    ...
```

Nox's output will then make it clear that the session was skipped and why.

FollowTheProcess:
@cjolowicz Didn't know about `session.skip`, that should be perfect. Weirdly though I still get the error!

image

I'm guessing that the 

```python
nox.session(..., venv_backend=conda)
```
is what triggers it? In which case it probably never gets to the `session.skip` line?

cjolowicz (approved):
Oh right, that makes sense.

PR #522, opened 2021-12-22
feat: add channel keyword to conda

### LINKED ISSUE #446: conda: set channel via keyword
Minor addition for a common situation.

What about having a `channel=` keyword argument for `conda_install` so that one could write `channel="conda-forge"` if selecting a channel? It would be a bit more natural than `"--channel=conda-forge"`, I think, and likely is very common.

It is possible to set multiple channels - this is required for bioconda, for example. This could be an iterable, `channel=["conda-forge", "bioconda"]`. Maybe the keyword should be "channels" and always takes a list?

Leave as is.

henryiii:
Closes #446.

This adds a channel keyword, which can take a string or a list/tuple of channels.

Other options included `channels=`, and only allowing for a sequence of channels, but that probably is optimizing for the less-common case. Simpler API though. Can change if desired.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -360,7 +360,11 @@ class Session:
         return nox.command.run(args, env=env, paths=self.bin_paths, **kwargs)
 
     def conda_install(
-        self, *args: str, auto_offline: bool = True, **kwargs: Any
+        self,
+        *args: str,
+        auto_offline: bool = True,
+        channel: Union[str, Sequence] = "",
+        **kwargs: Any,
     ) -> None:
         """Install invokes `conda install`_ to install packages inside of the
         session's environment.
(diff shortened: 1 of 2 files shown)


nox/sessions.py
```diff
@@ -360,7 +360,11 @@ def _run(
         return nox.command.run(args, env=env, paths=self.bin_paths, **kwargs)
 
     def conda_install(
-        self, *args: str, auto_offline: bool = True, **kwargs: Any
+        self,
+        *args: str,
+        auto_offline: bool = True,
+        channel: Union[str, Sequence] = "",
```
cjolowicz, line 366:
```suggestion
        channel: Union[str, Sequence[str]] = "",
```

Bare `Sequence` is equivalent to `Sequence[Any]`. Should we be more specific here?

cjolowicz (commented):
LGTM with a small nit regarding typing.

nox/sessions.py
henryiii: Yes, Sequence[str] is better.

cjolowicz (approved):

PR #523, opened 2021-12-22
feat: support setting an empty session list

### LINKED ISSUE #465: sessions: empty list of default sessions could print help

Sometimes I'd like for the "default" run of nox (with no sessions) to print a help dialog instead; say if there's no clear "default" like `["lint", "tests"]`.

I'd differentiate between `nox.options.sessions = None` and an empty iterable, `nox.options.sessions = []`. The former would be the default, and would run all sessions. The latter would instead print the `nox -l` help if no sessions were passed - it would mean "no sessions". Currently, `nox.options.sessions = []` does exactly the same thing as `nox.options.sessions = None` and runs all sessions, which is clearly not what would be expected from setting an empty list here.

A special token or setting could be used, but since `nox.options.sessions = []` misbehaves currently, I think that would be a natural place to put it.

henryiii:
Closes #465.

Setting an empty session list in the noxfile will default to listing the available sessions.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -258,6 +258,20 @@ And if you run ``nox --sessions lint`` Nox will just run the lint session:
     nox > ...
     nox > Session lint was successful.
 
+
+In the noxfile, you can specify a default set of sessions to run if so a plain
+``nox`` call will only trigger certain sessions:
+
+.. code-block:: python
+
+    import nox
+
+    nox.options.sessions = ["lint", "test"]
+
+If you set this to an empty list, Nox will not run any sessions by default, and
+will print a helpful message with the ``--list`` output when a user does not
+specify a session to run.
+
 There are many more ways to select and run sessions! You can read more about
 invoking Nox in :doc:`usage`.
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -259,7 +259,7 @@ And if you run ``nox --sessions lint`` Nox will just run the lint session:
     nox > Session lint was successful.
 
 
-In the noxfile, you can specify a default set of sessions to run if so a plain
+In the noxfile, you can specify a default set of sessions to run. If so, a plain
 ``nox`` call will only trigger certain sessions:
 
 .. code-block:: python
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -187,7 +187,7 @@ def filter_manifest(
     if global_config.pythons:
         manifest.filter_by_python_interpreter(global_config.pythons)
         if not manifest:
-            logger.error("Python version selection caused no sessions to selected.")
+            logger.error("Python version selection caused no sessions to be selected.")
             return 3
 
     # Filter by keywords.
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -177,6 +177,10 @@ def filter_manifest(
             logger.error("Error while collecting sessions.")
             logger.error(exc.args[0])
             return 3
+    elif global_config.sessions is not None:
+        # If the user did not specify sessions, and the noxfile is set
+        # to an empty list, then list the sessions instead.
+        global_config.list_sessions = True
 
     # Filter by python interpreter versions.
     # This function never errors, but may cause an empty list of sessions
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -174,14 +174,19 @@ def filter_manifest(
 
     # Filter by the name of any explicit sessions.
     # This can raise KeyError if a specified session does not exist;
-    # log this if it happens.
-    if global_config.sessions:
+    # log this if it happens. The sessions does not come from the noxfile
+    # if keywords is not empty.
+    if global_config.sessions is not None:
         try:
             manifest.filter_by_name(global_config.sessions)
         except KeyError as exc:
             logger.error("Error while collecting sessions.")
             logger.error(exc.args[0])
             return 3
+    if not manifest and not global_config.list_sessions:
+        print("No sessions selected. Please select a session with -s <session name>.\n")
+        _produce_listing(manifest, global_config)
+        return 0
 
     # Filter by python interpreter versions.
     if global_config.pythons:
@@ -207,20 +212,6 @@ def filter_manifest(
     if not manifest and not global_config.list_sessions:
         logger.error("No sessions selected after filtering by keyword.")
         return 3
-
-    # If the global_config is set to an empty list, and a list was not
-    # requested, and no filtering was done,
-    if (
-        global_config.sessions is not None
-        and not global_config.sessions
-        and not global_config.list_sessions
-        and not global_config.keywords
-        and not global_config.pythons
-    ):
-        manifest.filter_by_name(global_config.sessions)
-        print("No sessions selected. Please select a session with -s <session name>.\n")
-        _produce_listing(manifest, global_config)
-        return 0
 
     # Return the modified manifest.
     return manifest
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -167,6 +167,11 @@ def filter_manifest(
             the manifest otherwise (to be sent to the next task).
 
     """
+    # Shouldn't happen unless the noxfile is empty
+    if not manifest:
+        logger.error(f"No sessions found in {global_config.noxfile}.")
+        return 3
+
     # Filter by the name of any explicit sessions.
     # This can raise KeyError if a specified session does not exist;
     # log this if it happens.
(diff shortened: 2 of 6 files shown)


FollowTheProcess:
@henryiii LGTM, just thinking maybe there should be a quick mention in the tutorial section of the docs? Probably fits best under the `Selecting which sessions to run` section. Otherwise looks great, thanks!

cjolowicz:
See https://github.com/theacodes/nox/issues/465#issuecomment-1000282204

docs/tutorial.rst
```diff
@@ -258,6 +258,20 @@ And if you run ``nox --sessions lint`` Nox will just run the lint session:
     nox > ...
     nox > Session lint was successful.
 
+
+A noxfile author can specify a default set of sessions to run if you do not
+want a plain ``nox`` call to trigger all sessions:
```
henryiii, line 263:
```suggestion
A noxfile author can specify a default set of sessions to run if so a plain
``nox`` call will only trigger certain sessions:
```

nox/tasks.py
```diff
@@ -169,6 +169,11 @@ def filter_manifest(
             the manifest otherwise (to be sent to the next task).
 
     """
+    # Shouldn't happen unless the nox file is empty
```
henryiii, line 172:
```suggestion
    # Shouldn't happen unless the noxfile is empty
```

noxfile.py
```diff
@@ -27,6 +27,7 @@
 nox.options.sessions = ["tests", "cover", "blacken", "lint", "docs"]
 if shutil.which("conda"):
     nox.options.sessions.append("conda_tests")
+nox.options.sessions = []
```
cjolowicz, line 30: Leftover from debugging?

```suggestion
```

cjolowicz (commented):
Thanks for tackling this, this is looking great!

Here are a few minor things I noticed. I haven't had time to review the changed control flow yet.

tests/test_tasks.py
```diff
@@ -239,6 +239,19 @@ def test_filter_manifest_pythons():
...
+    config = _options.options.namespace(
+        sessions=(), pythons=("1.2",), keywords=(), posargs=[]
+    )
+    manifest = Manifest(
+        {"foo": session_func_with_python, "bar": session_func, "baz": session_func},
+        config,
+    )
+    return_value = tasks.filter_manifest(manifest, config)
+    assert return_value == 3
+    assert "Python version selection caused no sessions to selected." in caplog.text
```
FollowTheProcess, line 252:
```suggestion
    assert "Python version selection caused no sessions to be selected." in caplog.text
```

Looks like this is just a typo causing a failure :)

FollowTheProcess (approved):

DiddiLeija (approved):

FollowTheProcess:
Looking good overall! My only (minor) nit is that if you set `nox.options.sessions = []` and run bare `nox` you get the following:

image

However if you run `nox -l` you get the little selected message on the end:

image

I have two thoughts:

1) Ideally these two should be consistent with one another
2) I think all sessions shown when `nox.options.sessions = []` should appear as "unselected"

However this could just be my OCD showing so I'm happy to be overruled if we're happy with this as is 🙂

nox/tasks.py
```diff
@@ -180,10 +185,11 @@ def filter_manifest(
             return 3
 
     # Filter by python interpreter versions.
-    # This function never errors, but may cause an empty list of sessions
-    # (which is an error condition later).
     if global_config.pythons:
         manifest.filter_by_python_interpreter(global_config.pythons)
+        if not manifest:
```
cjolowicz, line 190: This is a subtle change in behavior. Combining `--list` with `--python` or `--keywords` should not lead to an error, even if no sessions are selected. In other words, `--list` should always produce a listing.

```suggestion
        if not manifest and not global_config.list_sessions:
```

nox/tasks.py
```diff
@@ -199,28 +205,32 @@ def filter_manifest(
         # (which is an error condition later).
         manifest.filter_by_keywords(global_config.keywords)
 
-    # Return the modified manifest.
-    return manifest
+    if not manifest:
```
cjolowicz, line 208: Same as above.

```suggestion
    if not manifest and not global_config.list_sessions:
```

nox/tasks.py
```diff
@@ -199,28 +205,32 @@ def filter_manifest(
...
+        logger.error("No sessions selected after filtering by keyword.")
+        return 3
 
+    if (
+        not global_config.keywords
+        and not global_config.pythons
+        and not global_config.sessions
+        and not global_config.list_sessions
+        and global_config.sessions is not None
+    ):
```
cjolowicz, line 218: I feel the intent could be expressed a little more clearly here. How do you feel about the suggestion below?

I omitted the checks for keywords and pythons because these conditions would have led to bailing out earlier.

```suggestion
    # List sessions if `nox.options.sessions` is an empty sequence.
    if (
        global_config.sessions is not None
        and not global_config.sessions
        and not global_config.list_sessions
    ):
```

cjolowicz:
I agree with @FollowTheProcess that we don't need to special-case the listing for the "no default sessions" case. To be precise, the sessions can be displayed with a minus because they are all deselected, and the explanatory sentence at the end can be kept. (Codewise, this would amount to dropping the ˋselectˋ parameter in ˋproduce_listingˋ.)

cjolowicz:
Also I'm confused, why are there asterisks in the screenshot if all sessions are deselected?

nox/tasks.py
henryiii: I redid this, but I kept the keyword/python filtering, since `nox -k keyword` should select and run the tests with `keyword`, rather than forcing it to _also_ be passed via `-s`, I think.

cjolowicz:
I realize now that you don't filter sessions by name when `nox.options.sessions` is an empty list. This also explains why `nox --list` displays all sessions as selected (with an asterisk) even though they're not.

Can we change the condition for this filter to check for `None` instead of falsiness?

You will need to check for a non-empty manifest after filtering by session name, like for the other filters. If the check fails, you can invoke `produce_listing`. This would replace the block at the end of the function.

I believe there are three good reasons for this change:

1. It would make the control flow easier to understand.
2. It's important that `nox --list` can be used to debug session selection.
3. Filtering by Python should not cause sessions to become selected. Currently, your PR changes this behavior (when the session list is empty).

As for selecting sessions by keyword even when `nox.options.sessions` is an empty list: This already works. The `nox.options.sessions` setting is ignored when `-k` is passed, see the function `_sessions_and_keywords_merge_func` in the `_options` module.

What do you think?

nox/tasks.py
cjolowicz: See https://github.com/theacodes/nox/pull/523#issuecomment-1001499599

nox/tasks.py
```diff
@@ -199,28 +205,33 @@ def filter_manifest(
...
+    # requested, and no filtering was done,
+    if (
+        global_config.sessions is not None
+        and not global_config.sessions
+        and not global_config.list_sessions
+        and not global_config.keywords
+        and not global_config.pythons
+    ):
+        manifest.filter_by_name(global_config.sessions)
+        print("No sessions selected. Please select a session with -s <session name>.\n")
```
cjolowicz, line 222: Should we use the logger here? It would make the message stand out from the session listing.

```suggestion
        logger.warning(
            "No sessions selected. Please select a session with -s <session name>.\n"
        )
```

We'd need to adapt the new tests to use `caplog.text` instead of `capsys.readouterr().out`.

henryiii: Is it an error / warning though? The session listing is a print.

henryiii:
> As for selecting sessions by keyword even when nox.options.sessions is an empty list: This already works. 

Ah, this was the piece I didn't know.  I wanted to put the check here, but didn't realize it would not cause an issue with filtering. It's at all not obvious that this will be ignored from the logic in this function! Seems like it should have it in the if statement.

henryiii:
(Haven't tested the recent comment by hand with the noxfile yet)

noxfile.py
```diff
@@ -28,6 +28,8 @@
 if shutil.which("conda"):
     nox.options.sessions.append("conda_tests")
 
+nox.options.sessions = []
```
cjolowicz, line 31: Somehow this debugging statement is now back?

```suggestion
```

I assume this wasn't your intention as `nox.options.sessions` is assigned prior to this line.

cjolowicz (approved):
Thanks so much @henryiii for all the work that went into this

noxfile.py
henryiii: Oops, thanks!

PR #524, opened 2021-12-22
feat: support FORCE_COLOR

### LINKED ISSUE #502: Provide a FORCE_COLOR variable?

On GitHub Actions, getting nox to print out color is a bit painful. You have to remember to add a `--forcecolor` every invocation (I happen to have one that does a ton of nox calls).

A new variable, either `FORCE_COLOR`, or `NOX_FORCE_COLOR`, something like that, could be added to imply `--forcecolor`. If both `FORCE_COLOR` and `NO_COLOR` were set, `NO_COLOR` would take precedence. From a quick search, tox does not have this, and pytest uses `--color=yes/no/auto`, ~~also without an environment variable~~ (`FORCE_COLOR`). MyPy secretly provides `MYPY_FORCE_COLOR`. So feel free to say no. :)

I would be very happy if nox simply used color output on GitHub Actions, not requiring any workaround. Though I think having a way to force color is still nice, though not as critical as `NO_COLOR`.

I also tried `TERM=xterm-color`, which was mentioned in https://github.com/python/mypy/issues/7771, but that did not produce color output.

My suggestions to PRs ratio is not great right now, so I'll try to fix that in the near future! 😳

henryiii:
Closes #502.

This is going with `FORCE_COLOR`, following PyPA/build, pytest, & tox. Could be changed.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -329,7 +329,8 @@ Controlling color output
 By default, Nox will output colorful logs if you're using in an interactive
 terminal. However, if you are redirecting ``stderr`` to a file or otherwise
 not using an interactive terminal, or the environment variable ``NO_COLOR`` is
-set, nox will output in plaintext.
+set, nox will output in plaintext. If this is not set, and ``FORCE_COLOR`` is
+present, color will be forced.
 
 You can manually control Nox's output using the ``--nocolor`` and ``--forcecolor`` flags.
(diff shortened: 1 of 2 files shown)


cjolowicz (approved):
LGTM

DiddiLeija (approved):
:+1:

PR #528, opened 2021-12-23
chore: bump linting versions to latest

henryiii:
I've bumped the linting versions to latests, cleaned up the mypy updates, and bumped from Python 3.8 to Python 3.9.

Are you interested in moving to pre-commit + pre-commit.ci, perhaps? That would automate this process of bumping pins and would also help new contributors by automatically running the auto-fixes ("blacken") on PRs if you want it to. I'd apply the method described in https://scikit-hep.org/developer/style and used in pybind11, cibuildwheel, etc.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,7 +4,7 @@ build:
   image: latest
 
 python:
-  version: 3.8
+  version: 3.9
   pip_install: true
 
 requirements_file: requirements-test.txt
--- a/noxfile.py
+++ b/noxfile.py
@@ -90,34 +90,36 @@ def cover(session):
     session.run("coverage", "erase")
 
 
-@nox.session(python="3.8")
+black_pins = ["black==21.12b0", "isort==5.10.1"]
+
+
+@nox.session(python="3.9")
 def blacken(session):
     """Run black code formatter."""
-    session.install("black==21.5b2", "isort==5.8.0")
+    session.install(*black_pins)
     files = ["nox", "tests", "noxfile.py", "docs/conf.py"]
     session.run("black", *files)
     session.run("isort", *files)
 
 
-@nox.session(python="3.8")
+@nox.session(python="3.9")
 def lint(session):
     session.install(
-        "flake8==3.9.2",
-        "black==21.6b0",
-        "isort==5.8.0",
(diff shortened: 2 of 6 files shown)


henryiii:
(will collide with #527, but that's fine, that's a good update too :) )

noxfile.py
```diff
@@ -90,34 +90,36 @@ def cover(session):
     session.run("coverage", "erase")
 
 
-@nox.session(python="3.8")
+black_pins = ["black==21.12b0", "isort==5.10.1"]
```
DiddiLeija, line 93: Something we can do is having the linting dependencies on a requirements file, and enable Dependabot updates? That way, we don't have to stay alert for PyPI updates, and we can focus on checking what Dependabot suggests. Remember that Dependabot will only change pinned versions.

.readthedocs.yml
```diff
@@ -4,7 +4,7 @@ build:
   image: latest
 
 python:
-  version: 3.8
+  version: 3.9
```
DiddiLeija, line 7: Ah, this will conflict with #527.

DiddiLeija:
(Sorry for my last strange review, I didn't read your post :sweat_smile:).

noxfile.py
henryiii: The two projects I know of using dependabot for Python (SciPy, manylinux) both dropped it gladly. I personally use it for actions updates, for which it works really well. I don't think I use it for Python anywhere, so I can't address why they didn't like it, though I think it partially was really finicky and not interested in accepting PRs to fix Python issues. I think pre-commit.ci is superior if you'd like to go that way. But not in this PR, I could whip up a pre-commit PR after this one if you'd like to see it.

cjolowicz (approved):
Thanks! Both of the two suggested methods of automating this (pre-commit.ci, Dependabot) work for me.

Tolker-KU:
It seems that the Python version for lint and docs is pinned and 3.8 in CI. I guess that should be updated to 3.9 as well?

henryiii:
Oops, sorry! Will fix.

henryiii:
(In fact, I thought I did change this, but I guess it didn't stick)

PR #529, opened 2021-12-23
ci: fix missing 3.8 -> 3.9 for CI

henryiii:
Followup to #528

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,10 +27,10 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: 3.9
       - name: Install Nox-under-test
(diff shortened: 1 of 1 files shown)


Tolker-KU:
Maybe not in scope here, but I wonder if we should give the lint and docs step the '--error-on-missing-interpreters' flag in CI to avoid this flying under the taser again?

henryiii:
Do they need to be pinned to a specific Python, or could they just use the default?

FollowTheProcess:
> Do they need to be pinned to a specific Python, or could they just use the default?

I've often thought this when running them locally and always having to use `--force-python` to run them. I always assumed there was a valid reason behind pinning these sessions?

If this is not the case, I think just using the default is the way to go 👍🏻

FollowTheProcess:
> Maybe not in scope here, but I wonder if we should give the lint and docs steps the '--error-on-missing-interpreters' flag in CI to avoid this flying under the radar again?

This is a good idea if we cannot drop the pin for some compatibility reaason.

I've raised an issue #531 to track this, thanks for bringing it up (and quickly fixing it @henryiii !)

PR #535, opened 2021-12-25
Correct bin paths in CondaEnv for Windows

### LINKED ISSUE #514: conda bin_paths are missing `Library\...` subdirectories on windows
**Describe the bug**

When running python code that requires external executables from `conda` packages - i.e. `ipopt`, they cannot be found on the PATH.

**How to reproduce**

This noxfile fails - it should find 'ipopt.exe' on the PATH:
```python
import nox

@nox.session(python="3.7", venv_backend="conda")
def conda_path_bug(session):
    session.conda_install("conda-forge::ipopt==3.11.1")  
    session.run("python", "-c", "import shutil; assert shutil.which('ipopt')")
```

**Expected behavior**

The PATH should be set to include the same entries as Anaconda's `cwp.py` provides:
```python
new_paths = pathsep.join([prefix,
                         join(prefix, "Library", "mingw-w64", "bin"),
                         join(prefix, "Library", "usr", "bin"),
                         join(prefix, "Library", "bin"),
                         join(prefix, "Scripts")])
env = os.environ.copy()
env['PATH'] = new_paths + pathsep + env['PATH']
env['CONDA_PREFIX'] = prefix
```

Tolker-KU:
Closes #514.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -231,9 +231,16 @@ class CondaEnv(ProcessEnv):
     @property
     def bin_paths(self) -> List[str]:
         """Returns the location of the conda env's bin folder."""
-        # see https://docs.anaconda.com/anaconda/user-guide/tasks/integration/python-path/#examples
+        # see https://github.com/conda/conda/blob/f60f0f1643af04ed9a51da3dd4fa242de81e32f4/conda/activate.py#L563-L572
         if _SYSTEM == "Windows":
-            return [self.location, os.path.join(self.location, "Scripts")]
+            return [
+                self.location,
+                os.path.join(self.location, "Library", "mingw-w64", "bin"),
+                os.path.join(self.location, "Library", "usr", "bin"),
+                os.path.join(self.location, "Library", "bin"),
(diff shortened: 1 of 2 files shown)


cjolowicz (approved):
LGTM

Disclaimer: I have no knowledge of conda internals. Took a look at the `conda.activators` module. Shell activators inherit from a common base class `_Activator`. Invoking its `activate` method ends up constructing this list of `PATH` entries for Windows, via the following call chain:

```
activate
  build_activate
    _build_activate_stack
      _add_prefix_to_path
        _get_starting_path_list
          _get_path_dirs
```

See https://github.com/conda/conda/blob/f60f0f1643af04ed9a51da3dd4fa242de81e32f4/conda/activate.py#L563-L572 as quoted in the code comment.

cjolowicz:
@thatcr does this fix your issue?

Tolker-KU:
> @thatcr does this fix your issue?

I've run the example noxfile stated in the example on Windows using the patched version of Nox and it works as accepted. (Also verified that it is failing on HEAD).

PR #536, opened 2021-12-25
Update dependencies for conda_tests

Tolker-KU:
Max version of colorlog in `requirements-conda-test.txt` does not support Python > 3.7 making `nox -s conda_tests` fail. The conda_tests session in not run in CI so this is only an issue when running locally.


cjolowicz (approved):
Thanks!

PR #537, opened 2021-12-26
Deprecate use of session.install() without a virtual env

### LINKED ISSUE #503: Running with `--no-venv` causes `session.install` to use unexpected interpreter
**Describe the bug**
Running with `--no-venv` causes `session.install` to be invoked with the `python` executable found on `PATH`, not the one running `nox`.

**How to reproduce**
Running in an environment where the executables `python` and `python3.8` are distinct:
```console
$ which python python3.8
/usr/bin/python
/usr/bin/python3.8
```

My `noxfile.py` just installs `pytest`.  Running we obtain:

```
$ python3.8 -m nox --no-venv
nox > Running session validate
nox > python -m pip install pytest
nox > Command python -m pip install failed with exit code 1:
/usr/bin/python: No module named pip
nox > Session validate failed.
```

Ignore the "No module named pip", the point being it's invoking `/usr/bin/python`, not `/usr/bin/python3.8`.

**Expected behavior**
Documentation on [forcing-sessions-backend](https://nox.thea.codes/en/stable/usage.html?highlight=no-venv#forcing-the-sessions-backend) suggests this should run "on the current python interpreter (the one running nox)".

**Possible fix**
I suspect a fix could be applied in `Session.install` [here](https://github.com/theacodes/nox/blob/main/nox/sessions.py#L476) to use `sys.executable` instead of the string `"python".  I'm not certain what conditions we'd need to check beforhand (how to verify we aren't running in a venv).

Tolker-KU:
Close #503.

I've scanned the docs and cannot see anything that should be changed based on this change. Please correct me if I'm wrong.

Also feel free to chime in on the wording of the warning message.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -149,7 +149,7 @@ If you want to disable virtualenv creation altogether, you can set ``python`` to
     def tests(session):
         pass
 
-Use of :func:`session.install()` is deprecated without a virtualenv since it modifies the global Python enviroment. If this is what you really want, use :func:`session.run()` and pip instead.
+Use of :func:`session.install()` is deprecated without a virtualenv since it modifies the global Python environment. If this is what you really want, use :func:`session.run()` and pip instead.
 
 .. code-block:: python
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -149,6 +149,14 @@ If you want to disable virtualenv creation altogether, you can set ``python`` to
     def tests(session):
         pass
 
+Use of :func:`session.install()` is deprecated without a virtualenv since it modifies the global Python enviroment. If this is what you really want, use :func:`session.run()` and pip instead.
+
+.. code-block:: python
+
+    @nox.session(python=False)
+    def tests(session):
+        session.run("pip", "install", "nox")
+
 You can also specify that the virtualenv should *always* be reused instead of recreated every time:
 
 .. code-block:: python
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -20,6 +20,7 @@ import pathlib
 import re
 import sys
 import unicodedata
+import warnings
 from typing import (
     Any,
     Callable,
@@ -487,6 +488,15 @@ class Session:
         ):  # pragma: no cover
             raise ValueError(
                 "A session without a virtualenv can not install dependencies."
+            )
+        if isinstance(venv, PassthroughEnv):
+            warnings.warn(
+                f"Session {self.name} does not have a virtual environment, "
+                "so use of session.install() is deprecated since it would modify "
+                "the global Python environment. If you're really sure that is "
+                'what you want to do, use session.run("pip", "install", ...) instead.',
+                category=FutureWarning,
+                stacklevel=2,
             )
         if not args:
             raise ValueError("At least one argument required to install().")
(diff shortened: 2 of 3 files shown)


nox/sessions.py
```diff
@@ -488,6 +489,15 @@ def install(self, *args: str, **kwargs: Any) -> None:
...
             )
+        if isinstance(venv, PassthroughEnv):
+            warnings.warn(
+                f"""Session {self.name} does not have a virtual environment, """
+                """so use of session.install() is deprecated since it would modify """
+                """the global Python environment. If you're really sure that is """
+                """what you want to do, use session.run("pip", "install", ...) instead.""",
+                category=DeprecationWarning,
+                stacklevel=2,
+            )
```
cjolowicz, line 500: This needs to be `FutureWarning` for it to be displayed to end users.

I also took the liberty to replace the triple quotes, and surround the session name with single quotes. Feel free to ignore.

```suggestion
            warnings.warn(
                f"Session {self.name!r} does not have a virtual environment, "
                "so use of session.install() is deprecated since it would modify "
                "the global Python environment. If you're really sure that is "
                'what you want to do, use session.run("pip", "install", ...) instead.',
                category=FutureWarning,
                stacklevel=2,
            )
```

FTR: The `stacklevel` setting ensures that users see the file name and line number of their noxfile, instead of a location in the Nox source code. 👍

cjolowicz:
> I've scanned the docs and cannot see anything that should be changed based on this change. Please correct me if I'm wrong.

I also did not find anything.

Should we add a note to `config.rst` explaining this, though? It can go into the section introducing `venv_backend="none"`, and can be worded similar to the deprecation warning. Can happen here or in a follow-up PR.

nox/sessions.py
Tolker-KU: Thanks for the review!

docs/config.rst
```diff
@@ -149,6 +149,14 @@ If you want to disable virtualenv creation altogether, you can set ``python`` to
     def tests(session):
         pass
 
+Use of :func:`session.install()` is deprecated without a virtualenv since it modifies the global Python enviroment. If this is what you really want, use :func:`session.run()` and pip instead.
```
cjolowicz, line 152: Our new `codespell` pre-commit hook found a typo here 💯 

```suggestion
Use of :func:`session.install()` is deprecated without a virtualenv since it modifies the global Python environment. If this is what you really want, use :func:`session.run()` and pip instead.
```

Tolker-KU: Good stuff! 

Fixed 👍

cjolowicz (approved):

FollowTheProcess:
Is the intention here to keep this warning for 1 release (presumably the next one?) and then in the following release make it so that doing this will become fully disallowed? I don't use calver so I don't know if there are any specific rules regarding this.

If so we should probably have a separate issue to track this as I know at least my memory isn't that good!

cjolowicz:
> Is the intention here to keep this warning for 1 release (presumably the next one?) and then in the following release make it so that doing this will become an actual error? I don't use calver so I don't know if there are any specific rules regarding this.
> 
> If so we should probably have a separate issue to track this as I know at least my memory isn't that good!

Opened #541 for this.

cjolowicz:
Thanks @Tolker-KU

PR #538, opened 2021-12-27
Fix current Conda env detection

### LINKED ISSUE #406: Fix conda current environment detection on github actions
On github actions the following command does not output the right environment's package list:

```python
session.run(*["conda", "list"])
```

I am not sure if this is because the command is run from a specific shell (I use `shell: bash -l {0}` in the [workflow definition file](https://github.com/python-qds/qdscreen/blob/7d8e70eab86c6eb324f8dee8680aaf190d32f516/.github/workflows/base.yml#L72)).

I am able to fix this by using explicit environment variables, this is equivalent to the "conda activate <env>" command according to [this source](https://github.com/gqmelo/exec-wrappers/issues/10#issuecomment-237716664) and also [this one](https://stackoverflow.com/a/42660674/7262247):

```python
session.run(*["conda", "list"], env={"CONDA_PREFIX": session.bin, "CONDA_DEFAULT_ENV": session.get_session_id()})
```

The result of both commands (without the fix and with the fix) can be seen on the workflow [here](https://github.com/python-qds/qdscreen/runs/2141057604?check_suite_focus=true#step:10:42). You see that the first displays

```
nox > conda list
 # packages in environment at /usr/share/miniconda/envs/noxenv:
...
```

while the second displays:

```
nox > conda list
 # packages in environment at /home/runner/work/qdscreen/qdscreen/.nox/tests-2-7:
...
```

Note that for some reason on my local computer `session.bin` contains the conda env root folder, while on github actions runner this `session.bin` contains the `<conda_env_root>/bin` subfolder so [I have to remove the "/bin" suffix](https://github.com/python-qds/qdscreen/blob/b2fecb9bf70b2f952f422f61614ef09a9b9abdad/noxfile.py#L84)... any reason why this is so ?

Anyway, this experiment makes me think that maybe it would be a good idea to always add those two environment variables to the `env` when the backend is `conda`, in order to mimic the behaviour of `activate`.

What do you think ?

Tolker-KU:
Closes #406.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -203,7 +203,7 @@ class CondaEnv(ProcessEnv):
         self.reuse_existing = reuse_existing
         self.venv_params = venv_params if venv_params else []
         self.conda_cmd = conda_cmd
-        super().__init__()
+        super().__init__(env={"CONDA_PREFIX": self.location})
 
     def _clean_location(self) -> bool:
         """Deletes existing conda environment"""
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 import os
+import re
 import shutil
+import subprocess
 import sys
 from textwrap import dedent
 from unittest import mock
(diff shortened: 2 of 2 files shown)


Tolker-KU:
Lint job failed with a weird error. Can someone with permission try to re-run it?
image

FollowTheProcess:
It seems GitHub actions is being a little spotty atm!

Tolker-KU:
> It seems GitHub actions is being a little spotty atm!

Indeed! Also some timeout related fails on CI for this PR now: #537

DiddiLeija:
Something similar happened at #526. That's weird.

henryiii:
https://www.githubstatus.com

GHA is having an issue currently.

cjolowicz (approved):
LGTM

FTR #406 mentioned a few other things that we might be missing: the `CONDA_DEFAULT_ENV` variable, and custom activation scripts used by some packages. I'm fine with dealing with those other things if and when concrete use cases are reported for them. (In fact I think waiting is preferrable here as that'll give us better insight about how to solve this correctly.)

PR #539, opened 2021-12-27
Fix project's noxfile sessions definition

FollowTheProcess:
Fixes an error in the project's own `noxfile.py` where the `blacken` session was explicitly
listed in `nox.options.sessions` but no longer exists due to being replaced by the pre-commit checks introduced in #530.

Without this fix, just running `nox` will result in the following:

image

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -24,7 +24,7 @@ import nox
 ON_WINDOWS_CI = "CI" in os.environ and platform.system() == "Windows"
 
 # Skip 'conda_tests' if user doesn't have conda installed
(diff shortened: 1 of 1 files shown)


FollowTheProcess:
Weird! The lint job has failed on the `checkout code` step, not seen that before

image

FollowTheProcess:
I'll try re-running it

cjolowicz:
> Weird! The lint job has failed on the first step, not seen that before
> 
> image

I've been seeing this today in unrelated repositories, so this should not be caused by recent changes in our CI.

cjolowicz (approved):

PR #543, opened 2021-12-27
Make Session.chdir() a context manager

### LINKED ISSUE #512: Run commands from within a specific directory

Tolker-KU:
Closes #512. Also see the discussion here: #513.

I'm a bit unsure if the changes to the docs are sufficient. Please suggest changes as you see fit.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -118,6 +118,23 @@ class Status(enum.Enum):
     SKIPPED = 2
 
 
+class _WorkingDirContext:
+    def __init__(self, dir: Union[str, os.PathLike]) -> None:
+        self._prev_working_dir = os.getcwd()
+        os.chdir(dir)
+
+    def __enter__(self) -> "_WorkingDirContext":
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        os.chdir(self._prev_working_dir)
+
+
 class Session:
     """The Session object is passed into each user-defined session function.
 
@@ -212,23 +229,7 @@ class Session:
         """
         return self._runner.global_config.invoked_from
 
-    class _WorkingDirContext:
-        def __init__(self, dir: Union[str, os.PathLike]) -> None:
-            self._prev_working_dir = os.getcwd()
-            os.chdir(dir)
-
-        def __enter__(self) -> "Session._WorkingDirContext":
-            return self
-
-        def __exit__(
-            self,
-            exc_type: Optional[Type[BaseException]],
-            exc_value: Optional[BaseException],
-            traceback: Optional[TracebackType],
-        ) -> None:
-            os.chdir(self._prev_working_dir)
-
-    def chdir(self, dir: Union[str, os.PathLike]) -> "Session._WorkingDirContext":
+    def chdir(self, dir: Union[str, os.PathLike]) -> _WorkingDirContext:
         """Change the current working directory.
 
         Can be used as a context manager to automatically restore the working directory::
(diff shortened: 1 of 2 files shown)


nox/sessions.py
```diff
@@ -210,10 +212,37 @@ def invoked_from(self) -> str:
         """
         return self._runner.global_config.invoked_from
 
-    def chdir(self, dir: Union[str, os.PathLike]) -> None:
-        """Change the current working directory."""
+    class _WorkingDirContext:
```
henryiii, line 215: I really dislike nested classes and functions. The main reason to do it is for capture, which is a bad, unreadable reason. Would this be better as an un-nested class?

Tolker-KU: Certainly not. It is one of my bad habits. I've moved it out in the global scope 👍

FollowTheProcess:
Could this be simplified further to a functional context manager? i.e. you wouldn't need the `WorkingDirContext` class at all?

Something like:

```python
@contextlib.contextmanager
def chdir(self, to: str) -> Iterator:
    original = os.getcwd()
    try:
        os.chdir(to)
        yield
    finally:
        os.chdir(original)
```

I haven't done much with context managers tbh so I'm not 100% that these are equivalent

Tolker-KU:
> Could this be simplified further to a functional context manager? i.e. you wouldn't need the `WorkingDirContext` class at all?
> 
> Something like:
> 
> ```python
> @contextlib.contextmanager
> def chdir(self, to: str) -> Iterator:
>     original = os.getcwd()
>     try:
>         os.chdir(to)
>         yield
>     finally:
>         os.chdir(original)
> ```
> 
> I haven't done much with context managers tbh so I'm not 100% that these are equivalent

As far as I know the problem with this is that it would break the current functionality. 

```python
@nox.session()
def some_session(session):
    session.chdir("path/to/somewhere")
    # Clean has been run here, so we are back in old working dir
```

Using the class approach we are able to keep the current functionality at the cost of some complexity.

cjolowicz (approved):

FollowTheProcess:
> > Could this be simplified further to a functional context manager? i.e. you wouldn't need the `WorkingDirContext` class at all?
> 
> > 
> 
> > Something like:
> 
> > 
> 
> > ```python
> 
> > @contextlib.contextmanager
> 
> > def chdir(self, to: str) -> Iterator:
> 
> >     original = os.getcwd()
> 
> >     try:
> 
> >         os.chdir(to)
> 
> >         yield
> 
> >     finally:
> 
> >         os.chdir(original)
> 
> > ```
> 
> > 
> 
> > I haven't done much with context managers tbh so I'm not 100% that these are equivalent
> 
> 
> 
> As far as I know the problem with this is that it would break the current functionality. 
> 
> 
> 
> ```python
> 
> @nox.session()
> 
> def some_session(session):
> 
>     session.chdir("path/to/somewhere")
> 
>     # Clean has been run here, so we are back in old working dir
> 
> ```
> 
> 
> 
> Using the class approach we are able to keep the current functionality at the cost of some complexity.

Cool, makes sense, thanks! This looks good to me then

FollowTheProcess (approved):

PR #547, opened 2021-12-28
Pin virtualenv version in lint session

### LINKED ISSUE #545: Strange lint CI failure?
**Describe the bug**

Noticed a lint job failure after merging #543 (which had no such failure on the PR):

See [here](https://github.com/theacodes/nox/runs/4649987680?check_suite_focus=true)

image

I tried re-running but got the same outcome, then on my fork I tried adding `python -m pip install --upgrade pip setuptools wheel` to all CI jobs but got the same [outcome](https://github.com/FollowTheProcess/nox/runs/4650005563?check_suite_focus=true)

Edit: Also tried injecting setuptools directly into the nox session with `session.install("setuptools", "pre-commit")`, same result

Edit2: It seemed to always happen on `isort` so I tried removing it from the pre-commit config but it then just failed on `black` with the exact same error

**How to reproduce**

Look at the CI lint job, it seems to happen on re-runs

**Expected behavior**

Should run to success like in the PR

FollowTheProcess:
Closes #545 

virtualenv today released [20.11.0](https://github.com/pypa/virtualenv/blob/main/docs/changelog.rst#features---20110) which internally bumps the embedded version of setuptools from 58.3.0 to 59.4.0, which must have contained some breaking change for pre-commit as it depends on [virtualenv>=20.0.8](https://github.com/pre-commit/pre-commit/blob/16f68254a84df8d69431bab2ba410770746b95bd/setup.cfg#L32) so it would have pulled this latest version in

Pinning virtualenv to 20.10.0 inside the lint session has fixed the issue

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -93,7 +93,9 @@ def cover(session):
 @nox.session(python="3.9")
 def lint(session):
     """Run pre-commit linting."""
-    session.install("pre-commit")
+    # Pin virtualenv for pre-commit run
+    # See https://github.com/theacodes/nox/issues/545
+    session.install("virtualenv==20.10.0", "pre-commit")
     session.run(
         "pre-commit", "run", "--all-files", "--show-diff-on-failure", *session.posargs
     )


cjolowicz:
Great you tracked this down!

I was just going to suggest adding this the `env` section in the CI workflow:

```yaml
    VIRTUALENV_SETUPTOOLS: "!=59.4.0"
```

But this is also good.

FollowTheProcess:
> Great you tracked this down!
> 
> I was just going to suggest adding this the `env` section in the CI workflow:
> 
> ```yaml
>     VIRTUALENV_SETUPTOOLS: "!=59.4.0"
> ```
> 
> But this is also good.

Happy to change to that if we want?

cjolowicz (approved):

noxfile.py
```diff
@@ -93,7 +93,9 @@ def cover(session):
 @nox.session(python="3.9")
 def lint(session):
     """Run pre-commit linting."""
-    session.install("pre-commit")
+    # Pin virtualenv for pre-commit run
+    # See https://github.com/theacodes/nox/issues/545
+    session.install("virtualenv==20.10.0", "pre-commit")
```
gaborbernat, line 98: I don't think this was the right choice. Note virtualenv will use the latest setuptools in 28 days even on 20.10.0 :thinking:

gaborbernat: This might work for another 21 days... but then will break again.

DiddiLeija: Any suggestions to deal with this?

gaborbernat: https://github.com/pre-commit/pre-commit/issues/2178#issuecomment-1002163763 forcing setuptools to use the stdlib via an env-var is likely a better approach

DiddiLeija: I see. Maybe later I can post a PR for this.

FollowTheProcess: Thanks @gaborbernat 👍🏻

FollowTheProcess: This will be fixed in #549, thanks everyone

PR #548, opened 2021-12-28
Fix local test failure to do with FORCE_COLOR

FollowTheProcess:
I noticed that some tests are now failing on my machine with a `KeyError` because the `FORCE_COLOR` env variable is not set (which is why they don't fail on CI) and we're trying to monkeypatch the value.

image

This PR adds a quick check to see if it's set first before monkeypatching the value for tests 👍🏻

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -602,7 +602,8 @@ def test_main_noxfile_options_with_sessions_override(
 
 @pytest.mark.parametrize(("isatty_value", "expected"), [(True, True), (False, False)])
 def test_main_color_from_isatty(monkeypatch, isatty_value, expected):
-    monkeypatch.delenv("FORCE_COLOR")
+    if os.getenv("FORCE_COLOR"):
+        monkeypatch.delenv("FORCE_COLOR")
     monkeypatch.setattr(sys, "argv", [sys.executable])
     with mock.patch("nox.workflow.execute") as execute:
         execute.return_value = 0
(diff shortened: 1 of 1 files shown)


tests/test_main.py
```diff
@@ -602,7 +602,8 @@ def test_main_noxfile_options_with_sessions_override(
 
 @pytest.mark.parametrize(("isatty_value", "expected"), [(True, True), (False, False)])
 def test_main_color_from_isatty(monkeypatch, isatty_value, expected):
-    monkeypatch.delenv("FORCE_COLOR")
+    if os.getenv("FORCE_COLOR"):
+        monkeypatch.delenv("FORCE_COLOR")
```
cjolowicz, line 606: This can be abbreviated:

```suggestion
    monkeypatch.delenv("FORCE_COLOR", raising=False)
```

tests/test_main.py
```diff
@@ -627,7 +628,8 @@ def test_main_color_from_isatty(monkeypatch, isatty_value, expected):
     ],
 )
 def test_main_color_options(monkeypatch, color_opt, expected):
-    monkeypatch.delenv("FORCE_COLOR")
+    if os.getenv("FORCE_COLOR"):
+        monkeypatch.delenv("FORCE_COLOR")
```
cjolowicz, line 632:
```suggestion
    monkeypatch.delenv("FORCE_COLOR", raising=False)
```

cjolowicz (commented):
Thanks for catching this. I misread the pytest docs, thinking that `raising=False` is the default.

tests/test_main.py
FollowTheProcess: Cool! I'll do that now

henryiii (approved):
Thanks for catching this!

PR #549, opened 2021-12-28
Fix virtualenv setuptools

FollowTheProcess:
Related to discussion in #547.

Related explanation: https://github.com/pre-commit/pre-commit/issues/2178#issuecomment-1002163763

Turns out a better solution to pinning virtualenv is enforcing the stdlib setuptools in CI through an environment variable, which this PR implements

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ on: [push, pull_request]
 env:
   FORCE_COLOR: "1"
   PRE_COMMIT_COLOR: "always"
+  SETUPTOOLS_USE_DISTUTIL: "stdlib"
 jobs:
   build:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,9 @@ on: [push, pull_request]
 env:
   FORCE_COLOR: "1"
   PRE_COMMIT_COLOR: "always"
-  SETUPTOOLS_USE_DISTUTILS: "stdlib" # See https://github.com/theacodes/nox/issues/545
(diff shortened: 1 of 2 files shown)


DiddiLeija (approved):
Thanks @FollowTheProcess!

.github/workflows/ci.yml
```diff
@@ -3,6 +3,9 @@ on: [push, pull_request]
 env:
   FORCE_COLOR: "1"
   PRE_COMMIT_COLOR: "always"
+  # See https://github.com/theacodes/nox/issues/545
+  # and https://github.com/pre-commit/pre-commit/issues/2178#issuecomment-1002163763
+  SETUPTOOLS_USE_DISTUTILS: "stdlib"
```
gaborbernat, line 8: You might want to do this within nox to avoid this issue locally too, not just in the CI :thinking:

DiddiLeija: Maybe providing an `env={SETUPTOOLS_USE_DISTUTILS: "stdlib"}` to the `session.run`?

FollowTheProcess: @gaborbernat I thought that but I've been trying to recreate this issue locally and it works fine, even with the versions of virtualenv and setuptools we're discussing, which is a bit weird!

FollowTheProcess: Scratch that! Just cleaned out pre-commit's cache and managed to recreate locally. Good suggestion 👍🏻

DiddiLeija (approved):
Now, testing on a local branch and a clean venv, this works!

FollowTheProcess:
Thanks all for the help 🎉

PR #550, opened 2021-12-28
Remove MANIFEST.in in favour of setup.cfg package data

FollowTheProcess:
This PR removes `MANIFEST.in` in favour of specifying required package data in `setup.cfg`

I've validated that the produced tarballs are equivalent by running `pyproject-build` then extracting the resulting tarball to verify it's contents:

### Before (with `MANIFEST.in`)

before

### After (with `setup.cfg` package_data)

after

The only entries in `MANIFEST.in` were `py.typed` and anything with a `.jinja2` extension which you can see in the images are still present 🙂

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,0 @@
-recursive-include nox *.jinja2
-include nox/py.typed
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ tox_to_nox =
     tox
 
 [options.package_data]
-nox = py.typed
+nox = py.typed, *.jinja2
 
 [flake8]
 extend-ignore = E501, W503, E203


DiddiLeija (approved):
Sounds fine to me!

henryiii:
Keep in mind, this will make the .jinga2's available in the wheel, while before it was only in the SDist. Not saying that's wrong or right, that's just a change. (py.typed obviously should be in the wheels)

FollowTheProcess:
> Keep in mind, this will make the .jinga2's available in the wheel, while before it was only in the SDist. Not saying that's wrong or right, that's just a change. (py.typed obviously should be in the wheels)

Ahh okay, didn't know that! Are there any unwanted consequences of this that you know of?

henryiii:
Looking at it, it seems like it needs to be there to work - hmm, include_package_data is True, so this might have already been there. If that's the case, there's certainly no harm. It might be a good idea to add a check-manifest check? Unless you plan to move to Flit or something soon.

henryiii:
It's usually a nice idea to include the tests folder in the SDist, but not the wheel. Currently it's not in the SDist.

FollowTheProcess:
> It's usually a nice idea to include the tests folder in the SDist, but not the wheel. Currently it's not in the SDist.

I personally see no real problem in including the tests in the SDist 👍🏻

henryiii:
Yes, it would.

FollowTheProcess:
Okay cool, it would be good to get some more input on including tests in the SDist and if we go that way I'll close this off and we can do that instead 👍🏻

FollowTheProcess:
Closing in favour of #552 and adding tests to the SDist (which requires `MANIFEST.in`)

PR #551, opened 2021-12-28
fix: include mention of mamba in more places

### LINKED ISSUE #546: Add mamba to options, error message, etc.
Mamba was recently added as an alternative to conda (see #448) but is missing in options and error messages. See for example

https://github.com/theacodes/nox/blob/f55e9c53581d78d9373eb2868837036fd598a7a1/nox/_options.py#L319

https://github.com/theacodes/nox/blob/f55e9c53581d78d9373eb2868837036fd598a7a1/nox/sessions.py#L677

I would like to have 'mamba' added where meaningful. An approach for fixing this might be searching for 'conda' and adding 'mamba' where appropriate.

henryiii:
Closes #546, followup to #448. Mamba should be mentioned in listings of backends.

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -316,8 +316,8 @@ options.add_options(
         noxfile=True,
         merge_func=_default_venv_backend_merge_func,
         help="Virtual environment backend to use by default for nox sessions, this is ``'virtualenv'`` by default but "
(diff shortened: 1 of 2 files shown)


FollowTheProcess (approved):

FollowTheProcess:
LGTM, thanks @henryiii 👍🏻

PR #552, opened 2021-12-28
style: check manifest

henryiii:
Adds a manual check for the manifest (runs when running `nox -s lint`, but won't auto-run on `pre-commit install` or on `pre-commit.ci`, because it's kind of slow).

This adds the missing tests folder to the SDist. Some users (especially packagers like conda-forge and linux distros) like having it in the SDist.

See #550.


FollowTheProcess:
LGTM 👍🏻

FollowTheProcess (approved):

DiddiLeija (approved):

PR #556, opened 2022-01-06
fix: support changing directory with TMPDIR

### LINKED ISSUE #555: Relative env-dir breaks tools
**Describe the bug**

Nox currently sets TMPDIR to a relative path - this causes temporary directory creation to break for tools run after a `session.cd`. At least TMPDIR should be an absolute path - and maybe the return value from `session.create_tmp` should be absolute, too? (It really should be a `pathlib.Path`, but I'm referring to backwards compatible changes here ;) )

(see https://github.com/scikit-hep/cookie/issues/60 - Maturin expects TMPDIR to be an existent directory)

**How to reproduce**
Use chdir then try to run a tool that makes temporary directories.

**Expected behavior**
Temporary directory creation should not be broken after changing directories.

henryiii:
This is a minimal fix to the bug, though maybe `create_tmp` should also be absolute? That's less important, because someone could construct the correct path, while TMPDIR simply breaks.

Closes #555

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -202,7 +202,7 @@ class Session:
         """Create, and return, a temporary directory."""
         tmpdir = os.path.join(self._runner.envdir, "tmp")
         os.makedirs(tmpdir, exist_ok=True)
-        self.env["TMPDIR"] = tmpdir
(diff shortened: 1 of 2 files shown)


cjolowicz (approved):

cjolowicz:
Thank you!

PR #560, opened 2022-01-11
docs: fix readthedocs

### LINKED ISSUE #559: readthedocs broken
**Describe the bug**

The docs are broken, the "stable" release did not build after the recent release (missing from changelog), with:

```
AttributeError: 'HTMLParser' object has no attribute 'unescape'
```

See https://readthedocs.org/projects/nox/builds/

This should be fixable by setting a minimum version on some of the docs dependencies (why are those mixed with the test dependencies?) - readthedocs installs old versions of things by default, and then it's a no-op if you don't force an upgrade. Why they install a version of (something that uses HTML) that doesn't support Python 3.9+, I'm not sure, but forcing sphinx>=4 or something like that will likely fix the problem. Downgrading to Python 3.8 for the docs would fix it, too, but that's the wrong way to initially try to fix it.

I would highly recommend enabling PR readthedocs builds (in the readthedocs settings), they are very handy for catching bugs in the docs builds.

Also, it looks like "latest" hasn't built in a year or more, I'm guessing it was never changed to `main` from `master`, perhaps? It makes "latest" much older than "stable".

henryiii:
Closes #559.

I have a couple of followup ideas, but this should at least get the docs building again.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -2,6 +2,6 @@ flask
 pytest
 pytest-cov
 recommonmark
-sphinx
+sphinx>=3.0
 sphinx-autobuild
 witchhazel


henryiii:
Not seeing a RtD PR build.

dhermes:
> Not seeing a RtD PR build.

Ditto; I unfortunately don't have admin on this repo so can't check if the webhook is installed

dhermes (approved):
LGTM; let's just merge and see if it resolves

henryiii:
Should be quite safe, it just might not fix the problem. :)

henryiii:
(Even if the web hook was enabled, sometimes if it was old it might need to be refreshed - I feel like I've had to fight with it before)

If this works, _most_ issues should be caught by the GHA docs build (except this one).

dhermes:
@henryiii 🤞  https://readthedocs.org/projects/nox/builds/15742871/

henryiii:
Ahh, I _almost_ replaced recommonmark with myst-parser ([it's deprecated](https://recommonmark.readthedocs.io/en/latest/)), but then I left it for followup. Looks like that's what triggering it. Instead of looking up the version of recommonmark needed, let's just go ahead and upgrade to myst.

PR #561, opened 2022-01-11
docs: use myst-parser for md

### LINKED ISSUE #559: readthedocs broken

henryiii:
Closes #559 again.

See https://github.com/readthedocs/recommonmark/issues/221 - recommonmark has been sun-setted, myst-parser is the replacement.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -39,7 +39,7 @@ sys.path.insert(0, os.path.abspath("."))
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",
-    "recommonmark",
+    "myst_parser",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,7 +1,7 @@
 flask
+myst-parser
 pytest
 pytest-cov
-recommonmark
 sphinx>=3.0
 sphinx-autobuild
 witchhazel


FollowTheProcess:
LGTM, might it be worth another PR build to check? @dhermes

henryiii:
I think we were just merging then trying to rebuild readthedocs, since PR docs builds are not working. But this does work both locally and with the GHA docs build.

cjolowicz (approved):

henryiii:
https://readthedocs.org/projects/nox/builds/15750719/ :D

dhermes:
The point of `stable` is that it points at a **released** version of the code / docs. Not really sure what a good strategy is for that (short of cutting a release).

henryiii:
I believe no changes have been made since the last release, so rebuilding with the current main would be identical to the last release. (As long as you don't leave it on current main).

cjolowicz:
Another possibility would be to make a [post-release](https://www.python.org/dev/peps/pep-0440/#post-releases), `2022.1.7.post1`. That would trigger the stable build, and give us a PyPI release that's clearly marked as a minor correction without user-visible changes.

henryiii:
Post releases should be metadata only (which is pretty much exactly what this is - docs), so that would also be fine. Though I'd still think that's a little overkill, the point of stable is to have docs for the latest release that don't also include features/changes added since the latest release, not that they have to be tied to an exact tag, but yes, it's fine.

henryiii:
(In pybind11 we have a `stable` branch and that's what the `stable` docs builds from, which allows us to make docs fixes after a release if we really need to - thought usually we don't)

PR #567, opened 2022-02-08
Missing interpreters now error the session on CI by default

### LINKED ISSUE #340: Change default behavior for `--error-on-missing-interpreters`
Is it possible to improve `--error-on-missing-interpreters` with "principle of least surprise" in mind? We don't want onerous failures during local dev, but the two changes are proposed:

- Mention the `--error-on-missing-interpreters` flag in the error / warning / success message if an interpreter was skipped
- Change `--error-on-missing-interpreters` to default **ON** (instead of **OFF**) if we've detected the code is in CI (`CI=true`) or in an automated environment (`sys.stdout.isatty()==False`)

----

**Original Content** (Original title: Test suite skipped with error "Python interpreter 3.8 not found.)

**Describe the bug**

The desired Python version was not available, the test suite is skipped entirely.

```
2020-08-21T17:46:46.1946078Z ##[group]Run nox
2020-08-21T17:46:46.1946291Z [36;1mnox[0m
2020-08-21T17:46:46.1990422Z shell: /bin/bash -e {0}
2020-08-21T17:46:46.1990565Z env:
2020-08-21T17:46:46.1990705Z   pythonLocation: /opt/hostedtoolcache/Python/3.7.8/x64
2020-08-21T17:46:46.1990822Z ##[endgroup]
2020-08-21T17:46:46.2954365Z Running session tests-3.8
2020-08-21T17:46:46.3002157Z Session tests-3.8 skipped: Python interpreter 3.8 not found.
```
image

**How to reproduce**

```
import nox

@nox.session(python=["3.7"])
def tests(session):
    """Run pytest suite"""
    session.create_tmp()
    session.install("-r", "requirements-test.txt")
    tests = session.posargs or ["tests/"]
    session.run("pytest", *tests)

# TODO: add lint, type checking, etc.
```

```
$ cat .github/workflows/data_prep_tools_test.yml 
# This workflow will install Python dependencies, run tests.
# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions

name: test

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
    
defaults:
  run:
    working-directory: tools

jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: [3.7]

    steps:
    - uses: actions/checkout@v2
    - name: Set up Python ${{ matrix.python-version }}
      uses: actions/setup-python@v2
      with:
    - name: Install dependencies
      run: |
        pip install nox
    - name: Test with nox
      run: |
        nox
```

**Expected behavior**

non-zero exit code

FollowTheProcess:
Closes #340 

This PR adds a check to determine if nox is running on CI when evaluating sessions with missing interpreters so that such sessions fail by default when being run on CI providers.

This has the desired effect:

**Skipped locally showing CI warning**

image

**CI=true causes an error**

image

*Note non-zero exit code*

**`--error-on-missing-interpreters` for reference**

image

I have some questions I'd like people's input on:

- Should this be documented or just rely on the warning message?
- The warning message is kinda long, any other suggestions that get the point across? Maybe just say it will fail in CI and leave it there without mentioning `--error-on-missing-interpreters`? Given that it will show any time a session is skipped a long message could get annoying.
- Currently we just check for `CI` existing in `os.environ` should it explicitly check for true?

### DIFF
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -157,6 +157,7 @@ def make_flag_pair(
     name: str,
     enable_flags: tuple[str, str] | tuple[str],
     disable_flags: tuple[str],
+    default: bool = False,
     **kwargs: Any,
 ) -> tuple[Option, Option]:
     """Returns two options - one to enable a behavior and another to disable it.
@@ -172,6 +173,7 @@ def make_flag_pair(
         *enable_flags,
         noxfile=True,
         merge_func=functools.partial(flag_pair_merge_func, name, disable_name),
+        default=default,
         **kwargs,
     )
 
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -419,7 +419,7 @@ options.add_options(
         ("--no-error-on-missing-interpreters",),
         group=options.groups["execution"],
         help="Error instead of skipping sessions if an interpreter can not be located.",
-        default=True if "CI" in os.environ else False,
+        default="CI" in os.environ,
     ),
     *_option_set.make_flag_pair(
         "error_on_external_run",
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -419,6 +419,7 @@ options.add_options(
         ("--no-error-on-missing-interpreters",),
         group=options.groups["execution"],
         help="Error instead of skipping sessions if an interpreter can not be located.",
+        default=True if "CI" in os.environ else False,
     ),
     *_option_set.make_flag_pair(
         "error_on_external_run",
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -901,9 +901,10 @@ class TestSessionRunner:
         assert result.status == nox.sessions.Status.SKIPPED
         assert "no parameters" in result.reason
 
-    def test_execute_skip_missing_interpreter(self, caplog):
+    def test_execute_skip_missing_interpreter(self, caplog, monkeypatch):
         runner = self.make_runner_with_mock_venv()
         runner._create_venv.side_effect = nox.virtualenv.InterpreterNotFound("meep")
+        monkeypatch.delenv("CI", raising=False)
 
         result = runner.execute()
 
@@ -913,12 +914,12 @@ class TestSessionRunner:
             "Missing interpreters will error by default on CI systems." in caplog.text
         )
 
-    def test_execute_missing_interpreter_on_CI(self):
+    def test_execute_missing_interpreter_on_CI(self, monkeypatch):
         runner = self.make_runner_with_mock_venv()
         runner._create_venv.side_effect = nox.virtualenv.InterpreterNotFound("meep")
+        monkeypatch.setenv("CI", "True")
 
-        with mock.patch.dict(os.environ, {"CI": True}):
-            result = runner.execute()
+        result = runner.execute()
 
         assert result.status == nox.sessions.Status.FAILED
         assert "meep" in result.reason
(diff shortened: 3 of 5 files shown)


DiddiLeija:
> Should this be documented or just rely on the warning message?

I _think_ some documentation is OK. Maybe a little note at https://nox.thea.codes/en/latest/tutorial.html#testing-against-different-and-multiple-pythons?

> Maybe just say it will fail in CI and leave it there without mentioning `--error-on-missing-interpreters`?

If that's the case, I would prefer some documentation, to say what this warning won't say?

DiddiLeija (approved):
Looks good for me now :)

FollowTheProcess:
@henryiii Does this work for you? Since #340 had quite a lot of discussion around this I'd like to get lots of eyes on this to make sure everyone is happy 🙂

cjolowicz:
Can we make it so `CI` determines just the *default* for the option?

This has two advantages IMO:

1. There is a single source of truth for the setting.
2. You can override the default in CI using `--no-error-on-missing-interpreters`. IOW overriding works both ways.

FollowTheProcess:
> Can we make it so CI determines just the default for the option?

Sure, that's a good idea, let me play around some more and get back to you 👍🏻

FollowTheProcess:
Huh, thought I had it working... standby 😂

nox/_option_set.py
```diff
@@ -157,6 +157,7 @@ def make_flag_pair(
     name: str,
     enable_flags: tuple[str, str] | tuple[str],
     disable_flags: tuple[str],
+    default: bool = False,
```
cjolowicz, line 160:
```suggestion
```

The parameter can be passed with kwargs.

nox/_option_set.py
```diff
@@ -172,6 +173,7 @@ def make_flag_pair(
         *enable_flags,
         noxfile=True,
         merge_func=functools.partial(flag_pair_merge_func, name, disable_name),
+        default=default,
```
cjolowicz, line 176:
```suggestion
```

nox/_options.py
```diff
@@ -419,6 +419,7 @@ def _session_completer(
         ("--no-error-on-missing-interpreters",),
         group=options.groups["execution"],
         help="Error instead of skipping sessions if an interpreter can not be located.",
+        default=True if "CI" in os.environ else False,
```
cjolowicz, line 422:
```suggestion
        default="CI" in os.environ,
```

tests/test_sessions.py
```diff
@@ -749,7 +749,7 @@ def make_runner(self):
                 envdir="envdir",
                 posargs=[],
                 reuse_existing_virtualenvs=False,
-                error_on_missing_interpreters=False,
+                error_on_missing_interpreters=True if "CI" in os.environ else False,
```
cjolowicz, line 752:
```suggestion
                error_on_missing_interpreters="CI" in os.environ,
```

cjolowicz (commented):
Thanks!

nox/_option_set.py
FollowTheProcess: I tried that first time round, but `make_flag_pair` passes the kwargs to both the enable and disable options so they cancel each other out. The default in the above case is only passed into the enable option.

nox/_options.py
FollowTheProcess: Nice 👍

nox/_option_set.py
cjolowicz: Oh, right.

FollowTheProcess:
Cool! Thanks for the review, should be good to go now if everyone's happy 🙂

PR #577, opened 2022-02-15
tox-to-nox: handle non-identifier env names

### LINKED ISSUE #573: tox-to-nox doesn't handle non-identifier env names
**Describe the bug**
tox env names can contain dash characters (and potentially other special characters) and tox-to-nox does not handle this well.

**How to reproduce**
Create a simple tox.ini like the one below.

```ini
[testenv:name-dash]

```

Run tox-to-nox and check the output.

```python
import nox


@nox.session()
def name-dash(session):
    session.install('.')
```

**Expected behavior**
For dashes, it's possible to translate the dashes to underscores. Another possibility is to fail, or at least warn that non-identifier characters were detected and the user should fix the env names before running nox.

ktbarrett:
Closes #573. Introduces a function which translates dashes to underscores in tox env names and prints a warning if any other non-identifier character is seen in a tox env name.

### DIFF
--- a/nox/tox_to_nox.jinja2
+++ b/nox/tox_to_nox.jinja2
@@ -2,7 +2,7 @@ import nox
 
 {% for envname, envconfig in config.envconfigs.items()|sort: %}
 @nox.session({%- if envconfig.basepython %}python='{{envconfig.basepython}}'{%- endif %})
-def {{envname}}(session):
+def {{fixname(envname)}}(session):
     {%- set envs = dict(envconfig.setenv) -%}
     {%- do envs.pop('PYTHONHASHSEED', None) -%}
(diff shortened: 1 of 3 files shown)


ktbarrett:
> Also would raising an exception be better here if we can't parse the session name?

That's an option. Not sure if it's better. If tox-to-nox fails, people are just going to change their tox.ini and re-run until it succeeds, the current approach doesn't require them to modify their tox.ini. It warns them once when the script is run, then Python throws a SyntaxError if they do not heed those warnings.

FollowTheProcess:
Yeah okay I see your point. Happy with this, thanks!

FollowTheProcess (approved):

FollowTheProcess:
Thanks @ktbarrett for the contribution!

PR #586, opened 2022-03-13
Fix typo in tutorial.rst

zhanpon:
This line should be `session.install("pytest")` from the context.


DiddiLeija (approved):
Good catch! Thank you!

Just waiting a review from another maintainer.

cjolowicz (approved):
Thanks!

PR #591, opened 2022-04-04
Add VENV_DIR to `dev` session in cookbook, fix #588

### LINKED ISSUE #588: Add an example `VENV_DIR` constant to `dev` session in cookbook?
Hi `nox` devs and maintainers, thank you for this great tool.

I'd like to suggest a very minor tweak to the `dev` session example in the cookbook
https://nox.thea.codes/en/stable/cookbook.html#instant-dev-environment

This example is exactly one of the things that I was hoping `nox` could give me.  
But I got confused because I defined my `VENV_DIR` as simply `pathlib.Path('.venv')` which resulted in the  following cryptic error:
```console
$ nox -s dev      
nox > Running session dev
nox > Creating virtual environment (virtualenv) using python in .nox/dev
nox > python -m pip install virtualenv
nox > virtualenv .venv
nox > Program .venv/bin/python not found.
nox > Session dev failed.
```

After staring at examples in Scikit-HEP I realized that what I needed to do was chain a call of `Path.resolve()` method onto my `VENV_DIR`:
```python
VENV_DIR = pathlib.Path('./.venv').resolve()   # nox will find this because of the absolute path 
```
Not quite clear to me why my simple `pathlib.Path('venv')` without the `resolve` doesn't work -- somehow `nox` is runnning things from a place that is not in my `cwd`? (I realize this is just my blissful ignorance about the $PATH speaking)

Anyway, I'm suggesting to just add an example `VENV_DIR` like this to the `dev` recipe, e.g.
```python
import os

import nox

# It's a good idea to keep your dev session out of the default list
# so it's not run twice accidentally
nox.options.sessions = [...] # Sessions other than 'dev'

VENV_DIR = pathlib.Path('./.venv').resolve()   # need to use absolute path 

@nox.session
def dev(session: nox.Session) -> None:
    ...
```
There is a comment explaining that `VENV_DIR` needs to be defined but the bit about needing to `resolve()` it would help I think

NickleDave:
This adds a `VENV_DIR` constant to the example `dev` session in the cookbook as described in #588

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -47,6 +47,11 @@ Enter the ``dev`` nox session:
     # so it's not run twice accidentally
     nox.options.sessions = [...] # Sessions other than 'dev'
 
+    # this VENV_DIR constant specifies the name of the dir that the `dev`
+    # session will create, containing the virtualenv;
+    # the `resolve()` makes it portable
+    VENV_DIR = pathlib.Path('./.venv').resolve()
+
     @nox.session
     def dev(session: nox.Session) -> None:
         """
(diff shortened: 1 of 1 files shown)


FollowTheProcess:
Looks great to me! Thanks @NickleDave 🎉

NickleDave:
Thank you!

PR #592, opened 2022-04-06
refactor: drop path.py

### LINKED ISSUE #583: Remove usage of py.path?
Currently, there is a dependency on `py.path`, which pytest is trying to remove, and it seems like nox should remove it too, now that Python 3 is required (since quite a while, actually). py.path is not recommended for use: https://py.readthedocs.io/en/latest/path.html. This would reduce a dependency, too!

Looking through the usages, I see `py.path.local.sysfind`, which with a little effort could be replaced by Python 3's [shutil.which](https://docs.python.org/3/library/shutil.html#shutil.which). I see `as_cwd`, which could be handled pretty easily (ideally looking like the [3.11 context addition](https://docs.python.org/3.11/library/contextlib.html#contextlib.chdir)). That seems to be it for non-test usages.

henryiii:
Closes #583. Removes a deprecated dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,9 +10,13 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-20.04, windows-2019]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
+        include:
+          - os: macos-latest
+            python-version: "3.9"
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
--- a/nox/command.py
+++ b/nox/command.py
@@ -39,7 +39,6 @@ class CommandFailed(Exception):
 
 def which(program: str, paths: list[str] | None) -> str:
     """Finds the full path to an executable."""
-    full_path = None
 
     if paths is not None:
         full_path = shutil.which(program, path=os.pathsep.join(paths))
--- a/nox/command.py
+++ b/nox/command.py
@@ -16,10 +16,9 @@ from __future__ import annotations
 
 import os
 import shlex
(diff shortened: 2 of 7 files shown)


henryiii:
I'm not likely to have time to power up a Windows machine to investigate why this is failing on Windows for a couple of days - in a workshop this week. If PATHEXT is off, then don't see why anything works, and if it's right, I don't see why this isn't finding things correctly. Maybe I've got about around `py` handling. 🤷

Also need to think about how to add a test for the "not found locally but found in the default path" branch for coverage, which is showing up on ubuntu _except for 3.10_. Which is odd.

DiddiLeija:
Let me check the Windows thing around the next days :wink:

henryiii:
I think such a scenario would be looking for something like `git`. It would not be available in the local paths that are being passed, but would be available in the system paths, so it would only be detected in the second run (AFAICT, if you give custom paths, the system paths are then not included).

henryiii:
That's embarrassing. Just don't look in history and pretend I wrote it correctly the first time. :)

In my defense, I somehow thought the new-in-3.7 `text=True` parameter also implied `capture_output`, even though I also knew it was just an alias for `universal_newlines=True`. That would have been handy, I think...

Also I needed to boot up a Windows box for other things, though I do feel slightly ridiculous needed it to just discover this.

I'd really have liked it if this was protected by type - if mypy could tell you couldn't access .stdout/stderr without capturing output in some way, for example!

henryiii:
Hmm, so it looks like `shutil.which` expects executables to follow Window's `PATHEXT` and looks for `.exe`, `.bat`, and such. The tests make a no-extension executable, which `shutil` doesn't find but the `py.path` tooling did. I'll have to investigate a bit further to see what the best solution is - it's probably at most just needed in the `commands.which` function if a workaround is needed.

PR #593, opened 2022-04-07
Drop 99% coverage threshold flag for 3.10 in noxfile

FollowTheProcess:
This PR drops the flag in our Noxfile that would set the `fail_under` threshold to 99% on python 3.10. This was a hangover from when we were testing against 3.10.0-rc.2 in GHA and coverage was raising a few false positives.

Judging by recent runs this flag can be removed now.

See this comment for further details: https://github.com/wntrblm/nox/pull/592#issuecomment-1091830649

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -68,16 +68,9 @@ def cover(session):
     if ON_WINDOWS_CI:
         return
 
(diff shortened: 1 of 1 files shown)


henryiii (approved):
:heart:

PR #594, opened 2022-04-07
feat: GitHub Action

henryiii:
The officially recommended action, `excitedleigh/setup-nox`, has been untouched for a year or so now, with [PRs building up](https://github.com/excitedleigh/setup-nox/pulls), and now is broken due to a PyPy update on the runners - https://github.com/excitedleigh/setup-nox/issues/419. All usages on all GitHub hosted runners are broken.

In the last year, GitHub has made it dramatically easier for a library to provide it's own action with composite actions, then recently composite actions actually became able to compose other actions (which, why was that not a launch feature for something called "composite" actions, but better late than never...). So it's now possible to provide a very simple action (than can be pinned to an exact nox version, and auto-updated via dependabot!) with minimal maintenance burden. We've been [providing one with cibuildwheel](https://github.com/pypa/cibuildwheel/blob/main/action.yml) for quite a while now, and it's been very popular has has worked well, especially with dependabot.

Unlike the javascript setup-nox action above, it's really easy to place any Python versions in this action, and have full control over when they are added/removed. This means there's also some choice here - should it activate all non-EOL Pythons, or all Pythons supported by Nox? There's a small[^1] time cost in activating a Python version that's pre-built but not pre-installed (must be downloaded), so I think all non-EOL Pythons is best - this action gracefully blends with adding setup-python's for the other Pythons. This also is a pretty close match for the current setup-nox action.

I'm got an example/test run here: https://github.com/scikit-build/scikit-build/pull/692.

[^1]: About 6 seconds on linux/windows, and 18 seconds on macOS, AFAICT. On linux/windows, these sometimes are already there - 2.7 is, 3.6 sometimes is.


FollowTheProcess:
Love this! Big +1 from me. Also a TIL I didn't know you could make actions this easily!

docs/tutorial.rst
```diff
@@ -27,17 +27,18 @@ Or you can be extra fancy and use `pipx`_:
...
+If you want to run ``nox`` within `GitHub Actions`_, you can use the `wntrblm/nox action`_, which installs the latest ``nox`` and makes available all active CPython and PyPY versions provided by the GitHub Actions environment. You can safely combine this with with ``setup-python`` for past end-of-life or development versions of Python, as well.
 
 .. _pip: https://pip.readthedocs.org
 .. _user site: https://packaging.python.org/tutorials/installing-packages/#installing-to-the-user-site
 .. _pipx: https://packaging.python.org/guides/installing-stand-alone-command-line-tools/
 .. _docker: https://www.docker.com/
 .. _thekevjames/nox images: https://hub.docker.com/r/thekevjames/nox
 .. _GitHub Actions: https://github.com/features/actions
-.. _excitedleigh/setup-nox action: https://github.com/marketplace/actions/setup-nox
+.. _wntrblm/nox action: https://github.com/marketplace/actions/nox
```
henryiii, line 40: FYI, this link assumes you click "publish this action" after this in GitHub Actions. In cibuildwheel, we've not done that (to avoid a "this is an action in the marketplace" banner at the top to make it look like it's only GitHub Actions). If you'd prefer not to publish to the marketplace" (which doesn't really do anything other than make it a tiny bit more searchable), then this link should point somewhere else, or not be a link at all.

I did check to make sure the name "nox" has not been taken yet.

FollowTheProcess: I can't see any harm in publishing it in this way, it'd be nice to have an official nox action in the marketplace :) If any other maintainers feel differently though please chime in!

henryiii:
I've got an example using the fork this PR is from linked above, here it is again:

https://github.com/scikit-build/scikit-build/pull/692

This would be a minimal steps

```yaml
steps:
- uses: actions/checkout@v3

- name: Setup nox
  uses: wntrblm/nox

- run: nox
```

@excitedleigh fixed the action 9 days ago! But didn't release a new version, so dependabot won't update it (needs a tag). I think it would be fine to list/have both (if it gets a working release) - that action might be slightly faster, but this action is version locked to nox (since it _is_ nox). A in-repository action like the one in this PR can actually pin via nox version automatically and be updated via dependabot, something other actions can't do.

FollowTheProcess:
@wntrblm/nox Any thoughts or objections? If no one objects I'll merge this in and probably publish the action to the GitHub marketplace (if I can do that with member perms, I don't actually know 🤷🏻‍♂️) as discussed up here: https://github.com/wntrblm/nox/pull/594#discussion_r845575202

docs/tutorial.rst
FollowTheProcess: It just clicked what you meant by this @henryiii! For anyone else as slow to catch on as me, this is what we mean: 

image

If we publish the action in the marketplace, the Nox repo will have that banner above it which implies it's just a GitHub action so IMO it's best to avoid as @henryiii did with cibuildwheel

DiddiLeija:
No objections from my side :+1:

theacodes:
Sounds great!

daisylb:
Well, now I can cross a lot of things off of my to-do list <3

PR #595, opened 2022-04-12
Add some missing generic type arguments

RobertCraigie:
This PR adds generic type arguments to type hints in `sessions.py`, I did not have time to go through the whole codebase unfortunately.

The benefit of adding type arguments is that type checkers like [Pyright](https://github.com/microsoft/pyright) will complain if you access a variable that has a potentially unknown type, for example I ran into this:

```py
import nox

@nox.session(reuse_venv=True)
def pyright(session: nox.Session) -> None:
    session.env['ENV_VARIABLE'] = '1'
    # Error: Type of "env" is partially unknown
    #            Type of "env" is "dict[Unknown, Unknown]"
```

If you want to avoid errors like these in the future, I'd highly recommend adding pyright to CI and using [--verifytypes](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#verifying-type-completeness).

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -108,7 +108,7 @@ class Status(enum.Enum):
 
 
 class _WorkingDirContext:
-    def __init__(self, dir: str | os.PathLike) -> None:
+    def __init__(self, dir: str | os.PathLike[str]) -> None:
         self._prev_working_dir = os.getcwd()
         os.chdir(dir)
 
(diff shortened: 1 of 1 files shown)


henryiii:
I'd recommend turning up the strictness of mypy, and then this would be enforced. I can take a stab at it.

FollowTheProcess:
Hi @RobertCraigie thanks for the PR! This looks great to me

FollowTheProcess:
I'm gonna get this one in so @RobertCraigie is recognised as a contributor then planning to merge #596 straight after to enforce going forward, unless anyone from the maintainer team screams "No!" in the meantime 👍🏻

PR #596, opened 2022-04-19
fix(types): full strictness checking on mypy

henryiii:
This is a followup to (and includes) #595. Enabling full strictness checking for mypy and bumping the version of mypy (which removes a few ignores).

There are still two sets of ignores that can likely go away in the future. There are some around py.path that will go away with #592. There are also likely some related to function decorators - by properly typing decorators with generics, we can tighten that up and make those work better too (but that's a followup for some other time).

Closes #595 if this gets merged before that, since it contains that commit.

- Add some missing generic arguments
- fix(types): add full strict checking
- fix(types): enable extra error checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,7 +63,7 @@ repos:
         additional_dependencies: *flake8-dependencies
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.941
+    rev: v0.942
     hooks:
       - id: mypy
         files: ^nox/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,11 +63,11 @@ repos:
         additional_dependencies: *flake8-dependencies
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.931
+    rev: v0.941
     hooks:
       - id: mypy
         files: ^nox/
-        args: [--show-error-codes]
+        args: []
         additional_dependencies:
           - types-jinja2
           - packaging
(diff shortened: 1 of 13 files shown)


nox/_decorators.py
```diff
@@ -34,24 +34,27 @@ def __new__(
...
 
-def _copy_func(src: Callable, name: str | None = None) -> Callable:
+T = TypeVar("T", bound=Callable[..., Any])
+
+
+def _copy_func(src: T, name: str | None = None) -> T:
     dst = types.FunctionType(
         src.__code__,
-        src.__globals__,  # type: ignore[attr-defined]
+        src.__globals__,
```
henryiii, line 43: These were fixed upstream in typeshed and included in mypy 0.94x.

nox/_option_set.py
```diff
@@ -83,7 +83,11 @@ def __init__(
         noxfile: bool = False,
         merge_func: Callable[[Namespace, Namespace], Any] | None = None,
         finalizer_func: Callable[[Any, Namespace], Any] | None = None,
-        default: Any | Callable[[], Any] = None,
+        default: bool
+        | str
+        | None
+        | list[str]
+        | Callable[[], bool | str | None | list[str]] = None,
```
henryiii, line 90: This should probably be generic eventually - `Option[str]` would hold a str, etc. But didn't look into it.

.pre-commit-config.yaml
```diff
@@ -63,11 +63,11 @@ repos:
         additional_dependencies: *flake8-dependencies
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.931
+    rev: v0.941
```
henryiii, line 66:
```suggestion
    rev: v0.942
```

forgot about the 2.

PR #597, opened 2022-04-19
tests: configure pytest

henryiii:
This configures pytest using pyproject.toml, and simplifies the noxfile a tiny bit - pytests's testfiles already has the correct behavior for session.posargs.


FollowTheProcess:
LGTM! Nice work 👍🏻

PR #598, opened 2022-04-19
chore: minor touchups

henryiii:
Several minor touchups too small on their own.

- chore: remove wheel
- chore: include flake8-bugbear's extra checks
- chore: fix pycln's arg in pre-commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
     rev: v1.2.5
     hooks:
       - id: pycln
-        args: [--config=pyproject.toml]
+        args: [--all]
         stages: [manual]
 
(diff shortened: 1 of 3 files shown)


FollowTheProcess:
Looks good @henryiii thanks! On a bit of a roll at the moment! Just for my own curiosity (and laziness to find out by myself 😄) why do we not need wheel anymore?

henryiii:
Also see https://github.com/pypa/packaging.python.org/pull/1050

henryiii:
> On a bit of a roll at the moment!

At least partially inspired by https://scikit-hep.org/developer/reporeview :)

FollowTheProcess:
> At least partially inspired by https://scikit-hep.org/developer/reporeview :)

This is super cool!

Issue #534, opened 2021-12-24
Open discussions, gitter, something


ktbarrett:
**How would this feature be useful?**
Issues only have "feature request" or "bug report" format, no question format; so I assume issues are not for questions. Github discussions are disabled. There is no gitter or discord linked. So, where are questions or discussions about nox supposed to occur?

**Describe the solution you'd like**
I would like to see a gitter or discord. I participate in and moderate a few gitter channels and really enjoy interactive problem solving, tangent discussions, and general sense of community that comes from that.

**Describe alternatives you've considered**
StackOverflow has a nox tag, but it is solely short Q&A format and can be pretty toxic, but it is searchable. Github Discussions can be free-form or Q&A format, they are searchable, and integrate with Github; however, in my experience people tend to not understand how to use it correctly (maybe this is a UX problem?).

theacodes:
We can use the Winterbloom discord, since I'm considering moving Nox over to github.com/wntrblm so it's easier to manage the contributors and notifications.

I'll make a #nox channel and y'all can mute notifications for everything else if you want.

https://discord.gg/D35cxRxc

DiddiLeija:
> I'm considering moving Nox over to github.com/wntrblm so it's easier to manage the contributors and notifications.

And... it happened :partying_face:

> https://discord.gg/D35cxRxc

Seems like this invite became invalid (or expired? [^1]). @theacodes, could you check if this invite is still functional, and provide a new one if not? That way, more people with intentions of joining will be there [^2].

[^1]: I suspect the invite expired (but maybe I'm wrong). Maybe a new invite without expire date is the solution?
[^2]: Also, what about mentioning this in the docs? Or maybe in the project's README? Maybe somewhere else?

FollowTheProcess:
Just further to this, as part of #600 I'm looking at including a link to the discord in the issue form config so it shows up as a method to ask a more general question not covered by the more rigid bug/feature forms. Much like [pip](https://github.com/pypa/pip) do here with IRC:

<img width="1276" alt="image" src="https://user-images.githubusercontent.com/62767721/165463110-70d0297e-5f62-4909-b84f-426fcf7d800f.png">

Is it possible to have a permanent invite link to the discord? If not this approach may not work as intended as we'd have to keep refreshing the link.

DiddiLeija:
> Is it possible to have a permanent invite link to the discord? If not this approach may not work as intended as we'd have to keep refreshing the link.

[...]

PR #600, opened 2022-04-23
Introduce GitHub issue forms

### LINKED ISSUE #585: Provide labels at issue templates?

Currently, when a user creates an issue, it doesn't has labels. Maybe we can alleviate the maintainer's work by automating the labels addition.

Add labels to the issue templates: `bug` at the bug template and `enhancement` at the feature request, maybe?

If we keep as is, it's ok for me too :+1:

FollowTheProcess:
Closes #585 

I've taken an initial stab at porting our issue templates to the new [GitHub forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates).

I've tried to keep the intent of the originals but also threw some of the nice features of the new templates at them for good measure!

I've also added a new form for a generic question about Nox as I believe this is useful for people who want to know something but don't want to raise a "scary" sounding bug or issue.

More than happy to tweak the contents if we want, just kicking things off 👍🏻 

You can view the new forms by clicking on the files and GitHub will render it how it will look when filing the issue 🙂


.github/ISSUE_TEMPLATE/question.yml
```diff
@@ -0,0 +1,20 @@
+name: ❓ Question
```
DiddiLeija, line 1: Are questions encouraged in this repo? Maybe I misunderstood #534, but maybe the Winterbloom Discord server (under the `#nox` channel) could fit better for questions.

However, we could keep this template, and also provide a link in a `ISSUE_TEMPLATE/config.yml` [^1] file for those who want to ask on DIscord.

Just tagging @wntrblm/nox to attract the maintainers' attention, to get some more opinions.

[^1]: `config.yml` reference is [here at the GitHub docs](https://docs.github.com/es/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser).

DiddiLeija (commented):
Strong +1 from me! I watched how these new templates worked in [pip](https://github.com/pypa/pip) and some other projects, and I think they will look pretty nice here!

.github/ISSUE_TEMPLATE/question.yml
FollowTheProcess: I like that idea!

DiddiLeija: I think I can write a `config.yml` after this PR, unless someone disagrees :)

crwilcox (approved):
+1 on this. Only comment is if we wouldn't benefit from making some fields required. But this is the closest to a port, by not doing that, and perhaps worth further discussion. 

Thanks for taking this on.

FollowTheProcess:
I've removed the question form from this as we're planning to include a permalink to the nox discord channel as discussed in #534

FollowTheProcess:
@crwilcox I think a sensible thing to do with making things required might be to use these templates for a period of time and see if new issues are lacking important information, then we can make those fields required or adjust other aspects of the forms going forward?

PR #606, opened 2022-04-30
fix: GitHub Action can now be used on ubuntu-18.04 runner

mayeut:
Always install nox with python3.10 in order to allow using the action with an ubuntu-18.04 runner.


.github/workflows/ci.yml
```diff
@@ -26,6 +26,15 @@ jobs:
           python -m pip install --disable-pip-version-check .
       - name: Run tests on ${{ matrix.os }}
         run: nox --non-interactive --error-on-missing-interpreter --session "tests-${{ matrix.python-version }}" -- --full-trace
+  action-default-tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04, windows-2019, windows-2022, macos-10.15, macos-11, macos-12]
```
FollowTheProcess, line 33: Small nitpick but could you change e.g. `windows-2022` to `windows-latest`, likewise for `ubuntu-20.04` and `macos-11`. Just means less changes when things update

noxfile.py
```diff
@@ -105,3 +106,32 @@ def docs(session: nox.Session) -> None:
...
+            session.run(
+                "python",
+                "-c",
+                f"import sys; assert sys.implementation.name == '{implementation}'",
+            )
+
+    @nox.session(python=["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"])
+    def github_actions_default_tests(session: nox.Session) -> None:
+        """Check default versions installed by the nox GHA Action"""
+        _check_python_version(session)
```
FollowTheProcess, line 137: It's okay to define these in the normal scope, just have a check at the top level where we set `nox.options.sessions` to extend the list to include these when we're on GHA. Similar to how we decide whether to run the conda tests

FollowTheProcess (changes requested):

noxfile.py
mayeut: I think there was some misunderstanding here. Indeed the intended original behavior was the one from your edit.
I wrongly assumed you didn't want those sessions to even show up with `nox -l` when run locally.
Will revert this change.

FollowTheProcess: Yep that's my bad sorry! I forgot we set the default sessions anyway with `nox.options.sessions` just define these the normal way and then we can call them directly with the `--session` flag in the GHA jobs to test the action 👍🏻

.github/workflows/ci.yml
mayeut: The goal here is to test on every GitHub-hosted runner available.
Given that goal, I think it's best expressed using their qualified names. Yes there's a (small) price to pay when runners are added and/or removed but that's probably the only way to ensure the nox action is working correctly on all available runners.

PS: `macos-latest` is (for now) `macos-11` and it just seems strange to have `macos-latest` to reference `macos-11` when `macos-12` is also present. The same will happen with `ubuntu-22.04`/`ubuntu-20.04`/`ubuntu-latest` in a near future https://github.com/actions/virtual-environments/issues/5428

FollowTheProcess: Fair enough, I suppose we should be testing against every runner. FYI the `-latest` is the latest _stable_ release of the OS, so macos-12 (Monterey) is in public beta

mayeut: added ubuntu 22.04 that's now in public beta: https://github.com/actions/virtual-environments/issues/5490

FollowTheProcess:
It might be a good idea (given the number of jobs that will be run here) to put a condition in the CI workflow to only run the actions tests if `action.yml` has changed?

Just so every PR isn't clobbered with a load of actions tests if they've not changed it

mayeut:
> It might be a good idea (given the number of jobs that will be run here) to put a condition in the CI workflow to only run the actions tests if action.yml has changed?

That makes sense. I'd say only run on push to `main` or when there are changes in the action. 

I can either create a new workflow file to use built-in filtering as you mentioned or introduce a new workflow dependency https://github.com/dorny/paths-filter to keep only one workflow file.
Please let me know your preference.

FollowTheProcess:
> I'd say only run on push to main or when there are changes in the action.

Yep sounds good to me!

I think a separate workflow file makes the most sense to me, that way we have one for the main project and one to test the GitHub action

mayeut:
> I'd say only run on push to main or when there are changes in the action.

Well, GitHub doesn't allow to do an "OR filter" for push so I used `main` and `**action**` for branch name on push.

FollowTheProcess (approved):

FollowTheProcess:
Thanks @mayeut 🎉

PR #608, opened 2022-04-30
Fix discord channel name in config.yml

FollowTheProcess:
This PR fixes the Nox channel name in the `config.yml` which, as it's preceeded by a '#', yaml assumes it's a comment and does not display everything after it:

image


DiddiLeija (approved):
Oh man, I forgot Markdown was used here :man_facepalming:!

PR #609, opened 2022-04-30
feat: allow defining a python version list for GHA action (alt)

mayeut:
This is another way to implement the feature discussed in #601.

It hopefully addresses any shortcomings in the implementation except for the yaml list which is still not possible but has been replaced with a simple enough comma-separated list.

It now supports any python version supported by `actions/setup-python@v3` with up-to 20 versions installed.
The python validation/helper script which is currently inlined in the action might benefit from being a standalone script allowing for proper testing. I can do this if maintainers want to pursue this further.

One thing discovered while working on this is that mixing `actions/setup-python` & `wntrblm/nox` actions might not be that safe.
e.g. let's say that for some reason that I need `pypy-3.9-nightly` because of a bug in `pypy-3.9`:
```
    # this breaks cpython3.9 testing (overridden by pypy-3.9-nightly)
    - uses: wntrblm/nox
    - uses: actions/setup-python@v3
      with:
        python-version: "pypy-3.9-nightly"
```
```
    # this is just no-op (pypy-3.9-nightly overridden by wntrblm/nox)
    - uses: actions/setup-python@v3
      with:
        python-version: "pypy-3.9-nightly"
    - uses: wntrblm/nox
```
```
    # we currently need to do this
    - uses: wntrblm/nox
    - uses: actions/setup-python@v3
      with:
        python-version: "pypy-3.9-nightly"
    - uses: actions/setup-python@v3
      with:
        python-version: "3.9"
```
```
    # with this PR 
    - uses: wntrblm/nox
      with:
        python-versions: "3.7, 3.8, 3.9, 3.10, pypy-3.8, pypy-3.9-nightly"
```

closes #601

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -20,3 +20,11 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./
       - run: nox --non-interactive --error-on-missing-interpreter --session github_actions_default_tests
+  action-all-tests:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./
+        with:
+          python-versions: "2.7.18, 3.4, 3.5, 3.6, 3.7, 3.8, 3.9, 3.10, 3.11-dev, pypy-2.7, pypy-3.7, pypy-3.8, pypy-3.9-v7.3.9"
+      - run: nox --non-interactive --error-on-missing-interpreter --session github_actions_all_tests
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,12 +29,14 @@ repos:
     hooks:
       - id: isort
         args: ["-a", "from __future__ import annotations"]
+        exclude: ^.github/action_helper.py$
 
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.32.0
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
+        exclude: ^.github/action_helper.py$
 
   - repo: https://github.com/tox-dev/pyproject-fmt
     rev: "0.3.3"
--- a/action.yml
+++ b/action.yml
@@ -1,5 +1,10 @@
 name: Setup Nox
-description: 'Prepares all python versions for nox'
+description: "Prepares all python versions for nox"
+inputs:
+  python-versions:
+    description: "comma-separated list of python versions to install"
+    required: false
+    default: "3.7, 3.8, 3.9, 3.10, pypy-3.7, pypy-3.8, pypy-3.9"
 branding:
   icon: package
   color: blue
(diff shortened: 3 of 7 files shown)


action.yml
```diff
@@ -1,35 +1,134 @@
 name: Setup Nox
-description: 'Prepares all python versions for nox'
+description: "Prepares all python versions for nox"
+inputs:
+  python-versions:
+    description: "comma-separated list of python versions to install"
+    required: true
+    default: "3.7, 3.8, 3.9, 3.10, pypy-3.7, pypy-3.8, pypy-3.9"
```
FollowTheProcess, line 7: Should this be `required: false` if we're specifying a default? i.e. would a user _have_ to specify this or could they just do something like:

```yaml
- name: Setup Nox
  uses: wntrblm/nox
```

action.yml
```diff
@@ -1,35 +1,134 @@
...
 branding:
   icon: package
   color: blue
 
 runs:
   using: composite
   steps:
+    - name: "Validate input"
+      id: validate-input
+      run: |
```
FollowTheProcess, line 17: I think this would definitely benefit from being a standalone script, and I'd ideally like to see a decent number of test cases to ensure it handles weird edges, there's plenty of opportunity in here to fail in an ugly way (e.g. the indexing)

action.yml
```diff
@@ -1,35 +1,134 @@
...
+            raise ValueError(f"Too many interpreters to install: {len(versions)} > 15")
+
+        print(f"::set-output name=interpreter_count::{len(versions)}")
+        for i, version in enumerate(versions):
+            print(f"::set-output name=interpreter_{i}::{version}")
+      shell: "${{ runner.os == 'Windows' && 'python' || 'python3' }} {0}"
     - uses: actions/setup-python@v3
       with:
-        python-version: "pypy-3.7"
+        python-version: "${{ steps.validate-input.outputs.interpreter_0 }}"
```
FollowTheProcess, line 73: Apologies for my flip flopping on this but if this level of additional work is what's required to avoid the `"['3.7', '3.8', '3.9'...]"` syntax then I don't think it's worth it and I'd be happier to live with the string list. At least with the former we don't have to have a custom validator?

noxfile.py
```diff
@@ -108,3 +108,52 @@ def docs(session):
...
+    if python_version[:2] != "2.":
+        session.run(
+            "python",
+            "-c",
+            f"import sys; assert sys.implementation.name == '{implementation}'",
+        )
+
+
+@nox.session(python=["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"])
+def github_actions_default_tests(session):
```
FollowTheProcess, line 136: Could we have a switch in the `noxfile.py` that detects whether it's on GHA (e.g. the `CI` env var) and only includes these sessions in the selected list if we are on CI so that if people just run `nox` locally they don't have a load of cluttered skipped sessions?

We do something similar to skip conda if a user does not have it installed: https://github.com/wntrblm/nox/blob/2ad60058159e78e8a559e4f3a2b50451ea7c295a/noxfile.py#L27-L30

FollowTheProcess (changes requested):
See review comments. Am I correct in thinking that the additional validation logic here is purely a result of avoiding the `"['3.6', '3.7', '3.8'...]"` syntax?

If so I don't think the increased complexity is worth it and I'd actually be happier with a string list syntax. Sorry that I've flip flopped on this but I didn't appreciate how much additional complexity would be required

noxfile.py
mayeut: I started reworking this in #606 which should probably be reviewed first I guess.

mayeut:
thanks for the review, I'm a bit short on time until this week-end so I will address most remarks then.
In the meantime, #606 is probably easier to review and included here (https://github.com/wntrblm/nox/pull/609#discussion_r862752916 is applicable to #606 so maybe we should start there ?)

action.yml
mayeut: done

action.yml
mayeut: > I think this would definitely benefit from being a standalone script

done

> and I'd ideally like to see a decent number of test cases to ensure it handles weird edges, there's plenty of opportunity in here to fail in an ugly way (e.g. the indexing).

will do if we decide to go further on this.

> However if this is only required because of supporting the slightly nicer list syntax then I'd argue the better approach is to simply live with the string list syntax.

that's not for a slightly nicer list syntax, more on that later.

mayeut:
> Am I correct in thinking that the additional validation logic here is purely a result of avoiding the `"['3.6', '3.7', '3.8'...]"` syntax?
>
> If so I don't think the increased complexity is worth it and I'd actually be happier with a string list syntax. Sorry that I've flip flopped on this but I didn't appreciate how much additional complexity would be required

The validation/helper logic is not only there to avoid the `"['3.6', '3.7', '3.8'...]"` syntax. That's only a small aspect of what's being achieved in this PR.

The main goal with this alternative to the implementation of #601 is for the `wntrblm/nox` action to be able to install any python version that can be installed with `actions/python` correctly. The list of python version that can be installed with the action in #601 is hardcoded meaning it will require some updates every now & then (e.g. when python 3.11 is released or python 3.7 EOL, ...).

The implementation here comes with a small increase in complexity but is in my opinion more usable and I feel that there will be less maintenance on the action in the long run.

FollowTheProcess:
> The implementation here comes with a small increase in complexity but is in my opinion more usable and I feel that there will be less maintenance on the action in the long run.

So this one is intended to replace #601 ?

mayeut:
> So this one is intended to replace https://github.com/wntrblm/nox/pull/601 ?

short answer: yes.
longer answer: I'm keeping #601 opened for now if maintainers think this added complexity is really too much to handle. #601 might be an acceptable middle ground.

noxfile.py
```diff
@@ -132,4 +133,28 @@ def _check_python_version(session: nox.Session) -> None:
 @nox.session(python=["3.7", "3.8", "3.9", "3.10", "pypy3.7", "pypy3.8", "pypy3.9"])
 def github_actions_default_tests(session: nox.Session) -> None:
     """Check default versions installed by the nox GHA Action"""
+    assert sys.version_info[:2] == (3, 10)
```
FollowTheProcess, line 136: Will this not fail on every other python?

.pre-commit-config.yaml
```diff
@@ -29,12 +29,14 @@ repos:
     hooks:
       - id: isort
         args: ["-a", "from __future__ import annotations"]
+        exclude: ^.github/action_helper.py$
```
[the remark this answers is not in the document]
mayeut: This has to do with supporting ubuntu 18.04
While the core CPython team dropped support for 3.6, ubuntu 18.04 is still supported and includes security updates for CPython 3.6 which is the default python version installed on ubuntu 18.04.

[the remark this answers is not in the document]
mayeut: This has to do with supporting ubuntu 18.04
While the core CPython team dropped support for 3.6, ubuntu 18.04 is still supported and includes security updates for CPython 3.6 which is the default python version installed on ubuntu 18.04.

.pre-commit-config.yaml
```diff
@@ -29,12 +29,14 @@ repos:
...
 
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.32.0
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
+        exclude: ^.github/action_helper.py$
```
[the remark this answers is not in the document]
mayeut: This has to do with supporting ubuntu 18.04
While the core CPython team dropped support for 3.6, ubuntu 18.04 is still supported and includes security updates for CPython 3.6 which is the default python version installed on ubuntu 18.04.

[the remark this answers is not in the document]
mayeut: This has to do with supporting ubuntu 18.04
While the core CPython team dropped support for 3.6, ubuntu 18.04 is still supported and includes security updates for CPython 3.6 which is the default python version installed on ubuntu 18.04.

noxfile.py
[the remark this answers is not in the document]
mayeut: This is the python version used to install/run nox, not the session one.
I added a check in order to assert that nox is indeed installed using the expected python version.

.github/workflows/action.yml
```diff
@@ -20,3 +20,11 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./
       - run: nox --non-interactive --error-on-missing-interpreter --session github_actions_default_tests
+  action-all-tests:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./
+        with:
+          python-versions: "2.7.18, 3.4, 3.5, 3.6, 3.7, 3.8, 3.9, 3.10, 3.11-dev, pypy-2.7, pypy-3.7, pypy-3.8, pypy-3.9-v7.3.9"
```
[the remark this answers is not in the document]
mayeut: The fact that Nox dropped support for Python < 3.7 is one thing and that's completely fine however this only covers the Python version used to install & run Nox. If running a session with EOL/dev python is still supported (& it is as far as I know), then supporting those in the action & this test does still make sense IMHO.

This test is reduced to run only on `windows-latest` which allows to install all versions of python supported by `actions/setup-python` that Nox might get as a session python.

FollowTheProcess:
Sorry for the long delay, been a bit of a busy period! This is looking really good and thanks @mayeut for all the effort you've put in so far. It would be good to get a couple other maintainers to weigh in with any thoughts/comments (@wntrblm/nox) but this looks good to me 👍🏻

theacodes:
LGTM

mayeut:
Thanks for the feedback. I rebased on main to fix the conflict.

FollowTheProcess (approved):

FollowTheProcess:
Thanks @mayeut 🎉

PR #610, opened 2022-05-02
Prefer type checking against Jinja2

hauntsaninja:
As of version 3.0, Jinja2 includes type hints. These should be preferred to the (no longer updated) stub package from typeshed.


FollowTheProcess:
LGTM, thanks @hauntsaninja 🎉

FollowTheProcess (approved):

PR #616, opened 2022-05-02
build: Support PEP 621

### LINKED ISSUE #615: Support PEP 621?

With the acceptance of [PEP 621](https://peps.python.org/pep-0621/) and a [number of tools](https://scikit-hep.org/developer/pep621#simple-python-packaging) supporting it now, it might be a good idea to move Nox to the standard. The more high profile projects we have adopting it, (hopefully) the wider it will spread.

It will also enable us to switch build backends with relative ease should the need (or desire) ever arise.

Our current build backend (setuptools) has (currently experimental) support for PEP 621 from `setuptools>=61.0.0`.

Nox to adopt the PEP 621 standard and put project metadata in `pyproject.toml`.

Note that we'll still need either a `setup.py` or `setup.cfg` present in order to enable editable installs with setuptools.

I have a working implementation in my fork here: https://github.com/FollowTheProcess/nox/blob/build/pep-621/pyproject.toml that's passing the `check-manifest` pre-commit hook and I've verified the contents of the sdist after running `pyproject-build --sdist --wheel`.

Just stay as it is, but I think projects like Nox can be the leaders here and help kick the community at large into adoption 🙂 

### Anything else?

Edit: Thought I'd just confirm that doing this won't break things like pipx (it doesn't) and I actually discovered that pipx itself has adopted PEP 621! (using the hatchlings build backend): https://github.com/pypa/pipx/blob/main/pyproject.toml

FollowTheProcess:
This PR brings Nox in line with [PEP 621](https://peps.python.org/pep-0621/), storing project metadata in pyproject.toml.

Closes #615


DiddiLeija (approved):
This looks ready to go IMO!

PR #617, opened 2022-05-03
ci: Update GitHub actions to latest and add dependabot

FollowTheProcess:
This PR updates both `actions/checkout` and `actions/setup-python` to the latest major version (v3), and adds a GitHub built in dependabot config to ensure they are kept up to date.

Also changed `pypa/gh-action-pypi-publish@master` to `pypa/gh-action-pypi-publish@v1.5.0` as a best practice type thing, incase the master branch ever breaks, also allows dependabot to chip in with this one too 🙂

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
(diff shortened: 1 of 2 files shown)


DiddiLeija (approved):
There're no blockers for me :)

PR #618, opened 2022-05-05
chore: Apply consistent format to pyproject.toml

FollowTheProcess:
Now we have a PEP 621 pyproject.toml I was looking around for an equivalent to the excellent [setup-cfg-fmt](https://github.com/asottile/setup-cfg-fmt) and found [pyproject-fmt](https://github.com/tox-dev/pyproject-fmt) which seems to do a similar set of things, i.e. not only formatting but sorting keys and cleaning up certain things.

I figured since we were formatting `setup.cfg` in pre-commit runs we should do the same with `pyproject.toml` 👍🏻

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,11 @@ repos:
       - id: pyupgrade
         args: [--py37-plus]
 
+  - repo: https://github.com/tox-dev/pyproject-fmt
+    rev: "0.3.3"
+    hooks:
+      - id: pyproject-fmt
+
   - repo: https://github.com/hadialqattan/pycln
     rev: v1.3.2
     hooks:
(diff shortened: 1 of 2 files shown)


DiddiLeija (approved):

henryiii:
FYI, it's pretty early in the project, though a good idea. Mid to later year hopefully it will support the other features of setup-cfg-fmt, like the classifier validation and such. Maybe you'd have an opinion on https://github.com/tox-dev/pyproject-fmt/issues/22 as a neutral party. :D

henryiii:
FYI, it's pretty early in the project, though a good idea. Mid to later year hopefully it will support the other features of setup-cfg-fmt, like the classifier validation and such. Maybe you'd have an opinion on https://github.com/tox-dev/pyproject-fmt/issues/22 as a neutral party. :D

FollowTheProcess:
> Maybe you'd have an opinion on tox-dev/pyproject-fmt#22 as a neutral party. :D

I actually prefer the way they've done it 👀 I like that it reads as "this project uses this build system, and that requires these things"

PR #619, opened 2022-05-06
fix: PEP 621 license field

FollowTheProcess:
I've discovered (by accident) something quite unfortunate about the PEP 621 pyproject.toml `license` field. I recently ported [pytoil](https://github.com/FollowTheProcess/pytoil) to support PEP 621 and released a new version with the following config:

```toml
license = {file = "LICENSE"}
```

However this gives a rather ugly representation of the entire file on PyPI:

image

This can be fixed by simply passing the name of the license in the `text` key like so:

```toml
license = {text = "Apache-2.0"}
```

Which results in what you'd expect:

image

And I can confirm that the full LICENSE file is still included when running `pyproject-build`:

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ version = "2022.1.7"
 description = "Flexible test automation."
 readme = "README.md"
 requires-python = ">=3.7"
-license = { file = "LICENSE" }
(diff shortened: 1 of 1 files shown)


DiddiLeija (approved):
Good catch!

PR #620, opened 2022-05-07
docs: remove marketplace link for `wntrblm/nox`

mayeut:
Per https://github.com/wntrblm/nox/pull/594#discussion_r859472266, `wntrblm/nox` will not be published on the marketplace.


FollowTheProcess (approved):
Good catch

PR #627, opened 2022-06-15
feat: Add session tags

### LINKED ISSUE #339: Add functionality similar to pytest.mark
It can simplify filtering specific sessions to be run. It can simplify writing long commands like `nox -s tests, lint, type, ...`. It can also extend the functionality of `nox.options.sessions`.

I have two ideas how this can be done.
Basically it can be run by: 
`nox -m [mark-name]` if feature called "Marks" as in pytest,
`nox -t [tag-name]` if feature called "Tags".
And nox will run only sessions annotated by this mark/tag.

**1st option**
In `noxfile.py` it can be configured either by:
```python
@nox.mark("mark_name")
@nox.session()
def ...
``` 
or
 ```python
@nox.tag("tag_name")
@nox.tag("other_tag_name")
@nox.session()
def ...
```

**2nd option**
```python
nox.options.tags = {
    "default": ["session1", "session2"],  # this might replace `nox.options.sessions`
    "tag1": ["session1", "session4"],
    "tag2": ["session2", "session3"],
    etc.
}
```

Personally I prefer the 2nd option, I think it makes it more clear and concise.
And also I prefer "Tags" over "Marks" name, IMO "Tags" more precisely describe what it is. On the other hand, "Marks" might be known for those who use pytest.

edgarrmondragon:
Closes #339

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -432,6 +432,58 @@ read more about parametrization and see more examples over at
 .. _pytest's parametrize: https://pytest.org/latest/parametrize.html#_pytest.python.Metafunc.parametrize
 
 
+Session tags
+------------
+
+You can add tags to your sessions to help you organize your test suite:
+
+.. code-block:: python
+
+    @nox.session(tags=["style", "fix"])
+    def black(session):
+        session.install("black")
+        session.run("black", "my_package")
+
+    @nox.session(tags=["style", "fix"])
+    def isort(session):
+        session.install("isort")
+        session.run("isort", "my_package")
+
+    @nox.session(tags=["style"])
+    def flake8(session):
+        session.install("flake8")
+        session.run("flake8", "my_package")
+
+
+If you run ``nox -t style``, Nox will run all three sessions:
+
+.. code-block:: console
+
+    * black
+    * isort
+    * flake8
+
+
+If you run ``nox -t fix``, Nox will only run the ``black`` and ``isort``
+sessions:
+
+.. code-block:: console
+
+    * black
+    * isort
+    - flake8
+
+
+If you run ``nox -t style fix``, Nox will all sessions that match *any* of
+the tags, so all three sessions:
+
+.. code-block:: console
+
+    * black
+    * isort
+    * flake8
+
+
 Next steps
 ----------
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -435,7 +435,7 @@ read more about parametrization and see more examples over at
 Session tags
 ------------
 
-You can add tags to your sessions to help you organize your test suite:
+You can add tags to your sessions to help you organize your development tasks:
 
 .. code-block:: python
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -74,12 +74,15 @@ If you have a :ref:`configured session's virtualenv <virtualenv config>`, you ca
     nox --python 3.8
     nox -p 3.7 3.8
 
-You can also use `pytest-style keywords`_ to filter test sessions:
+You can also use `pytest-style keywords`_ using ``-k`` or ``--keywords``, and
+tags using ``-t`` or ``--tags`` to filter test sessions:
 
 .. code-block:: console
 
     nox -k "not lint"
     nox -k "tests and not lint"
+    nox -k "not my_tag"
+    nox -t "my_tag" "my_other_tag"
 
 .. _pytest-style keywords: https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests
 
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -178,7 +178,7 @@ class Manifest:
         ]
 
     def filter_by_tags(self, tags: list[str]) -> None:
-        """Filter sessions using pytest-like tag expressions.
+        """Filter sessions by their tags.
 
         Args:
             tags (list[str]): A list of tags which session names
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -177,7 +177,7 @@ class Manifest:
             if keyword_match(keywords, x.signatures + x.tags + [x.name])
         ]
 
-    def filter_by_tags(self, tags: str) -> None:
+    def filter_by_tags(self, tags: list[str]) -> None:
         """Filter sessions using pytest-like tag expressions.
 
         Args:
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -177,11 +177,11 @@ class Manifest:
             if keyword_match(keywords, x.signatures + x.tags + [x.name])
         ]
 
-    def filter_by_tags(self, tags: list[str]) -> None:
+    def filter_by_tags(self, tags: Sequence[str]) -> None:
         """Filter sessions by their tags.
 
         Args:
-            tags (list[str]): A list of tags which session names
+            tags (Sequence[str]): A list of tags which session names
                 are checked against.
         """
         self._queue = [x for x in self._queue if set(x.tags).intersection(tags)]
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -181,11 +181,8 @@ class Manifest:
         """Filter sessions using pytest-like tag expressions.
 
         Args:
-            tags (str): A Python expression of tags which session names
+            tags (list[str]): A list of tags which session names
                 are checked against.
-
-        Raises:
-            SyntaxError: If the tag expression is invalid.
         """
         self._queue = [x for x in self._queue if set(x.tags).intersection(tags)]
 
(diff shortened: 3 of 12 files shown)


DiddiLeija:
Looks like a nice start. I haven't verified the behavior yet, but I think it's what we want (any thoughts @wntrblm/nox?). Also, I approved the CI runs to discard any failure.

FollowTheProcess:
This is looking good. Although I've checked out your PR and can't seem to get the desired behaviour:

```python
# noxfile.py

@nox.session(tags=["hello"])
def hello(session: nox.Session) -> None:
    """
    Does something.
    """
    session.log("hello")

```

When I run `nox -t hello` I get the following:

image

So not sure what's going on there?

It would also be really good to incorporate the tags somehow into the output of `nox --list`, although I think that would be better in a fresh PR, let's get the functionality merged first then work on how we show it

nox/manifest.py
```diff
@@ -172,9 +172,23 @@ def filter_by_keywords(self, keywords: str) -> None:
...
+
+        Args:
+            tags (str): A Python expression of tags which session names
+                are checked against.
+
+        Raises:
+            SyntaxError: If the tag expression is invalid.
+        """
+        self._queue = [x for x in self._queue if set(x.tags).intersection(tags)]
+
```
FollowTheProcess, line 191: Is tags really just a string here? Everywhere else it seems to be `list[str]`? If it is a string, the expression `set(x.tags).intersection(<a_string>)` will always evaluate to an empty set (falsy):

image


Also AFAICT this would not raise a SyntaxError?

edgarrmondragon:
> ```python
> # noxfile.py
> 
> @nox.session(tags=["hello"])
> def hello(session: nox.Session) -> None:
>     """
>     Does something.
>     """
>     session.log("hello")
> ```

@FollowTheProcess Thanks for reviewing. I tried it with your example and it's working as expected:

```console
$ nox -t hello
nox > Running session hello
nox > Creating virtual environment (virtualenv) using python3.9 in .nox/hello
nox > hello
nox > Session hello was successful.

$ nox -k "hello or other"
nox > Running session hello
nox > Creating virtual environment (virtualenv) using python3.9 in .nox/hello
nox > hello
nox > Session hello was successful.

$ noxpy -k "not hello"
nox > No sessions selected after filtering by keyword.
```

nox/manifest.py
edgarrmondragon: Ah, good catch! Yeah, this type annotation should be `list[str]`. I guess `mypy` didn't catch it because `Namespace.tags` doesn't have a static type.

FollowTheProcess (changes requested):
Weirdly it worked fine for me after I removed and re-cloned the repo 🤷🏻 so obviously something to do with my laptop and not the code so we're all good there.

Just a few minor changes, otherwise this is looking great

docs/usage.rst
```diff
@@ -74,12 +74,15 @@ If you have a :ref:`configured session's virtualenv <virtualenv config>`, you ca
...
-You can also use `pytest-style keywords`_ to filter test sessions:
+You can also use `pytest-style keywords`_ using ``-k`` or ``--keywords``, and
+tags using ``-t`` or ``--tags`` to filter test sessions:
 
 .. code-block:: console
 
     nox -k "not lint"
     nox -k "tests and not lint"
+    nox -k "not my_tag"
+    nox -t "my_tag" "my_other_tag"
```
crwilcox, line 85: will -k match tags as well then?

nox/manifest.py
```diff
@@ -172,9 +172,23 @@ def filter_by_keywords(self, keywords: str) -> None:
...
         """
         self._queue = [
-            x for x in self._queue if keyword_match(keywords, x.signatures + [x.name])
+            x
+            for x in self._queue
+            if keyword_match(keywords, x.signatures + x.tags + [x.name])
         ]
 
+    def filter_by_tags(self, tags: list[str]) -> None:
+        """Filter sessions using pytest-like tag expressions.
```
crwilcox, line 181: I wonder if this was refactored at a point. This comment makes this all seem a bit more powerful than it seems to be.

1) What is pytest like about this? As best I can tell tags don't support additional keywords and just checks whether or not any of the specified tags are present?

2) When I read 'Python expression' I start to think some sort of eval() is going on, but we are just comparing strings right?

tests/test_tasks.py
```diff
@@ -243,6 +244,66 @@ def test_filter_manifest_keywords_syntax_error():
     assert return_value == 3
 
 
+@pytest.mark.parametrize(
+    "tags,session_count",
```
crwilcox, line 248: are tags case sensitive? I wonder if adding a case here for that would make sense?

```
        (["Foo"], 0),
```

similarly, most of the tags match function names. if we have the tag `foo_bar_baz` it should also match 0

nox/manifest.py
```diff
@@ -172,9 +172,23 @@ def filter_by_keywords(self, keywords: str) -> None:
...
+
+        Args:
+            tags (str): A Python expression of tags which session names
+                are checked against.
+
+        Raises:
+            SyntaxError: If the tag expression is invalid.
+        """
```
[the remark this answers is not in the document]
edgarrmondragon: Ah, yes! I think I let copilot auto-generate this docstring. Big mistake 😬

nox/manifest.py
edgarrmondragon, line 181:
```suggestion
        """Filter sessions by their tags.
```

edgarrmondragon: That's correct. This filter just matches by the raw tags. It's the keywords filter that supports tag expressions.

docs/usage.rst
edgarrmondragon: @crwilcox that's correct, tags can be used in expressions too :)

docs/tutorial.rst
```diff
@@ -432,6 +432,58 @@ read more about parametrization and see more examples over at
 .. _pytest's parametrize: https://pytest.org/latest/parametrize.html#_pytest.python.Metafunc.parametrize
 
 
+Session tags
+------------
+
+You can add tags to your sessions to help you organize your test suite:
```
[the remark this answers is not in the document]
edgarrmondragon: @FollowTheProcess Done!

tests/test_tasks.py
edgarrmondragon: @crwilcox thanks for the suggestions, I added test cases for tags not found in `test_filter_manifest_tags_not_found`. I also just changed the function names so make sure we're matching tags and not session names. Let me know what you think.

nox/manifest.py
FollowTheProcess: In that case I think we should change the docstring to reflect that e.g. `"Filter sessions on matching tags"`

edgarrmondragon: @FollowTheProcess yup, I'd changed it to `"Filter sessions by their tags"`. Let me know if that's more like it.

FollowTheProcess: That's great, thanks

nox/manifest.py
```diff
@@ -172,9 +172,20 @@ def filter_by_keywords(self, keywords: str) -> None:
                 session names are checked against.
         """
         self._queue = [
-            x for x in self._queue if keyword_match(keywords, x.signatures + [x.name])
+            x
+            for x in self._queue
+            if keyword_match(keywords, x.signatures + x.tags + [x.name])
         ]
 
+    def filter_by_tags(self, tags: Sequence[str]) -> None:
```
FollowTheProcess, line 180: Any particular reason for using `Sequence[str]` here? Everywhere else it's a `list[str]`, if we can be more specific and use `list` I don't see the benefit of using the more generic `Sequence` IMO

FollowTheProcess (changes requested):

nox/manifest.py
edgarrmondragon: @FollowTheProcess I decided to use the same signature as [`Manifest.filter_by_python_interpreter`](https://github.com/wntrblm/nox/blob/d1bdf0940eda7213dc2bd6407cf7107dfc986672/nox/manifest.py#L157), and it also made sense to follow the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). Wdyt?

FollowTheProcess: I'd say just stick with a `list[str]` personally, it's a `list[str]` everywhere else it shows up it makes more sense to keep it consistent IMO rather than have the one off `Sequence` here. As for the robustness thing, that doesn't matter so much here as it's internal stuff we can change the signature of without breaking people

edgarrmondragon: @FollowTheProcess Sounds good to me. I've pushed the change.

FollowTheProcess (approved):
LGTM, thanks @edgarrmondragon  🎉

PR #628, opened 2022-06-20
feat: expose no_install and install_only as public API

kngwyu:
Just an idea, but I feel it convenient if they are exposed as public API.
I'm using [pip-sync](https://github.com/jazzband/pip-tools/) to install dependencies, so I want to re-use `--no-install` flag for disabling running `pip-sync`. I sometimes pass `posargs` to external tools like pytest, so I want to avoid `nox -s ~ -- --no-sync`.

I feel that some other people may find them useful, but I don't have any other use case. Suggestions are welcome 🙂

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -162,6 +162,22 @@ class Session:
         return self._runner.posargs
 
     @property
+    def no_install(self) -> bool:
+        """
+        Whether ``--no-install`` is passed to the ``nox`` command line.
+        This property may be useful if a user is using their own installing
+        command (e.g., ``pip-sync``).
+        """
+        return self._runner.global_config.no_install
+
+    @property
+    def install_only(self) -> bool:
+        """
+        Whether ``--install-only`` is passed to the ``nox`` command line.
+        """
+        return self._runner.global_config.install_only
+
+    @property
     def virtualenv(self) -> ProcessEnv:
         """The virtualenv that all commands are run in."""
         venv = self._runner.venv
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -23,7 +23,7 @@ import re
 import sys
 import unicodedata
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


kngwyu:
I often use `-r` and `--no-install` simultaneously to speed up tests. Also, my point is to keep the behavior of `session.install` and `session.run("pip-sync")` consistent when `--no-install` or `--install-only` is given.

FollowTheProcess:
> I often use -r and --no-install simultaneously to speed up tests.

In case you didn't know, you can use `-R` as a shortcut for this combo (see [here](https://nox.thea.codes/en/stable/usage.html?highlight=reuse#re-using-virtualenvs))

What's your workflow where you need this feature? Could you give me a little example of what you'd do with these new attributes I cant quite picture why you'd need it in my head

kngwyu:
@FollowTheProcess 

Like this snippet:
```python
if not session.no_install or session.install_only:
    session.run("pip-sync", requirements)
```

FollowTheProcess:
Ahh okay that makes sense, in that case I'm happy with this. Just need to get these covered in a test to make the CI green and we should be okay 👍🏻

kngwyu:
Thanks, I just added some tests to make coverage happy

cjolowicz:
This looks like the use case for session.run_always, or am I missing something? That function won't run under --no-install.

kngwyu:
> This looks like the use case for session.run_always, or am I missing something? That function won't run under --no-install.

Ah, you're right, thank you. I feel like it'd be a better solution than manually writing if sentences. I still think that it's OK to expose these variables since they are specified by users, but now I don't have strong motivation to merge this PR 😅

FollowTheProcess:
Ahh nice! is this still required then @kngwyu ?

kngwyu:
I'm sorry but I don't have clear motivation now

PR #629, opened 2022-07-11
fix: use setup-python v4.1's environmentless setup

henryiii:
This removes the dependence on the final output being reasonable to run nox. Targeting only <3.7 is valid now.

CC @mayeut

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -12,12 +12,12 @@ branding:
 runs:
   using: composite
   steps:
-    - uses: actions/setup-python@v4
+    - uses: actions/setup-python@v4.1.0
       id: localpython
       with:
         python-version: "3.7-3.10"
         activate-environment: false
-        
+
     - name: "Validate input"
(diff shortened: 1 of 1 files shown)


henryiii:
`v4` has been updated.

FollowTheProcess (approved):
LGTM @henryiii thanks!

DiddiLeija (approved):
Thanks @henryiii!

mayeut:
Sorry I did not have time to review this until today.
I opened #632 to fix the environmentless setup of the `localpython` step, [GHA only warns on invalid input names](https://github.com/actions/runner/issues/933).

One thing worth mentioning is that with this, the nox action won't honor the python version specified by the user as an input of the action for the `major.minor` version used to install nox, it will just use the `localpython` one. probably 99.9% of the times, they will be the same so that's not really an issue in the real world but at least worth mentioning.

PR #632, opened 2022-07-14
fix: invalid input in action's localpython step

mayeut:
GHA does not error out on invalid input name but instead only shows a warning which is why this did go unnoticed.


FollowTheProcess (approved):
Nice, good catch!

PR #639, opened 2022-07-28
Add short -N alias for --no-reuse-existing-virtualenvs

### LINKED ISSUE #624: Add a single character shortcut for --no-reuse-existing-virtualenvs

A common way to use Nox that saves a lot of time when re-running sessions over and over by avoiding recreating venvs is to set:

```py
nox.options.reuse_existing_virtualenvs = True
```

You can see the `pip` developers use this:

- https://github.com/pypa/pip/blob/main/noxfile.py#L19

When dependencies _do_ change, the developer can then run:

```
nox --no-reuse-existing-virtualenvs -s test
```

Given this can be a very common work flow, avoiding having to type out `--no-reuse-existing-virtualenvs` on the command line would be a great improvement to UX.

### A note on `tox`

What I'm describing above is also the default workflow for `tox` users. Many people moving from `tox` will be familiar with this workflow. (e.g. `tox` re-uses venvs by default, and `tox -r` is an opt-in recreation).

Add a new alias for `--no-reuse-existing-virtualenvs`, for example `-C` for "Create"

```
nox -C -s test
```

johnthagen:
Closes #624

### DIFF
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -160,7 +160,7 @@ def flag_pair_merge_func(
 def make_flag_pair(
     name: str,
     enable_flags: tuple[str, str] | tuple[str],
-    disable_flags: tuple[str],
+    disable_flags: tuple[str, str] | tuple[str],
     default: bool = False,
     **kwargs: Any,
 ) -> tuple[Option, Option]:
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -363,7 +363,10 @@ options.add_options(
     *_option_set.make_flag_pair(
         "reuse_existing_virtualenvs",
         ("-r", "--reuse-existing-virtualenvs"),
-        ("-nr", "--no-reuse-existing-virtualenvs",),
+        (
+            "-nr",
+            "--no-reuse-existing-virtualenvs",
+        ),
         group=options.groups["environment"],
         help="Re-use existing virtualenvs instead of recreating them.",
     ),
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -364,7 +364,7 @@ options.add_options(
         "reuse_existing_virtualenvs",
         ("-r", "--reuse-existing-virtualenvs"),
         (
-            "-R",
+            "-N",
             "--no-reuse-existing-virtualenvs",
         ),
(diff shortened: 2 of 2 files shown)


johnthagen:
```
$ python nox --help
...
  -nr, --no-reuse-existing-virtualenvs
                        Disables --reuse-existing-virtualenvs if it is enabled in the Noxfile.
```

johnthagen:
@DiddiLeija @crwilcox @ktbarrett Could you review?

Specifically, are we sure we want to set a precedent of using a multi-letter short option (`-nr`) vs picking another single letter short option (such as `-C`)?

ktbarrett:
Looks fine to me, but I'm no maintainer. 

>are we sure we want to set a precedent of using a multi-letter short option

The precedent is already set by a number of other options. Scrolling up a few lines shows `-fb`, `-db`, and a couple others.

johnthagen:
> The precedent is already set by a number of other options. Scrolling up a few lines shows `-fb`, `-db`, and a couple others.

That is a good point and something I missed.

FollowTheProcess:
Is `-N` available? That might be better than `-nr`?

johnthagen:
It appears that `-N` is available. I'm happy to change to that if that is the consensus of the Nox team.

DiddiLeija:
No objections from me :)

FollowTheProcess:
> It appears that -N is available. I'm happy to change to that if that is the consensus of the Nox team.

I'm easy either way really it's not a dealbreaker. I just don't like the look of "short" options with more than 1 character (I know we already have some) We have a precedent for capitals with `-R` for reuse and no install (from memory) so `-N` would fit better IMO

johnthagen:
Sounds good. I've changed to `-N`. I also prefer this way.

When merging, if someone could hit the "Squash" button in the GitHub UI, I'd appreciate it.

FollowTheProcess:
> When merging, if someone could hit the "Squash" button in the GitHub UI, I'd appreciate it.

We only allow squashes anyway 👍🏻

FollowTheProcess (approved):
LGTM, nice job thanks!

PR #641, opened 2022-08-02
Export session name in envvar NOX_CURRENT_SESSION

### LINKED ISSUE #636: Set environment variable when running commands

Setting an environment variable in the environment of `session.run`'d commands would allow the child process to know it is being executed by nox and perhaps even what session of nox is being. Personally, I would find this useful when `session.run`ing `bash` or `code` (using nox as a development environment manager) so that I can adjust my PS1 with the name of the session I'm in.

When running any command with `session.run`, the command environment is adjusted to set the `NOX_CURRENT_SESSION` variable with the name of the current session.

It's entirely possible to do this myself using `env` argument to `session.run`. However it is something innocuous that takes the burden off the user without considerable maintenance effort.

### Anything else?

_No response_

ktbarrett:
Closes #636. The environment variable `NOX_CURRENT_SESSION` is populated with the name of the current session in all subcommands of that session. 

There were a couple places to place this small bit of code. Each location caused some problems with the existing tests. The current place was chosen because it intuitively makes sense to take code that the user would write at the top of their session functions and move it up into the thing that calls the session function. Let me know if you think it belongs elsewhere.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -724,6 +724,7 @@ class SessionRunner:
             with cwd:
                 self._create_venv()
                 session = Session(self)
+                session.env["NOX_CURRENT_SESSION"] = session.name
                 self.func(session)
 
             # Nothing went wrong; return a success.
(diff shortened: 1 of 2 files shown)


FollowTheProcess (approved):
LGTM, nice and clean implementation, great work!

DiddiLeija (approved):
Thanks @ktbarrett!

PR #645, opened 2022-08-16
fix: Include action helper in MANIFEST.in

### LINKED ISSUE #644: Missing action_helper in PyPI tarball
### Current Behavior

In PyPI tarball, there is `test_action_helper.py` module but the module this one should test (`.github/action_helper.py`) is not there.

### Expected Behavior

I think that the tarball should either contain both files or neither of them.

### Steps To Reproduce

1. Download and unpack the tarball.
2. Run tests.

### Environment

```Markdown
Does not matter
```

### Anything else?

_No response_

FollowTheProcess:
Closes #644 

Added `.github/action_helper.py` to manifest file.

I unpacked the tarball generated by [pypa/build](https://github.com/pypa/build) and you can see that it gets correctly included:

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
         args: [--py37-plus]
 
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: "0.3.3"
+    rev: "0.3.5"
     hooks:
       - id: pyproject-fmt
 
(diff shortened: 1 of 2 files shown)


FollowTheProcess:
Hmm, looks like an issue with pyproject-fmt, I think we're a bit out of date I'll update that to fix this PR run and then probably do another PR updating all the hooks.

image

PR #649, opened 2022-09-04
fix: nox.session.run-ing commands with pathlib.Path arguments

Julian:
Everything previously really worked fine in theory, but the debug string
which is logged (and calls shlex.quote) was blowing up trying
to quote a pathlib.Path object. I'm somewhat surprised upstream
shlex.quote doesn't support pathlib.Path objects (yet?) but for
now this just falls back to a cruder representation.

(Obviously there are other ways to handle this, like calling str() on all the args, let me know if you prefer something else).

This also apparently doesn't work on Windows, but that's because `subprocess.Popen` is apparently broken with WindowsPaths. The upstream bug is  https://github.com/python/cpython/issues/85815.

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -71,8 +71,7 @@ def _clean_env(env: dict[str, str] | None) -> dict[str, str] | None:
 
 
 def _shlex_join(args: Sequence[str]) -> str:
-    # shlex.join() was added in Python 3.8
-    return " ".join(shlex.quote(arg) for arg in args)
+    return " ".join(shlex.quote(os.fspath(arg)) for arg in args)
 
 
 def run(
@@ -92,10 +91,7 @@ def run(
         success_codes = [0]
 
     cmd, args = args[0], args[1:]
-    try:
-        full_cmd = f"{cmd} {_shlex_join(args)}"
-    except TypeError:
-        full_cmd = f"{cmd} {args}"
+    full_cmd = f"{cmd} {_shlex_join(args)}"
 
     cmd_path = which(cmd, paths)
 
--- a/nox/command.py
+++ b/nox/command.py
@@ -92,7 +92,10 @@ def run(
         success_codes = [0]
 
     cmd, args = args[0], args[1:]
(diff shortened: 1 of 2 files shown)


henryiii:
I'd think `os.fspath(x) if isinstance(x, os.PathLike) for x in args` would be the right way to handle it (without introducing bugs if something non-pathlike gets passed in that has a str representation - which everything in Python does).

Julian:
Happy to do that instead obviously -- do you mean just what I just pushed (I don't think one needs to do the isinstance check, unless you meant to filter out non-fspath args silently, which seems unlikely)?

henryiii:
Actually, I guess it doesn't need a check, since `os.fspath` passes through `str` / `bytes` and only works on `os.PathLike` otherwise. It just seems a little odd to pass a non-path into `os.fspath`, and there will be non-paths coming in. It also might be a bit more extendable - if any other object protocols were ever added, you'd then need the check.

henryiii:
FYI, I recently wrote a class that wrapped a commandline tool, and realized I could define `__fspath__()` instead of making up an arbitrary attribute holding the path to the wrapped exe. Things like this would cause it to just work as if it was the wrapped exe (this is exactly what I did with my own run function). Even without it, `os.fspath(thing)` is easier to remember than `thing.some_name_I_made_up`.

But the important thing is the `str` is different than `__fspath__()`, and the fspath one is the correct one to use.

Julian:
Right I've made CLIs that worked precisely as you describe (where they allow lots of fun objects to become filepaths). Certainly supporting *more* objects would indeed be nice.

Julian:
Anything else you'd like to see changed from this one?

FollowTheProcess (approved):
I was initially concerned about using `os.fspath` on any argument, whether or not it is "PathLike". But it turns out that if `os.fspath` gets passed `str` or `bytes` they are returned unchanged so no worries there! Looks good to me @Julian thanks!

Link to `os.fspath`: https://docs.python.org/3/library/os.html#os.fspath

Julian:
Thanks!

Issue #218, opened 2019-07-18
Support stdout file redirects in session.run()


schuderer:
**How would this feature be useful?**
When building projects, I find myself having to use commands which generate output to stdout (such as `pipenv lock -r`) which I want to redirect to a file (e.g. `>docs/requirements.txt`). This currently is not possible with `session.run()`, because, as I understand it, special characters such as `>` are being escaped before they are run.

**Describe the solution you'd like**
It would be handy if `session.run()` would accept something like `session.run("echo", "hello", ">hello.txt")` or `session.run("echo", "hello", ">", "hello.txt")` and pass the redirect through so it gets executed when running the command.

**Describe alternatives you've considered**
I am using `Makefile`s and, for cross-platform reasons, `make.bat`. I also played around with task-specific shell/batch files. Having `nox` at hand, I am hoping I can centralize all my test-, but also more general build-related processes into one neat and tidy script, replacing non-Windows-compatible `Makefile`s and reducing clutter.

theacodes:
Hi, we added this in #180. You can do

```python
with open("pipenv-log.txt", "w") as out:
    session.run("pipenv", "lock", "-r", stdout=out)
```

henryiii:
Maybe dumb question: is there a way to get the output directly in a variable, without a file? I don't seem to be get a StringIO/BytesIO to work here, or a `redirect_stdout`. I get `io.UnsupportedOperation: fileno`.

henryiii:
Also, I don't think this was ever documented?

henryiii:
Ahh, output is returned if this is silent. I thought I'd done this somewhere before. (Ironically, probably for something very similar to what I'm doing now, grabbing a version).

PR #651, opened 2022-09-13
Add docs for #180

hynek:
I found this extremely useful feature only by accident and Googling!

Not sure if y'all enjoy retroactive changelog editing…

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -305,18 +305,15 @@ class Session:
                # Display coverage report even when tests fail.
                session.run("coverage", "report")
 
-        You can redirect the output of a command into an open file. For example to get the
-        current Git commit ID::
+        If you pass ``silent=True``, you can capture the output of a command that would
+        otherwise be shown to the user. For example to get the current Git commit ID::
 
-            with tempfile.TemporaryFile() as out:
-                session.run(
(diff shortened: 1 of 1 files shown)


theacodes:
@hynek thanks for sending this!

There's actually an even easier way to accomplish the example shown:

```python
commit = session.run("git", "rev-parse", "--short", "HEAD", external=True, silent=True)
```

hynek:
aaargh now i understand [Henry's comment](https://github.com/wntrblm/nox/issues/218#issuecomment-1005178204). I never read the full help string b/c i didn't want silence.

I guess i'll just rewrite example (but leave the new API docs)? 😅

theacodes:
Yeah that would be awesome. It's still a useful thing to document, I'd just
want it to have a more practical example.

DiddiLeija:
Yeah, it's worthy to document this. The CI failed, so I decided to re-run the failed checks.

hynek:
I've rewritten the example and clarified the arguments of stdout/stderr. I don't think they need their own examples; with silent=True they're rather fringe options.

theacodes (approved):

theacodes:
Thank you so much, @hynek!

PR #652, opened 2022-09-18
Enhance environment variables passing

### LINKED ISSUE #253: Unexpected (buggy?) environment passing
**Describe the bug**
When I do `session.run("stuff", env={})`, I would expect to run the command in an environment where no variable is defined, however a copy of the local environment is made. Similarly, if I do `session.run("stuff", env={"CREEPY": "TRUE"})` I would expect to run in an environment where only `$CREEPY` is defined, while this only updates the copy of the current environment.

**How to reproduce**
Check the environment when running with `env={}`.

**Expected behavior**
If I pass `env={}`, I would expect to work in a clean environment. At the moment, the workaround is to do for instance `session.virtualenv.env = {"SYSTEMROOT": os.getenv("SYSTEMROOT", "")}` at the beginning of each session, but I find that a bit too cumbersome.

franekmagiera:
Closes #253 

Introduced a new parameter `include_invocation_env_vars` to the `session.run` related methods. It can be a boolean or an iterable of vars that should be included from the nox invocation environment. Some env vars (inspired by [tox](https://tox.wiki/en/latest/config.html#conf-passenv) are always passed in to ensure proper functionality of stdlib functions and pip.

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -52,16 +52,16 @@ def which(program: str, paths: list[str] | None) -> str:
     raise CommandFailed(f"Program {program} not found")
 
 
-def _clean_env(env: dict[str, str] | None) -> dict[str, str] | None:
-    if env is None:
-        return None
-
+def _clean_env(env: dict[str, str] | None) -> dict[str, str]:
     clean_env = {}
 
     # Ensure systemroot is passed down, otherwise Windows will explode.
-    clean_env["SYSTEMROOT"] = os.environ.get("SYSTEMROOT", "")
+    if sys.platform == "win32":
+        clean_env["SYSTEMROOT"] = os.environ.get("SYSTEMROOT", "")
 
-    clean_env.update(env)
+    if env is not None:
+        clean_env.update(env)
+
     return clean_env
 
 
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -272,7 +272,11 @@ class Session:
             raise nox.command.CommandFailed() from e
 
     def run(
-        self, *args: str, env: Mapping[str, str] | None = None, **kwargs: Any
+        self,
+        *args: str,
+        env: dict[str, str] | None = None,
+        include_outer_env: bool = True,
+        **kwargs: Any,
     ) -> Any | None:
         """Run a command.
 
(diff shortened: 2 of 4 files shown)


franekmagiera:
I think this is ready for review.

nox/command.py
```diff
@@ -57,16 +101,16 @@ def which(program: str, paths: list[str] | None) -> str:
...
-    if env is None:
-        return None
-
+def _clean_env(env: dict[str, str] | None) -> dict[str, str]:
     clean_env = {}
 
-    # Ensure systemroot is passed down, otherwise Windows will explode.
-    clean_env["SYSTEMROOT"] = os.environ.get("SYSTEMROOT", "")
+    for var in _ENV_VARS_TO_ALWAYS_INCLUDE:
+        if var in os.environ:
```
crwilcox, line 108: I think this has the potential to be a race as written (possible the environment var is there when checked on the first check, but not on assignment).

nox/sessions.py
```diff
@@ -391,23 +416,44 @@ def run_always(
...
+            include_invocation_env_vars=include_invocation_env_vars,
+            **kwargs,
+        )
 
     def _run(
-        self, *args: str, env: Mapping[str, str] | None = None, **kwargs: Any
+        self,
+        *args: str,
+        env: dict[str, str] | None = None,
+        include_invocation_env_vars: Iterable[str] | bool = True,
```
crwilcox, line 430: I think this approach is worth discussing. In a way it is a subtle improvement over the suggestion in the bug thread (where passing an env var name with None as value meant "use the system". I also think it may more clear as it keeps this to just keys.

That said, I am not a huge fan of type of this var and that, in a way, it ends up overriding truthiness of the dict.

I think having the on/off makes a lot of sense, as this allows users of nox to control their sessions fully. If most folks are going to use this to bring 1-2 vars, it doesn't seem too overwhelming to need to do it via env:

```
result = session.run(
            ...
            env={"SYSTEM_ROOT": os.environ["SYSTEM_ROOT"]},
            ...
        )
```

That said, if it is more common for a user need many of them, this does cut down a bit on boilerplate. Though I can also imagine using a dict comp to do this in my code anyway:
```
env = {k: os.environ.get(k) for k in ["SYSTEM_ROOT", "LANGUAGE"]}
```

FollowTheProcess: I think the switch by itself is a good idea, potentially a rename to something a bit easier e.g. `include_outer`? But I agree on the type I think it should be just a bool on/off (with a default of `True` so the current default behaviour does not change).

I think 90%+ people who have a need to define env vars will have only a few. And if not the dictionary comp above works well, could even use a `.env` file or similar

FollowTheProcess: I'm also not a huge fan of maintaining a list of "special" env vars to always pass through, although I'm not entirely sure how else we'd do it

franekmagiera:
Thanks for doing the review! I agree @crwilcox's proposal. I think it addresses the original issue sufficiently, is clearer and easier to understand. This should be ready for another look.

theacodes:
This looks great to me, merging. :)

PR #653, opened 2022-09-23
Fix example in docs using `session.posargs`

paw-lu:
Current example feeds `session.posargs` directly into `session.run`. I don't _think_ this works—as `session.posargs` is a list. Added an unpacking argument `*` as a quick fix.

---

Thanks for the great work here!


FollowTheProcess (approved):
Hey @paw-lu nice catch! Thanks 🎉

PR #657, opened 2022-10-07
chore: some refurb suggestions and cleanup old Python 2-3.6 stuff

henryiii:
I tried running `refurb`, and it had a few good suggestions, I also found some old Python workarounds (for 3.6.1 and before) that could be cleaned up.

### DIFF
--- a/nox/_typing.py
+++ b/nox/_typing.py
@@ -14,33 +14,8 @@
 
 from __future__ import annotations
 
-__all__ = ["TYPE_CHECKING", "ClassVar", "NoReturn", "Python"]
+__all__ = ["Python"]
 
-import typing as _typing
+from typing import Sequence, Union
 
-try:
-    from typing import TYPE_CHECKING
-except ImportError:
-    try:
-        from typing import TYPE_CHECKING
-    except ImportError:
-        TYPE_CHECKING = False
-
-try:
-    from typing import NoReturn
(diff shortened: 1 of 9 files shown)


nox/_typing.py
```diff
@@ -14,33 +14,8 @@
 
 from __future__ import annotations
 
-__all__ = ["TYPE_CHECKING", "ClassVar", "NoReturn", "Python"]
```
henryiii, line 17: TYPE_CHECKING was added in 3.5.2, ClassVar in 3.5.3, and NoReturn in 3.5.4/3.6.2. All are now always present.

DiddiLeija (approved):
Sounds great to me! Thanks for the cleanup @henryiii!

henryiii:
Ahh, mistyped. Refurb.

henryiii:
It's a linter built using some of the mypy internals. So I'm therory it can do a few checks that other linters would have trouble getting. It's focused on "modernizing", a bit like pyupgrade, but willing to make suggestions that pyupgrade would not implement (like try/except pass for contextlib.suppress - pyupgrade can't add an import). I've found some of the suggestions helpful, and it's really easy to `pipx run refurb <src>` and just see what it suggests.

FollowTheProcess:
Ahh nice, I'll check that out!

PR #659, opened 2022-10-08
chore: switch to hatchling

henryiii:
Quick examples of moving to hatchling. Slightly less config, 2 seconds faster, less verbose build messages, ~~reduced chance of version conflicts~~, and slightly cleaner output. Other packages using hatchling include pipx, black, tox4, etc.

Just thought I'd throw it out there. Current system is completely fine too. Mostly making the PR since I tried it out and already had it. One benefit of PEP 621 is that it's easy to switch to the best performing backend. :)

PS: It occurred to me on the version conflicts (some packages limit setuptools < 60), that you don't have hatchling or setuptools when you have a wheel, so that should never be a problem. This is very much just a developer choice.

From #655.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ repository = "https://github.com/wntrblm/nox"
 nox = "nox.__main__:main"
 tox-to-nox = "nox.tox_to_nox:main"
 
+
 [tool.hatch]
 metadata.allow-ambiguous-features = true  # disable normalization (tox-to-nox) for back-compat
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
-build-backend = "setuptools.build_meta"
+build-backend = "hatchling.build"
 requires = [
-  "setuptools>=61",
+  "hatchling"
 ]
 
 [project]
@@ -63,11 +63,6 @@ repository = "https://github.com/wntrblm/nox"
 nox = "nox.__main__:main"
 tox-to-nox = "nox.tox_to_nox:main"
 
-
-[tool.setuptools]
-zip-safe = false
-include-package-data = true
-package-data = { "nox" = [ "py.typed" ] }
 
 [tool.isort]
 profile = "black"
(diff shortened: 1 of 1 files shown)


henryiii:
@ofek, I just swapped from `setuptools` to `hatchling`, and the `tox_to_nox` extra (which is declared that way and used that way) is getting normalized to `tox-to-nox`, which causes it to silently start failing. I don't see any way to provide a backward compatible `tox_to_nox` extra. AFAICT, this is more likely Pip's issue, since [PEP 685](https://peps.python.org/pep-0685/) was accepted? I see no mention there of PEP 685 support.

ofek:
https://hatch.pypa.io/latest/config/metadata/#allowing-ambiguous-features

ofek:
> Other packages using hatchling

See https://hatch.pypa.io/dev/users/

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [build-system]
-build-backend = "setuptools.build_meta"
+build-backend = "hatchling.build"
 requires = [
-  "setuptools>=61",
+  "hatchling"
```
henryiii, line 4:
```suggestion
  "hatchling",
```

henryiii:
The pyproject.toml formatter is insisting the comment be removed.

ofek:
https://github.com/tox-dev/pyproject-fmt/issues/16

pyproject.toml
```diff
@@ -62,11 +62,10 @@ repository = "https://github.com/wntrblm/nox"
 nox = "nox.__main__:main"
 tox-to-nox = "nox.tox_to_nox:main"
 
+# Disable extra name normalization (tox-to-nox) for back-compat.
+[tool.hatch]
+metadata.allow-ambiguous-features = true
```
henryiii, line 67:
```suggestion
[tool.hatch]
metadata.allow-ambiguous-features = true  # disable normalization (tox-to-nox) for back-compat
```

FollowTheProcess:
Nice, looks like this is basically ready to go?

henryiii:
Yes, except for the issue with pyproject.fmt not allowing a comment.

FollowTheProcess:
image

Looks like the recent fail is just adding an extra newline?

FollowTheProcess (approved):
Looks great 🎉 Thanks @henryiii

PR #662, opened 2022-10-20
Make GitHub Actions `uses` example runnable

paw-lu:
Currently example for GitHub actions will not run, and needs a version tag. Defaulted to using `@latest`.

(At least partially) closes #643

---

Thanks for all the work here. Really enjoy using nox.

PR #664, opened 2022-10-20
Point action docs to 2022.8.7 not latest

FollowTheProcess:
Fixes a mention in the docs to `@latest` which doesn't actually refer to a valid tag, my bad for not spotting! 👍🏻 Thanks @henryiii

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -36,6 +36,7 @@ If you want to run ``nox`` within `GitHub Actions`_, you can use the ``wntrblm/n
     # setup nox with all active CPython and PyPY versions provided by
     # the GitHub Actions environment i.e.
     # python-versions: "3.7, 3.8, 3.9, 3.10, pypy-3.7, pypy-3.8, pypy-3.9"
+    # this uses version 2022.8.7 but any Nox tag will work here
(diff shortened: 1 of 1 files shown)


DiddiLeija (approved):
Thank you both @FollowTheProcess and @henryiii!

Small question: Is `@main` possible, instead of `@latest`? If not, I'm ok with merging this :+1:

DiddiLeija:
> Small question: Is `@main` possible, instead of `@latest`? If not, I'm ok with merging this 👍

Ahh sorry, I didn't read #662 yet :/

henryiii:
`@main` is fine too, you can put any branch or tag name there.

DiddiLeija:
Ok. I'm personally more comfortable with `@main`, but a tag is fine too (however, we'll have to update that tag frequently).

FollowTheProcess:
A specific tag is always preferable IMO, it means a user of the action is always getting an exact version rather than whatever is on the main branch (which may, although unlikely, break). What do we think?

DiddiLeija:
> If we need to keep it up to date with new versions there's things like bump2version that can find and replace, or we just leave it with a comment that says e.g "any tagged version will work here"

That would resolve my concerns of using a specific tag!

Issue #658, opened 2022-10-07
FEAT: json session listing


henryiii:
### How would this feature be useful?

GitHub actions (and maybe others) let you configure the jobs dynamically using JSON. This was used very effectively by cibuildwheel (see https://iscinumpy.dev/post/cibuildwheel-2-10-0/#only-210). I was thinking this might be really nice for nox, too. Then the session list would be the source of truth for the CI Python jobs.

### Describe the solution you'd like

A new flag, like `--json-sessions`, could be added to augment the `-l` list sessions command. It doesn't have to be json, but that's likely the easiest to process if there were multiple items in the output (and you could include handy items like the help text, and add to it later without breaking usage). Something like this:

```console
$ nox --json-sessions
{"name": "tests-3.7", "help": "Run test suite with pytest."}
{"name": "tests-3.8", "help": "Run test suite with pytest."}
...
```

It should be possible to combine `--json-sessions` with `-s` session selections, just like it is with `-l` today, which would make it easy to produce just the correct subset.

### Describe alternatives you've considered

The current output could be parsed, but it's not very amenable to parsing, and more info (like the base session name & the python interpreter) would be nice to have. It doesn't have to be json, but it should be easily machine readable (which json is). And technically I used json lines above, though regular json is fine too - either is easy to run through jq. It also should be only the active values (or have a bool for selected on/off).

### Anything else?

Here's with another option (python) added:

```console
$ nox --json-sessions
{"name": "tests-3.7", "python": "3.7", "help": "Run test suite with pytest."}
{"name": "tests-3.8", "python": "3.7",  "help": "Run test suite with pytest."}
...
{"name": "lint", "python": "3.x",  "help": "Run pre-commit linting."}
...
```

Then you could easily have enough for `python-version`.

I've not carefully worked this through, but thought I'd put it down to see what people thought.

FollowTheProcess:
+1 from me, I like this idea

PR #665, opened 2022-10-20
feat: --json --list-sessions

henryiii:
Closes https://github.com/wntrblm/nox/issues/658.

No tests yet. Putting out an initial design.

Here's the output on nox itself:

```bash
nox --json -l -s cover lint | jq
```
```json
[
  {
    "session": "cover",
    "name": "cover",
    "description": "Coverage analysis.",
    "python": null,
    "tags": [],
    "call_spec": {}
  },
  {
    "session": "lint",
    "name": "lint",
    "description": "Run pre-commit linting.",
    "python": "3.9",
    "tags": [],
    "call_spec": {}
  }
]
```

And it can be processed by jq:

```console
$ nox --json -l -s tests | jq '.[].session'
"tests-3.7"
"tests-3.8"
"tests-3.9"
"tests-3.10"
"tests-3.11"
```

Thoughts?

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -30,6 +30,11 @@ To list all available sessions, including parametrized sessions:
     nox --list
     nox --list-sessions
 
+If you'd like to use the output in later processing, you can add ``--json`` to
+get json output for the selected session. Fields include ``session`` (pretty
+name), ``name``, ``description``, ``python`` (null if not specified), ``tags``,
+and ``call_spec`` (for parametrized sessions).
+
 
 .. _session_execution_order:
 
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -265,6 +265,13 @@ options.add_options(
         help="List all available sessions and exit.",
     ),
     _option_set.Option(
+        "json",
+        "--json",
+        group=options.groups["sessions"],
+        action="store_true",
+        help="JSON output formatting. Requires list-sessions currently.",
+    ),
+    _option_set.Option(
         "sessions",
         "-s",
         "-e",
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -184,6 +184,11 @@ def filter_manifest(manifest: Manifest, global_config: Namespace) -> Manifest | 
         _produce_listing(manifest, global_config)
         return 0
 
+    # JSON output requires list sessions also be specified
+    if global_config.json and not global_config.list_sessions:
+        logger.error("Must specify --list-sessions with --json")
+        return 3
+
     # Filter by python interpreter versions.
     if global_config.pythons:
         manifest.filter_by_python_interpreter(global_config.pythons)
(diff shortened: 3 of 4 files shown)


henryiii:
Any thoughts on the general design?

FollowTheProcess:
Hey @henryiii sorry this one slipped by me! I like it and was actually surprised at how little needed to change to support this. For those wondering how it ends up looking:

image

Just a couple of questions more than anything.

When sessions don't have a python the output from this is `null` e.g. from our own `docs` session:

image

I wonder if there's something more helpful we can put here? This seems to be when we use the `nox.session` decorator without calling it so Nox uses the python that called it to do the virtualenv, could we populate the JSON python version with the version from `sys.version_info`? A quick POC on my local copy of your branch seems to work okay:

```python
report.append(
      {
          "python": session.func.python or f"{sys.version_info.major}.{sys.version_info.minor}",
      }
  )
```

Finally a very minor one but I'm not sure of the usefulness of the name `call_spec` to anyone other than nox developers. I think a more user friendly term here would be good, something that makes it clearer this is to do with `parametrize`?

Other than that though this looks great!

henryiii:
> Null output

This can be used to detect when the Python version doesn't matter. I'd assume a user would want to fill nulls with `3.x` on GitHub Actions, and possibly something else on other platforms. (I thought about injecting `3.x`, but that could be a bit GHA specific, whereas I'd want this to be general? You can always replaced a known, unique string, though, so this wouldn't be too bad.

I wouldn't want to inject the actual running Python because you lose the info that it doesn't matter, and you might not be running the listing on the same Python as you run the check later - that shouldn't matter.

> `call_spec`

I wasn't sure what to call it and being consistent seemed the simplest thing. `¯\_(ツ)_/¯`

FollowTheProcess:
Ahh yeah good point re. "python doesn't matter"

> I wasn't sure what to call it and being consistent seemed the simplest thing

That's fair, naming is hard after all! None of this was a blocker for me, this looks great 👍🏻

henryiii:
That doesn't mean it has to be named "call_spec", I just was explaining where the name came from. :)

FollowTheProcess:
Well I'm in a similar boat to be honest, I can't really think of a good name either 👀

FollowTheProcess:
When dumping the JSON is it worth prettying it with `indent` or something do you think? Obviously computers don't care but it makes it a bit more human readable without using e.g. `jq`?

henryiii:
> is it worth prettying it with `indent`

I considered it, in fact I did do that at first, but dropped it when it's so easy to do it with jq, and the point of the output is to be processed, not to be human readable. But this is not something where saving a few whitespace characters matters, so indent is fine too.

My next plan is to try this on a repo somewhere, to see what processing this looks like. I've fixed my blog post's example something like 3-4 times because I missed things. That will probably make good documentation, too. :)

FollowTheProcess:
Yeah agreed, I think it's fine to leave it out. Even if people pipe it to files as soon as they open them in e.g. VSCode it'll probably format it nicely anyway.

> That will probably make good documentation, too. :)

This would be great! If you can find some useful tricks it might be work making a new recipe in the [Cookbook](https://nox.thea.codes/en/stable/cookbook.html)

I'd be interested to see how people might use this actually, it feels like it could be quite powerful in e.g. CI systems but I can't quite think of a use case in anything I'm doing at the moment

FollowTheProcess:
Where did you get to on this @henryiii? With some tests this should be ready to go as far as I can tell?, would be great to get this in for the next release :)

henryiii:
Finally tried this out: https://github.com/scikit-build/scikit-build-sample-projects/pull/31 :)

henryiii:
Okay, unless someone wants something like `--format=json` instead of `--json`, this is ready. Probably needs a cookbook example or two eventually.

FollowTheProcess (approved):
Looks great! Thanks @henryiii

PR #667, opened 2022-10-27
chore: move to 3.11 final

henryiii:
This moves to using Python 3.11 final on GHA as well as uses it by default for the action.

While doing this, I noticed we are using the deprecated set-output string. I'll try to make a followup that fixes this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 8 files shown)


henryiii:
>  I'll try to make a followup that fixes this.

Looks like #668 beat me to it. That's fine :) I'm happy to rebase if you want to put that one in first.

.github/workflows/ci.yml
```diff
@@ -27,9 +27,9 @@ jobs:
           python -m pip install --disable-pip-version-check .
       - name: Run tests on ${{ matrix.os }}
         run: nox --non-interactive --error-on-missing-interpreter --session "tests-${{ matrix.python-version }}" -- --full-trace
-        if: matrix.python-version != '3.11-dev'
+        if: matrix.python-version != '3.11'
       - name: Run tests on ${{ matrix.os }}
-        if: matrix.python-version == '3.11-dev'
+        if: matrix.python-version == '3.11'
         run: nox --non-interactive --error-on-missing-interpreter --session "tests-3.11" -- --full-trace
```
FollowTheProcess, line 33: We can just get rid of the `if` entirely now can't we and just use the `{{ matrix.python-version }}`?

henryiii: Ahh, yes, I think so. Can try to do it tonight.

FollowTheProcess:
Nice! Thanks @henryiii 🎉

PR #668, opened 2022-10-27
Action: Replace deprecated set-output command

airtower-luna:
The `set-output` workflow command [has been deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and replaced with writing to an environment file. Update `.github/action_helper.py` to use the new file.

### DIFF
--- a/.github/action_helper.py
+++ b/.github/action_helper.py
@@ -65,9 +65,9 @@ def setup_action(input_: str) -> None:
     if len(versions) > 20:
         raise ValueError(f"too many interpreters to install: {len(versions)} > 20")
 
-    print(f"::set-output name=interpreter_count::{len(versions)}")
+    print(f"interpreter_count={len(versions)}")
     for i, version in enumerate(versions):
-        print(f"::set-output name=interpreter_{i}::{version}")
+        print(f"interpreter_{i}={version}")
 
 
 if __name__ == "__main__":
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/action.yml'
+      - '.github/action_helper.py'
       - 'action.yml'
 env:
   FORCE_COLOR: "1"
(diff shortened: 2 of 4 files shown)


.github/action_helper.py
```diff
@@ -65,9 +66,10 @@ def setup_action(input_: str) -> None:
     if len(versions) > 20:
         raise ValueError(f"too many interpreters to install: {len(versions)} > 20")
 
-    print(f"::set-output name=interpreter_count::{len(versions)}")
-    for i, version in enumerate(versions):
-        print(f"::set-output name=interpreter_{i}::{version}")
+    with open(os.environ["GITHUB_OUTPUT"], "a") as fh:
```
henryiii, line 69: I was thinking it could just print the output, then this could be run as `python action_helper.py > $GITHUB_OUTPUT`. Might be a bit easier to test, too?

airtower-luna: Good point, that way the test doesn't need the fixture that monkeypatches the `GITHUB_OUTPUT` variable, and the risk of the test accidentally setting an output if there's a bug is gone. :smiley_cat:

airtower-luna: And updated.

DiddiLeija (approved):

FollowTheProcess:
This looks good, just needs a conflict resolution because I merged #667 in before this one 👍🏻

airtower-luna:
Does anyone have an idea what went wrong in the "macos-latest, 3.8" build? The log looks to me like a `KeyboardInterrupt` within Nox' process management bubbled up to pytest, but that's code I haven't touched, and it passed CI on `main`, so I'm confused what I should do. :thinking:

FollowTheProcess:
Looks like a spurious error to me, I'll re-run and see what happens

FollowTheProcess:
Okay, well now it's 3.11 apparently!

FollowTheProcess:
5th time's the charm it seems!

FollowTheProcess:
Thanks @airtower-luna 🎉

PR #670, opened 2022-11-21
Fix assertion in GHA tests

FollowTheProcess:
I noticed that on the recent release, the `action_default_tests` were failing asserting that the calling python was 3.10 when in fact it was 3.11 due to #667 

This PR updates the assertion to 3.11

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -2,13 +2,13 @@ name: Action
 on:
   push:
     branches:
-      - 'main'
-      - '**action**'
+      - "main"
+      - "**action**"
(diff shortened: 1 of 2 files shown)


FollowTheProcess:
Hmm looks like Ubuntu 18.04 still defaults to 3.10?

FollowTheProcess:
Also, only just noticed but we only run `actions_all_tests` on Windows for some reason?

https://github.com/wntrblm/nox/blob/6957a4c1bf4f8f403fc30664812cbb19eb0b2417/.github/workflows/action.yml#L25

PR #675, opened 2022-12-05
ci: some minor fixes

henryiii:
Fixes to get CI working again, and some speed improvements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,14 +1,22 @@
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
+
 env:
   FORCE_COLOR: "1"
-  PRE_COMMIT_COLOR: "always"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-20.04, windows-latest, macos-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,14 +1,22 @@
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
```
henryiii, line 4: Without this, CI is currently running twice on things like dependabot PRs.

.github/workflows/ci.yml
```diff
@@ -1,14 +1,22 @@
...
+  pull_request:
+
 env:
   FORCE_COLOR: "1"
-  PRE_COMMIT_COLOR: "always"
```
henryiii, line 5: I believe this was fixed in pre-commit.

henryiii: Hmm, maybe not. Just a sec.

henryiii: No, I take that back.

.github/workflows/ci.yml
```diff
@@ -18,10 +26,11 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Set up Miniconda
         uses: conda-incubator/setup-miniconda@v2
-        if: matrix.python-version != '3.11'
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
+          miniforge-variant: Mambaforge
+          use-mamba: true
```
henryiii, line 33: A lot faster & supports 3.11.

henryiii:
The coverage issue:

https://github.com/wntrblm/nox/blob/cc710bde9d6a8781833144bac02a5f4581d9eca7/nox/virtualenv.py#L373-L394

I think this is never None on Ubuntu 22.04 & Python 3.11. Shouldn't the 100% requirement be on the combined coverage, rather than partial? Anyway, should this be marked as not being required to be covered, etc?

crwilcox (approved):

FollowTheProcess:
Hmmm wonder why it would never be `None` on ubuntu-latest py3.11? Weird.

Agree on combined coverage being used for the `fail-under`, is this a config option in coverage? I did have a quick look but couldn't see anything obvious.

If we can't do that then yeah I agree we should exclude it from coverage measurement.

This all looks good to me 👍🏻

PR #678, opened 2022-12-09
chore: long term fix for bugbear opinionated checks

henryiii:
This is a better fix for the opinionated checks for bugbear. Since they may be added supporting newer than EoL Pythons, it’s best to just activate the known safe ones. Otherwise this might have to be fixed every release that adds a new opinionated check.


FollowTheProcess (approved):
Nice! 🎉

PR #679, opened 2022-12-09
Add invite link to the discord server to CONTRIBUTING.md

dcermak:


FollowTheProcess:
Any issues with the invite link being public? Not overly familiar with discord so don't know if this is the done thing or not

dcermak:
Many projects that I have seen using discord put a join link into their readme or somewhere else in a prominent place, so I guess there shouldn't be huge ramifications. Of course this will make it easier for people to join, so the server could become more crowded and require more moderation from the server admin.

DiddiLeija:
> Of course this will make it easier for people to join, so the server could become more crowded and require more moderation from the server admin.

The Winterbloom server is a nicely moderated server, it has their rules, and the community is big (even without a lot of Nox-interested people). So I don't think there's a problem at all :)

theacodes:
Absolutely fine for the discord invite to be public.

DiddiLeija (approved):

FollowTheProcess:
Thanks @dcermak 🎉

Issue #673, opened 2022-11-30
Compatibility with tox 4


frenzymadness:
### Current Behavior

Tox 4 is now in release candidate 1 and it might make sense to take a look at how to support it here in tox_to_nox. Right now, tests results are:

```
FAILED tests/test_tox_to_nox.py::test_trivial - AttributeError: module 'tox.c...
FAILED tests/test_tox_to_nox.py::test_skipinstall - AttributeError: module 't...
FAILED tests/test_tox_to_nox.py::test_usedevelop - AttributeError: module 'to...
FAILED tests/test_tox_to_nox.py::test_commands - AttributeError: module 'tox....
FAILED tests/test_tox_to_nox.py::test_deps - AttributeError: module 'tox.conf...
FAILED tests/test_tox_to_nox.py::test_env - AttributeError: module 'tox.confi...
FAILED tests/test_tox_to_nox.py::test_chdir - AttributeError: module 'tox.con...
FAILED tests/test_tox_to_nox.py::test_dash_in_envname - AttributeError: modul...
FAILED tests/test_tox_to_nox.py::test_non_identifier_in_envname - AttributeEr...
FAILED tests/test_tox_to_nox.py::test_descriptions_into_docstrings - Attribut...
```

All failures caused by

```
def main() -> None:
        parser = argparse.ArgumentParser(description="Converts toxfiles to noxfiles.")
        parser.add_argument("--output", default="noxfile.py")
    
        args = parser.parse_args()
    
>       config = tox.config.parseconfig([])
E       AttributeError: module 'tox.config' has no attribute 'parseconfig'
```

### Expected Behavior

It would be nice to have nox compatible with tox 4 when they release a final version.

### Steps To Reproduce

Run nox tests with tox==4.0.0rc1

### Environment

```Markdown
- OS:
- Python:
- Nox:
```

FollowTheProcess:
This is definitely something we'd like to continue to support. I've just had a little poke around `4.0.0rc1` and can't see an equivalent way of Nox loading the tox config file, admittedly I only did a quick scan.

It's worth looking into how we'd do this before deciding how to proceed

FollowTheProcess:
So it turns out there is no supported way of parsing the tox config in version 4 that we can use. See the [discussion](https://discuss.python.org/t/tool-tox-4-on-the-horizon/6537/29?u=followtheprocess).

Looks like the best we can do is:

- Detect which tox version is installed at runtime
- If <4 then use the current parsing method
[...]

PR #687, opened 2023-01-24
First implementation of tox_to_nox for tox 4

frenzymadness:
Fixes: #673 

It seems to me that the easiest way forward is to call `tox config`, parse the output and then preprocess it into a dictionary so it can be easily used in the jinja template.

What do you think?

From this tox.ini:
```
[tox]
envlist = black,mypy,py{36,37,38,39,310,311,312}
isolated_build = True

[testenv]
commands =
    python -m platform
    pytest -v {posargs} test.py
deps =
    pytest

[testenv:black]
description="better than blue"
skip_install=true
deps=black
commands=black --check --diff --extend-exclude=".tox|test" ./

[testenv:mypy]
use_develop=true
deps=mypy
commands=mypy --strict -p marshalparser
```
it now generates this noxfile.py
```python
import nox


@nox.session()
def DEFAULT(session):
    """"""
    session.install('.')

@nox.session(python='/home/lbalhar/.virtualenvs/nox/bin/python')
def testenv_black(session):
    """"better than blue""""
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('black')
    session.run('black --check --diff '--extend-exclude=.tox|test' ./')

@nox.session(python='/home/lbalhar/.virtualenvs/nox/bin/python')
def testenv_mypy(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('mypy')
    session.install('-e', '.')
    session.run('mypy --strict -p marshalparser')

@nox.session(python='py310')
def testenv_py310(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

@nox.session(python='py311')
def testenv_py311(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

@nox.session(python='py312')
def testenv_py312(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

@nox.session(python='py36')
def testenv_py36(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

@nox.session(python='py37')
def testenv_py37(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

@nox.session(python='py38')
def testenv_py38(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

@nox.session(python='py39')
def testenv_py39(session):
    session.env['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'
    session.env['PYTHONIOENCODING'] = 'utf-8'
    session.install('pytest')
    session.install('.')
    session.run('python -m platform')
    session.run('pytest -v test.py')

```

I guess that:
* the default env variables should be skipped.
* the default environment should be skipped.

Correct?

### DIFF
--- a/nox/tox_to_nox.py
+++ b/nox/tox_to_nox.py
@@ -17,24 +17,38 @@
 from __future__ import annotations
 
 import argparse
+import os
 import pkgutil
-from typing import Any, Iterator
+import re
+from configparser import ConfigParser
+from pathlib import Path
+from subprocess import check_output
+from typing import Any, Iterable
 
 import jinja2
 import tox.config
+from tox import __version__ as TOX_VERSION
 
-_TEMPLATE = jinja2.Template(
-    pkgutil.get_data(__name__, "tox_to_nox.jinja2").decode("utf-8"),  # type: ignore[union-attr]
-    extensions=["jinja2.ext.do"],
-)
+TOX4 = TOX_VERSION[0] == "4"
+
+if TOX4:
+    _TEMPLATE = jinja2.Template(
+        pkgutil.get_data(__name__, "tox4_to_nox.jinja2").decode("utf-8"),  # type: ignore[union-attr]
+        extensions=["jinja2.ext.do"],
+    )
+else:
+    _TEMPLATE = jinja2.Template(
+        pkgutil.get_data(__name__, "tox_to_nox.jinja2").decode("utf-8"),  # type: ignore[union-attr]
+        extensions=["jinja2.ext.do"],
+    )
 
 
-def wrapjoin(seq: Iterator[Any]) -> str:
+def wrapjoin(seq: Iterable[Any]) -> str:
     return ", ".join([f"'{item}'" for item in seq])
 
 
 def fixname(envname: str) -> str:
-    envname = envname.replace("-", "_")
+    envname = envname.replace("-", "_").replace("testenv:", "")
     if not envname.isidentifier():
         print(
             f"Environment {envname!r} is not a valid nox session name.\n"
@@ -49,7 +63,61 @@ def main() -> None:
 
     args = parser.parse_args()
 
-    config = tox.config.parseconfig([])
+    if TOX4:
+        output = check_output(["tox", "config"], text=True)
+        original_config = ConfigParser()
+        original_config.read_string(output)
+        config: dict[str, dict[str, Any]] = {}
+
+        for name, section in original_config.items():
+            if name == "DEFAULT":
+                continue
+
+            config[name] = dict(section)
+            # Convert set_env from string to dict
+            set_env = {}
+            for var in section.get("set_env", "").strip().splitlines():
+                k, v = var.split("=")
+                if k not in (
+                    "PYTHONHASHSEED",
+                    "PIP_DISABLE_PIP_VERSION_CHECK",
+                    "PYTHONIOENCODING",
+                ):
+                    set_env[k] = v
+
+            config[name]["set_env"] = set_env
+
+            config[name]["commands"] = [
+                wrapjoin(c.split()) for c in section["commands"].strip().splitlines()
+            ]
+
+            config[name]["deps"] = wrapjoin(section["deps"].strip().splitlines())
+
+            for option in "skip_install", "use_develop":
+                if section.get(option):
+                    if section[option] == "False":
+                        config[name][option] = False
+                    else:
+                        config[name][option] = True
+
+            if os.path.isabs(section["base_python"]) or re.match(
+                r"py\d+", section["base_python"]
+            ):
+                impl = (
+                    "python" if section["py_impl"] == "cpython" else section["py_impl"]
+                )
+                config[name]["base_python"] = impl + section["py_dot_ver"]
+
+            change_dir = Path(section.get("change_dir"))
+            rel_to_cwd = change_dir.relative_to(Path.cwd())
+            if str(rel_to_cwd) == ".":
+                config[name]["change_dir"] = None
+            else:
+                config[name]["change_dir"] = rel_to_cwd
+
+    else:
+        config = tox.config.parseconfig([])
+
     output = _TEMPLATE.render(config=config, wrapjoin=wrapjoin, fixname=fixname)
 
     with open(args.output, "w") as outfile:
--- a/nox/tox_to_nox.py
+++ b/nox/tox_to_nox.py
@@ -24,7 +24,7 @@ from collections.abc import Iterator
 from configparser import ConfigParser
 from pathlib import Path
 from subprocess import check_output
-from typing import Any, Iterable
+from typing import Any
 
 import jinja2
 import tox.config
--- a/nox/tox_to_nox.py
+++ b/nox/tox_to_nox.py
@@ -20,11 +20,10 @@ import argparse
 import os
 import pkgutil
 import re
-from collections.abc import Iterator
 from configparser import ConfigParser
 from pathlib import Path
 from subprocess import check_output
-from typing import Any
+from typing import Any, Iterable
 
 import jinja2
 import tox.config
@@ -44,7 +43,7 @@ else:
     )
 
 
-def wrapjoin(seq: Iterator[Any]) -> str:
+def wrapjoin(seq: Iterable[Any]) -> str:
     """Wrap each item in single quotes and join them with a comma."""
     return ", ".join([f"'{item}'" for item in seq])
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -31,12 +31,29 @@ if shutil.which("conda"):
     nox.options.sessions.append("conda_tests")
 
 
-@nox.session(python=["3.7", "3.8", "3.9", "3.10", "3.11"])
-def tests(session: nox.Session) -> None:
+@nox.session
+@nox.parametrize(
+    "python, tox_version",
+    [
+        (python, tox_version)
+        for python in ("3.7", "3.8", "3.9", "3.10", "3.11")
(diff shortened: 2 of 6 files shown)


nox/tox_to_nox.py
```diff
@@ -21,20 +21,31 @@
...
-import tox.config
+from tox import __version__ as TOX_VERSION
 
-_TEMPLATE = jinja2.Template(
-    pkgutil.get_data(__name__, "tox_to_nox.jinja2").decode("utf-8"),  # type: ignore[union-attr]
-    extensions=["jinja2.ext.do"],
-)
+TOX3 = TOX_VERSION[0] == "3"
+
+if TOX3:
```
FollowTheProcess, line 28: I think it's better to check for tox 4 here and flip the conditional? That way we only do the tox 4 logic for tox 4, everything else goes through the previous implementation

FollowTheProcess (commented):
Hey @frenzymadness thanks for the PR! I agree with the approach of parsing the config from the command line, I think that's realistically the best we can do for tox 4 compatibility.

Just a few comments from my side but this looks like a great start 👍🏻 Will need some tests before it's ready to merge but appreciate the draft to sign off the approach

nox/tox_to_nox.py
frenzymadness: Done

nox/tox_to_nox.py
```diff
@@ -21,20 +21,31 @@
...
+    import tox.config
+    _TEMPLATE = jinja2.Template(
+        pkgutil.get_data(__name__, "tox_to_nox.jinja2").decode("utf-8"),  # type: ignore[union-attr]
+        extensions=["jinja2.ext.do"],
+    )
+else:
+    from subprocess import check_output
+    from configparser import ConfigParser
```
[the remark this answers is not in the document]
frenzymadness: Done

frenzymadness:
Right now, all the tests are happy with both tox versions. I skipped just the test of the invalid identified in the env name because it seems that tox 4 does not support it.

Another problem is how to include tests with tox 3 into nox's tests. I was thinking about parametrizing the main test session but it seems to be a waste of resources to run all the tests again with tox 3 if the vast majority doesn't depend on tox. My solution is to downgrade tox to version <4 and then run the pytest again in the same session but only the tests of tox_to_nox.

FollowTheProcess:
Just looks like a couple of uncovered sections in tox_to_nox

frenzymadness:
> Just looks like a couple of uncovered sections in tox_to_nox

What do you mean? I'm looking at both jinja2 templates side-by-side and they both seem to handle the same configuration options from tox.ini.

frenzymadness:
The problem reported by coverage was actually caused by two conditions that were never false so I can remove them, nice. All linter should be happy now, at least they are happy on my machine.

frenzymadness:
And now we have a different problem with dependencies. argcomplete 2.0.0 requires `importlib-metadata<5,>=0.23; python_version == "3.7"` but the latest tox depends on `importlib-metadata>=6; python_version < "3.8"` which means that they cannot be installed together in an environment with Python 3.7 and therefore both pytest runs tests with tox 3 and hence the coverage is less than 100 %.

We can force install tox 4 in a dedicated step in Python 3.7 env but there is a risk that it breaks nox or its dependencies.

I'd set a lower coverage limit for Python 3.7 or skip the coverage there entirely. What do you think?

frenzymadness:
@FollowTheProcess what do you think about the proposed solutions?

FollowTheProcess:
Hmm, so this would mean anyone installing Nox on 3.7 would never get tox 4 right? So if that's the case I don't think we need to be super strict on getting to 100% because users would never hit that branch either?

I'd personally be happy to mark that branch as excluded from coverage when on 3.7, with a nice explanatory comment etc. I think coverage allows you to do this, not 100% sure

frenzymadness:
Let's see what will CI say now.

frenzymadness:
The problem here is in the way how the Github Actions are configured. If you run nox test locally, it's green because the coverage report is combined from all runs with all Python versions and therefore the lower overage from Python 3.7 is not a problem. But in GH actions, the coverage report is divided into jobs for each tested Python version which makes the job for Python 3.7 fail on too low coverage.

I see two ways forward:
1. Do not run coverage for Python 3.7 on GH actions.
2. Reconfigure GH actions to collect coverage for individual jobs and run coverage report for the collected data in a separate job.

What do you prefer?

frenzymadness:
I've implemented what I think is the simplest solution here - no `session.notify("cover")` for Python 3.7. That way, the coverage report is run only if called explicitly or if tests run for more than this one environment which should also fix the coverage so it should be green all the time.

noxfile.py
```diff
@@ -46,6 +46,28 @@ def tests(session: nox.Session) -> None:
...
+    session.install("tox<4")
+    session.run(
+        "pytest",
+        "--cov=nox",
+        "--cov-config",
+        "pyproject.toml",
+        "--cov-report=",
+        "tests/test_tox_to_nox.py",
+        env={"COVERAGE_FILE": f".coverage.{session.python}.tox3"},
+    )
```
FollowTheProcess, line 70: Seems like a good case for `nox.parametrize`? Can parametrize on python 3.7 - 3.11 as well as tox 3 and 4?

FollowTheProcess: Tried this myself this morning and I think it's doing what we want, what do you think?:

```python
@nox.session
@nox.parametrize(
    "python, tox",
    [
        (python, tox)
        for python in ("3.7", "3.8", "3.9", "3.10", "3.11")
        for tox in ("3", "4")
        if (python, tox) != ("3.7", "4")
    ],
)
def tests(session: nox.Session, tox: str) -> None:
    """Run test suite with pytest."""
    session.create_tmp()  # Fixes permission errors on Windows
    session.install("-r", "requirements-test.txt")
    session.install(".", "jinja2", f"tox=={tox}")
    session.run(
        "pytest",
        "--cov=nox",
        "--cov-config",
        "pyproject.toml",
        "--cov-report=",
        *session.posargs,
        env={"COVERAGE_FILE": f".coverage.{session.python}"},
    )
    session.notify("cover")
```

frenzymadness: I had this ready from the previous night and my implementation is little bit different. Take a look at it and let me know which one do you like more.

FollowTheProcess: Yeah yours works too 👍🏻 Happy with that

frenzymadness:
But now we are back at the problem with coverage in GH action. If we split the jobs in GH action in the same way we have them in the parametrized sessions in the noxfile, the coverage will fail for all of them because they won't be tested with both tox versions at the same time.

I'm going to enable GH actions in my fork and try to implement a single job for the coverage reporting.

frenzymadness:
Finally, after many failures, it's done. Now, each session in noxfile has its own job in GH actions. The produced coverage files are uploaded by the individual jobs. When all testing jobs finish, the coverage job starts, downloads the files, combines them and provides the report.

Locally, it no longer makes sense to run coverage if you test only one environment, because it cannot be 100 %, so the notification of coverage session has been removed.

You can see the last run in my fork here: https://github.com/frenzymadness/nox/actions/runs/4224370541

mtelka:
tox 3 is dead.  Please go forward and merge this.  Thank you.

crwilcox:
@FollowTheProcess I think it might be worth 'resurecting' this and merging. There are some straightforward conflicts to handle, and this would be an improvement for users. If you are also on board I can get this over the line if needed.

Thanks @frenzymadness for picking this up.

nox/tox_to_nox.py
```diff
@@ -37,7 +51,7 @@ def wrapjoin(seq: Iterator[Any]) -> str:
 
 def fixname(envname: str) -> str:
     """Replace dashes with underscores and check if the result is a valid identifier."""
-    envname = envname.replace("-", "_")
+    envname = envname.replace("-", "_").replace("testenv:", "")
```
crwilcox, line 54: I'm not sure why this is needed, and if it would have a negative effect on old tox (tox <= 3)

FollowTheProcess:
Yeah, sorry been swamped for a while and not got chance to be around as much as I'd have liked! I'm on board 👍🏻

nox/tox_to_nox.py
crwilcox: Verified purpose.

crwilcox (approved):

PR #688, opened 2023-01-27
Add support for `NOXPYTHON`, `NOXEXTRAPYTHON` and `NOXFORCEPYTHON`

### LINKED ISSUE #672: Support a `NOXPYTHON` environment variable similar to `NOXSESSION`

It'd save a few keystrokes and would work as a default for `--pythons`.

Check for `NOXPYTHON` environment variable:

```shell
NOXPYTHON=3.11 nox -s tests
NOXPYTHON=3.10,3.11 nox -s tests
```

While we're at it, it might make sense to also add env vars for `--extra-pythons` and `--force-pythons`.

Continue using `--pythons`, or include the Python version in `NOXSESSION`, i.e.

```dotenv
NOXSESSION=tests-3.10
```

but this couples the default session with a Python version

### Anything else?

_No response_

edgarrmondragon:
Closes #672

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
       - id: isort
         args: ["-a", "from __future__ import annotations"]
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -147,6 +147,9 @@ def default_env_var_list_factory(env_var: str) -> Callable[[], list[str] | None]
 
     Args:
         env_var (str): The name of the environment variable to look up.
+
+    Returns:
+        A callback that retrieves a list from a comma-delimited environment variable.
     """
 
     def _default_list() -> list[str] | None:
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -18,7 +18,7 @@ import argparse
 import functools
 import os
 import sys
-from typing import Any, Sequence
+from typing import Any, Callable, Sequence
 
 from nox import _option_set
 from nox.tasks import discover_manifest, filter_manifest, load_nox_module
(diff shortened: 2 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
```
DiddiLeija, line 28: Not sure if `isort` is the only dependency that requires an update. Would like to address that in another thread, though.

edgarrmondragon: Yeah, I only update `isort` since it was causing the lint check to fail

DiddiLeija: Yup, I've just noticed that. Weird issue though :/

edgarrmondragon: @DiddiLeija I've started https://github.com/wntrblm/nox/pull/690 to update all the pre-commit hooks.

Perhaps the pre-commit.ci integration should be added to the repo so `pre-commit autoupdate` is run automatically.

DiddiLeija: > Perhaps the pre-commit.ci integration should be added to the repo so `pre-commit autoupdate` is run automatically.

I think we've discussed that before, am I right @wntrblm/nox @henryiii ?

henryiii: I'm a big fan of pre-commit.ci.

Poetry-core 1.5.0 broke isort. So you can't build anything before isort 5.12.0 from source. A little ironic given Poetry's stance on SemVer...

I'd like to move over to using Ruff soon, I'll probably make a PR in the near future.

FollowTheProcess: Big fan of ruff, would love to see that in Nox!

edgarrmondragon: +1 on Ruff 😄

DiddiLeija: I don't know a lot about Ruff, but seems like a nice approach! :smile:

nox/_options.py
```diff
@@ -142,11 +142,18 @@ def _envdir_merge_func(
...
-    """Looks at the NOXSESSION env var to set the default value for sessions."""
-    nox_env = os.environ.get("NOXSESSION")
-    env_sessions = nox_env.split(",") if nox_env else None
-    return env_sessions
+def default_env_var_list_factory(env_var: str) -> Callable[[], list[str] | None]:
+    """Looks at the env var to set the default value for a list of env vars.
+
+    Args:
+        env_var (str): The name of the environment variable to look up.
+    """
```
FollowTheProcess, line 150: Would ideally like a `Returns` section here, particularly as we're returning a callable which is fine and a good solution for this, just probably not what most people will expect

FollowTheProcess (changes requested):
Nice! Just a small nit around a docstring. Love the sphinx tabs by the way they look great!

nox/_options.py
edgarrmondragon: @FollowTheProcess thanks for the feedback! Done: d1887d5413d6c9d37caa98a455a2243bcc452ec9

FollowTheProcess (approved):
Nice work, LGTM

DiddiLeija (approved):
Thank you @edgarrmondragon!

edgarrmondragon:
I think all the feedback's been addressed. Feel free to merge 😄

PR #691, opened 2023-02-08
chore: move to using Ruff

henryiii:
Ruff can take the place of a lot of other tools, and is 10-100x faster (written in Rust), plugins can't get out of sync, zero dependencies, can auto-fix in some cases, pyproject.toml configuration, and poetry-core updates can't break it (like isort). :)

cibuildwheel, build, scipy, and others have already moved to it.

There are some other nice codes we could enable, but stuck with either what was there before or easy-to-enable (minimal changes) ones.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,18 +26,18 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: "0.7.0"
+    rev: "0.9.1"
     hooks:
       - id: pyproject-fmt
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.243
+    rev: v0.0.249
     hooks:
       - id: ruff
         args: ["--fix"]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.0.0
+    rev: v1.0.1
     hooks:
       - id: mypy
         files: ^nox/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/tox-dev/pyproject-fmt
-    rev: "0.9.1"
+    rev: "0.7.0"
     hooks:
       - id: pyproject-fmt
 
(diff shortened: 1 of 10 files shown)


henryiii:
I can remove some of the puffed hooks too, I missed that.

FollowTheProcess:
Big thumbs up from me, love ruff. I've switched to it on a lot of personal projects.

The one thing I'd say though is it does get released a lot (not a bad thing at all) so I think if we're going to adopt it I'd like to see us also adopt pre-commit CI to always keep these things up to date for us 👍🏻

@theacodes Is this something you'd be on board with? Would need some
setup/permissions on the repo to go in this direction 🙂

henryiii:
While Ruff updates frequently, it's actually more stable than flake8, since it has zero dependencies and is just a binary. So the only forced reason to update would be for new platforms (like a new version of Python), but other than that, it should work indefinitely. Something like a dependency updating can't break it.

.pre-commit-config.yaml
```diff
@@ -82,13 +55,8 @@ repos:
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.10.0
     hooks:
-      - id: python-check-blanket-noqa
-      - id: python-check-blanket-type-ignore
-      - id: python-no-log-warn
-        exclude: ^tests/test_sessions.py$
```
henryiii, line 88: Ruff's implementation is smarter than a simple grep, so it can tell session.warn isn't logging's warn! :)

henryiii:
Looks like pyproject-fmt is broken, it just keeps adding an extra line of space between tool.ruff and previous entries. It's also removing space between project headings, which I don't like - I personally like one line between related sections, and two lines between unrelated sections (like tool.stuff sections). Looks like it's https://github.com/tox-dev/pyproject-fmt/issues/69 and will be fixed in https://github.com/tox-dev/pyproject-fmt/pull/70.

cjolowicz (approved):

FollowTheProcess:
The pre commit CI thing was never a blocker for this, just something to consider adopting going forward. Happy to see this come in 👍🏻

FollowTheProcess (approved):

DiddiLeija (approved):
> The pre commit CI thing was never a blocker for this, just something to consider adopting going forward. Happy to see this come in 👍🏻

In that case LGTM!

PR #696, opened 2023-03-05
docs: recipe for GHA generate

henryiii:
Follow up to #665 based on https://github.com/scikit-build/scikit-build-sample-projects/pull/31.

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -131,3 +131,34 @@ Even more so with a sprinkling of Nox:
         session.run("git", "push", "--tags", external=True)
 
 Now a simple ``nox -s release -- patch`` will automate your release (provided you have Bump2Version set up to change your files). This is especially powerful if you have a CI/CD pipeline set up!
+
+
+Generating a matrix with GitHub Actions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Nox knows what sessions it needs to run. Why not tell GitHub Actions what jobs to run dynamically? Using the ``--json`` flag and a bit of json processing, it's easy:
+
+.. code-block:: yaml
+
+    jobs:
+      generate-jobs:
+        runs-on: ubuntu-latest
+        outputs:
+          session: ${{ steps.set-matrix.outputs.session }}
+        steps:
+        - uses: actions/checkout@v3
+        - uses: wntrblm/nox@main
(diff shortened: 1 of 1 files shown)


docs/cookbook.rst
```diff
@@ -131,3 +131,34 @@ Even more so with a sprinkling of Nox:
...
+    jobs:
+      generate-jobs:
+        runs-on: ubuntu-latest
+        outputs:
+          session: ${{ steps.set-matrix.outputs.session }}
+        steps:
+        - uses: actions/checkout@v3
+        - uses: wntrblm/nox@main
+        - id: set-matrix
+          run: echo session=$(nox --json -l | jq -c '[.[].session]') | tee --append $GITHUB_OUTPUT
```
henryiii, line 152:
```suggestion
          run: echo session=$(nox --json -l | jq -c '[.[].session]') >> $GITHUB_OUTPUT
```

This can also be written this way, simpler, but harder to debug if something goes wrong.

FollowTheProcess:
Love this 👍🏻

FollowTheProcess (approved):

docs/cookbook.rst
cjolowicz: Can you explain why the `tee` version is easier to debug?

Without `set -o pipefail`, would this silently ignore errors from Nox?

henryiii: If you look into the step, then you can see what it's actually doing if you use `tee`.

Ahh, looks like there's some complication to pipefail. It's only set if you specify `shell: bash`, and not in the default shell (even if that is bash). https://github.com/actions/runner/issues/1955 - I think the best thing here would probably be to set `shell: bash`.

Though if nox does fail (which should be rare for a "list" action), then there should't be useful output, so later steps would fail, and being able to see that output would probably still be easier to debug by having it in the logs rather than having it placed in `$GITHUB_OUTPUT`.

henryiii:
```suggestion
          shell: bash
          run: echo session=$(nox --json -l | jq -c '[.[].session]') | tee --append $GITHUB_OUTPUT
```

cjolowicz (approved):

docs/cookbook.rst
henryiii: (Feel free to merge this if you like it better - I do)

PR #702, opened 2023-04-22
Don't run python 2.7 virtualenv tests for newer versions of virtualenv

### LINKED ISSUE #701: Address test breakage, virtualenv >= 20.22.0 , python 2 support removal
### Current Behavior

While verifying pr #700 via CI the python2 reuse test was failing. I replicated this locally (WSL Linux) and was able to get all tests to pass by restricting the virtualenv range. The test for 2.7 likely needs to be conditional on the version of virtualenv.  

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

crwilcox:
fixes #701 

Newer virtualenv versions have dropped py2.7 support.

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -476,7 +476,7 @@ def test_create_reuse_oldstyle_virtualenv_environment(make_one):
 
 @enable_staleness_check
 def test_inner_functions_reusing_venv(make_one):
-    venv, location = make_one(reuse_existing=True, interpreter="3.10")
+    venv, location = make_one(reuse_existing=True)
     venv.create()
 
     # Drop a venv-style pyvenv.cfg into the environment.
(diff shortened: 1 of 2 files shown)


alex:
Looks like it's still sad about coverage :-(

alex (approved):

tests/test_virtualenv.py
```diff
@@ -24,12 +24,15 @@
...
 import pytest
+import virtualenv
+from packaging import version
 
 import nox.virtualenv
 
 IS_WINDOWS = nox.virtualenv._SYSTEM == "Windows"
 HAS_CONDA = shutil.which("conda") is not None
 RAISE_ERROR = "RAISE_ERROR"
+VIRTUALENV_VERSION = virtualenv.version.version
```
stefanor, line 35: That's only available in recent versions of virtualenv, using `hatch-vcs`.

In older versions, the path was: `virtualenv.__version__` (which is still available today)

PR #712, opened 2023-05-14
action: use multiple interpreters features from 'actions/setup-python@v4'

mayeut:
The nox action can be simplified now that setup-python allows to setup multiple interpreters natively.

This also adds support for `~` version specifier (e.g. `~3.12.0-0`) which mostly should not be necessary as setup-python now supports `allow-prereleases` which transforms `3.12` into `~3.12.0-0` when searching for CPython 3.12.
The `~` specifier might be interesting for those who want a minimum patch version, e.g. `~3.11.1` <=> `>=3.11.1 <3.12`.

### DIFF
--- a/.github/action_helper.py
+++ b/.github/action_helper.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import sys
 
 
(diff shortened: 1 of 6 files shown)


theacodes:
Awesome, thanks for doing this!

PR #718, opened 2023-06-21
chore: save session name in `Func`

### LINKED ISSUE #717: Save session name in `Func`

The session name manually provided at `@nox.session(name="foo")` does get sent to `nox._decorators.Func(name="foo")`, but it's then ignored, and not saved to the instance.

Also, if the session name isn't manually provided, it'll be inferred from the function name, but then won't even be sent to `Func`.

https://github.com/wntrblm/nox/blob/720f1ea538bbf7ceb7fad3eafaf662437a7183e7/nox/_decorators.py#L60

https://github.com/wntrblm/nox/blob/720f1ea538bbf7ceb7fad3eafaf662437a7183e7/nox/registry.py#L88-L89

Saving the session name, be it manually provided or automatically inferred, would make it easier for other developers to extend Nox functionality with other libraries.

I'd like for the session name, regardless of whether it was manually provided, to be sent to `Func` and saved as an instance variable there.

_No response_

### Anything else?

I believe this is a simple and quick enough change, so I'll be sending a PR with a possible implementation.

MicaelJarniac:
Closes #717

### DIFF
--- a/nox/_decorators.py
+++ b/nox/_decorators.py
@@ -66,6 +66,7 @@ class Func(FunctionDecorator):
         self.func = func
         self.python = python
         self.reuse_venv = reuse_venv
+        self.name = name
         self.venv_backend = venv_backend
         self.venv_params = venv_params
         self.should_warn = should_warn or {}
--- a/nox/registry.py
+++ b/nox/registry.py
@@ -85,8 +85,11 @@ def session_decorator(
     if python is None:
         python = py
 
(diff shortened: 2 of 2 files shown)


MicaelJarniac:
This may or may not make sense, though. I'm not familiar with the codebase at all. It doesn't seem to break anything, at least.

theacodes (approved):
looks great. :)

PR #720, opened 2023-06-23
fix(types): improve typing

henryiii:
This fixes an issue with `os.PathLike[str]` being accepted but not present in the typing. It also uses `collections.abc` wherever possible instead of `typing` (can't fully change all of them until 3.9+ is supported, due to runtime requirements). It also uses `Sequence`/`Mapping`/`Iterable` for input parameters (you should be generic in input, specific in output) - doing so adds a few copy functions and improves safely a bit, fewer arguments are allowed to be mutated later this way). A few `TypeVars` were added - they can be replaced with native syntax in Python 3.12+.

Noticed in https://github.com/scikit-build/scikit-build-core/pull/408 where mypy found two real problems in my noxfile, and one false positive, which this fixes.

### DIFF
--- a/nox/_decorators.py
+++ b/nox/_decorators.py
@@ -18,7 +18,8 @@ import copy
 import functools
 import inspect
 import types
-from typing import TYPE_CHECKING, Any, Callable, Iterable, TypeVar, cast
+from collections.abc import Iterable, Mapping, Sequence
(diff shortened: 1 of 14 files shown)


theacodes (approved):
LGTM!

theacodes:
Patched up some conflicts, waiting for CI to merge

PR #722, opened 2023-07-03
chore: ruff moved to astral-sh

henryiii:
See https://github.com/scientific-python/cookie/pull/200.

Also bumped the version while touching it.


DiddiLeija (approved):
+1 from me!

theacodes (approved):

Issue #708, opened 2023-04-24
Tests fail when NO_COLOR=1 is set in the environment


mgorny:
### Current Behavior

We're running the test suite with `NO_COLOR` set in the containing environment to prevent tools from outputting ANSI color codes to build logs. However, this causes nox tests to fail, seemingly causing it to conflict with `--force-color` option used by a test:

```pytb
============================================================== FAILURES ===============================================================
_______________________________________________ test_main_color_from_isatty[True-True] ________________________________________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f7c74791b40>, isatty_value = True, expected = True

    @pytest.mark.parametrize(("isatty_value", "expected"), [(True, True), (False, False)])
    def test_main_color_from_isatty(monkeypatch, isatty_value, expected):
        monkeypatch.delenv("FORCE_COLOR", raising=False)
        monkeypatch.setattr(sys, "argv", [sys.executable])
        with mock.patch("nox.workflow.execute") as execute:
            execute.return_value = 0
            with mock.patch("sys.stdout.isatty") as isatty:
                isatty.return_value = isatty_value
    
                # Call the main function.
                with mock.patch.object(sys, "exit"):
                    nox.__main__.main()
    
                config = execute.call_args[1]["global_config"]
>               assert config.color == expected
E               AssertionError: assert False == True
E                +  where False = Namespace(help=None, version=None, list_sessions=None, json=None, sessions=None, pythons=None, keywords=None, tags=None, posargs=[], verbose=None, add_timestamp=None, default_venv_backend=None, force_venv_backend=None, no_venv=False, reuse_existing_virtualenvs=False, no_reuse_existing_virtualenvs=None, R=False, noxfile='noxfile.py', envdir=None, extra_pythons=None, force_pythons=None, stop_on_first_error=False, no_stop_on_first_error=None, error_on_missing_interpreters=False, no_error_on_missing_interpreters=None, error_on_external_run=False, no_error_on_external_run=None, install_only=None, no_install=False, report=None, non_interactive=None, nocolor=True, forcecolor=False, color=False, invoked_from='/tmp/portage/dev-python/nox-2023.04.22/temp/pytest-of-portage/pytest-0/test_main_noxfile_options_with1').color

[...]

PR #723, opened 2023-07-04
Change _color_finalizer behavior

stasfilin:
Issue: https://github.com/wntrblm/nox/issues/708

### Expected Behavior
`Tests passing. Ideally, I think --force-color should override NO_COLOR in the containing environment rather than collide with it.`

With this modification, if `--force-color` is provided, it will return `True` and **enforce colors**, even if `NO_COLOR` is set in the environment. If `--no-color` is provided or `NO_COLOR` is set in the environment, it will return `False` and **disable** colors unless `--force-color` is also provided.

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -177,7 +177,7 @@ def _color_finalizer(value: bool, args: argparse.Namespace) -> bool:
     if args.forcecolor:
         return True
 
-    if args.nocolor or 'NO_COLOR' in os.environ:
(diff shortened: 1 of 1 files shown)


stasfilin:
Added a new commit to fix the linter issue 

From 
```python
if args.nocolor or 'NO_COLOR' in os.environ:
```

to
```python
if args.nocolor or "NO_COLOR" in os.environ:
```

theacodes:
Looks good, thanks!

PR #726, opened 2023-07-13
Removed the problematic Python 2.7.18 version

stasfilin:
Removed the problematic Python 2.7.18 version, since Python 2.7 is no longer supported and the specific version 2.7.18 was not found.
I removed it from the python-versions list.

Reference: https://github.com/actions/setup-python/issues/672

image


stasfilin:
Reference: https://github.com/actions/setup-python/issues/672

theacodes:
Thanks, @stasfilin!

PR #730, opened 2023-07-24
feat: support for reuse_venv option

### LINKED ISSUE #488: [BUG] --no-reuse-existing-virtualenvs doesn't seem to work
**Describe the bug**

Adding `--no-reuse-existing-virtualenvs` does not seem to ignore the cache, and instead just reuses the existing environments anyway. Happened on the previous version, and then did a `brew upgrade`, and 2021.10.1 also has this problem. Unless I misunderstand this option, pretty sure it's broken.

**How to reproduce**

```python
@nox.session(reuse_venv=True)
def type(session):
    session.install('mypy')
    session.run('mypy')
```
(simplified slightly, should be unimportant)

```console
$ nox --no-reuse-existing-virtualenvs -s type
nox > Running session type
nox > Re-using existing virtual environment at .nox/type.
```

**Expected behavior**
The existing virtual environment should not have been reused. (It really was, not just visually; I was trying to remove an item from the install list, but it remained installed).

samypr100:
Closes #488

This PR introduces a new option `reuse_venv` at the CLI and nox.options level as discussed in #488. It takes 4 possible values, namely `no`, `yes`, `never`, and `always` as discussed in the issue.
The code that evaluates if a venv should be reused has been switched to use this new option, and backwards compatiblity with `reuse_existing_virtualenvs` is kept using appropriate `finalizer_func` and `merge_func` on `_options.py`.

I've added tests to test new functionality as well do backwards compatibility checks and hopefully 🤞 updated the docs properly. I also added a new fixture `generate_noxfile_options` that helps with more dynamic configuration of `resources/noxfile_options.py` used in the tests to reduce boilerplate.

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -148,6 +148,30 @@ def _envdir_merge_func(
     return command_args.envdir or noxfile_args.envdir or ".nox"
 
 
+def _reuse_venv_merge_func(
+    command_args: argparse.Namespace, noxfile_args: argparse.Namespace
+) -> str:
+    """Merge reuse_venv from command args and Noxfile while maintaining
+    backwards compatibility with reuse_existing_virtualenvs. Default is "no".
+
+    Args:
+        command_args (_option_set.Namespace): The options specified on the
+            command-line.
+        noxfile_Args (_option_set.Namespace): The options specified in the
+            Noxfile.
+    """
+    # back-compat scenario with no_reuse_existing_virtualenvs/reuse_existing_virtualenvs
+    if command_args.no_reuse_existing_virtualenvs:
+        return "no"
+    if (
+        command_args.reuse_existing_virtualenvs
+        or noxfile_args.reuse_existing_virtualenvs
+    ):
+        return "yes"
+    # regular option behavior
+    return command_args.reuse_venv or noxfile_args.reuse_venv or "no"
+
+
 def default_env_var_list_factory(env_var: str) -> Callable[[], list[str] | None]:
     """Looks at the env var to set the default value for a list of env vars.
 
@@ -199,9 +223,19 @@ def _force_pythons_finalizer(
 
 
 def _R_finalizer(value: bool, args: argparse.Namespace) -> bool:
-    """Propagate -R to --reuse-existing-virtualenvs and --no-install."""
+    """Propagate -R to --reuse-existing-virtualenvs and --no-install and --reuse-venv=yes."""
     if value:
+        args.reuse_venv = "yes"
         args.reuse_existing_virtualenvs = args.no_install = value
+    return value
+
+
+def _reuse_existing_virtualenvs_finalizer(
+    value: bool, args: argparse.Namespace
+) -> bool:
+    """Propagate --reuse-existing-virtualenvs to --reuse-venv=yes."""
+    if value:
+        args.reuse_venv = "yes"
     return value
 
 
@@ -414,6 +448,18 @@ options.add_options(
             " creating a venv. This is an alias for '--force-venv-backend none'."
         ),
     ),
+    _option_set.Option(
+        "reuse_venv",
+        "--reuse-venv",
+        group=options.groups["environment"],
+        noxfile=True,
+        merge_func=_reuse_venv_merge_func,
+        help=(
+            "Controls existing virtualenvs recreation. This is ``'no'`` by"
+            " default, but any of ``('yes', 'no', 'always', 'never')`` are accepted."
+        ),
+        choices=["yes", "no", "always", "never"],
+    ),
     *_option_set.make_flag_pair(
         "reuse_existing_virtualenvs",
         ("-r", "--reuse-existing-virtualenvs"),
@@ -422,7 +468,8 @@ options.add_options(
             "--no-reuse-existing-virtualenvs",
         ),
         group=options.groups["environment"],
-        help="Reuse existing virtualenvs instead of recreating them.",
+        help="This is an alias for '--reuse-venv=yes|no'.",
+        finalizer_func=_reuse_existing_virtualenvs_finalizer,
     ),
     _option_set.Option(
         "R",
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -27,6 +27,13 @@ import argcomplete
 from nox import _option_set
 from nox.tasks import discover_manifest, filter_manifest, load_nox_module
 
+if sys.version_info < (3, 8):  # pragma: no cover
+    from typing_extensions import Literal
+else:  # pragma: no cover
+    from typing import Literal
+
+ReuseVenvType = Literal["no", "yes", "never", "always"]
+
 """All of Nox's configuration options."""
 
 options = _option_set.OptionSet(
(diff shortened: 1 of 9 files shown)


nox/_options.py
```diff
@@ -146,6 +146,30 @@ def _envdir_merge_func(
     return command_args.envdir or noxfile_args.envdir or ".nox"
 
 
+def _reuse_venv_merge_func(
```
johnthagen, line 149: question (non-blocking): Would it be possible to use `Literal["no", "yes", "never", "always"]` to improve type checking/IDE behavior?

Is there any way to make it so that IDEs/mypy can type check a user's input to `reuse_venv=` before running Nox?

samypr100: I didn't include it since we're still supporting 3.7 in the tests. Is there an alternative here?

johnthagen: One option could be [`typing-extensions`](https://pypi.org/project/typing-extensions/):

```py
from typing_extensions import Literal
```

samypr100: Good catch, didn't notice `typing_extensions` was already installed

johnthagen: Yeah, this should have us covered

https://github.com/wntrblm/nox/blob/274df9be1b977d7f4a6323f2dcb7ace6e49bf0bd/pyproject.toml#L47

samypr100: See c9d6ad97c41ec02795552fb873330897527de40a

johnthagen:
@theacodes We're excited to get to use this feature, was wondering if you were able to review or your thoughts? Thanks!

johnthagen:
@FollowTheProcess Would you be willing to review this MR? Thanks.

tests/test_tasks.py
```diff
@@ -24,6 +24,7 @@
 from unittest import mock
 
 import pytest
+from test_main import generate_noxfile_options  # noqa: F401
```
edgarrmondragon, line 27: This is a fixture so you don't need to (and shouldn't) import it, I think

samypr100: Thanks @edgarrmondragon, seems I'd get a `fixture 'generate_noxfile_options' not found` error if I do that. I could have moved it to a `conftest.py` but since there wasn't one already I decided for this approach instead. Thoughts?

edgarrmondragon: Ah, yeah it's defined in a different module. It's probably fine to add a `conftest.py` with this fixture alone imo.

samypr100: @edgarrmondragon Moved to `conftest.py` in a64e632fba1bf497ddf4662b458e22763cbcaecc

nox/_options.py
```diff
@@ -23,6 +23,13 @@
 from nox import _option_set
 from nox.tasks import discover_manifest, filter_manifest, load_nox_module
 
+if sys.version_info < (3, 8):  # pragma: no cover
+    from typing_extensions import Literal
+else:  # pragma: no cover
+    from typing import Literal
```
edgarrmondragon, line 29:
```suggestion
if sys.version_info < (3, 8):
    from typing_extensions import Literal
else:
    from typing import Literal
```

I think you can remove the `pragma` comments here. Coverage for the two is stored in a separate file (`.coverage.3.x`) and combined by the `cover` session.

samypr100: Good point, I decided to keep it the same as the rest of the code (e.g. [command.py](https://github.com/wntrblm/nox/blob/main/nox/command.py#L27)) for consistency.

funkyfuture:
i only looked on the user docs and i'm wondering:

- could `always` and `never` be used for `nox.options.reuse_venv`?
- if so, do these values make any sense at that level / should they be disallowed to protect users from confusion?

thanks for the implementation!

funkyfuture:
the prose about the feature and its usage is fine, but as a user i think the section would aso benefit from a summary such as https://github.com/wntrblm/nox/issues/488#issuecomment-932443974 or a table.

henryiii (approved):

henryiii:
`--reuse-venv` is a bit longer then `--reuse` (and you always have to specify a value, too, so it's a little lengthy on the command line, but the current long option is `--reuse-existing-virtualenvs`. The option draws a parallel with `reuse_venv: bool = ...` in the sessions, so I think I like it, but wouldn't object to the shorter `--reuse` either, especially since there's not a short option for it. Anyone have thoughts? Otherwise, this looks good to me.

nox/sessions.py
```diff
@@ -795,6 +793,20 @@ def _create_venv(self) -> None:
...
 
+    def reuse_existing_venv(self) -> bool:
+        return any(
+            (
+                # forces every session to reuse its env
+                self.global_config.reuse_venv == "always",
+                # sessions marked True will always be reused unless never is specified
+                self.func.reuse_venv is True
+                and self.global_config.reuse_venv != "never",
+                # session marked False will never be reused unless always is specified
```
cjolowicz, line 804: This comment doesn't relate to the code below

nox/_options.py
```diff
@@ -422,7 +475,8 @@ def _tag_completer(
             "--no-reuse-existing-virtualenvs",
         ),
         group=options.groups["environment"],
-        help="Reuse existing virtualenvs instead of recreating them.",
+        help="This is an alias for '--reuse-venv=yes|no'.",
```
cjolowicz, line 478: I'd love to see this called out in the docs as well

cjolowicz:
Personally I like the more explicit `--reuse-venv`. There are other things one might "reuse" (wheels?). AFAIU `-r` is also a short option for `--reuse-venv=yes`.

nox/_options.py
samypr100: It should be in the docs, near the end of this statement in `config.rst`:

`* ``nox.options.reuse_existing_virtualenvs`` is equivalent to specifying :ref:`--reuse-existing-virtualenvs <opt-reuse-existing-virtualenvs>`. You can force this off by specifying ``--no-reuse-existing-virtualenvs`` during invocation. Alias of ``nox.options.reuse_venv``.`

nox/sessions.py
samypr100: I tried to clarify this function with even more documentation here 715bc9a63d2098968c176bb154407a06ddb890c3, let me know if it's better.

cjolowicz: This is super helpful

nox/_options.py
cjolowicz: The docs should state explicitly that `--reuse-venv=yes` is equivalent to `--reuse-existing-virtualenvs`.

I'm happy for this to land as-is though. We probably want to deprecate the longer version anyway.

(OT: It might also be a nice improvement to allow specifying `--reuse-venv` without a value, defaulting to `yes`.)

cjolowicz (approved):

nox/_options.py
samypr100: Agree, I didn't explicitly deprecate it on this PR but it can be done so in a follow up one. When that happens I think it makes sense to augment `--reuse-venv` to default to `yes` when specified. Note `-r` is a valid shorthand for `--reuse-venv=yes`

samypr100: > The docs should state explicitly that --reuse-venv=yes is equivalent to --reuse-existing-virtualenvs.

Done in fdf876c16f2605ad4319c0bc42f68108ad15ceb6

PR #732, opened 2023-08-01
docs: Fix and update "Instant Dev Environment" recipe

paduszyk:
This adds a **missing** import of `pathlib` module and modifies the way `Path` object is used:
- usually `Path` is imported directly to the namespace;
- slash operator of "`pathlib`-way" for defining paths.


stasfilin (approved):
LGTM

PR #736, opened 2023-08-10
chore: use 2x faster black mirror

henryiii:
~~Also bumping some pre-commit versions.~~


FollowTheProcess (approved):
LGTM

DiddiLeija (approved):

edgarrmondragon:
Would probably be good to go one step forward and use Ruff's formatter

henryiii:
Yeah, I played with that a bit. There are some format changes, since Ruff's formatter doesn't implement the --preview style (yet). I can try a PR now that #744 is in.

PR #745, opened 2023-11-01
chore: ruff-format

henryiii:
I did this in two steps, to illustrate that ruff-format matched black in non-preview style.

See https://github.com/wntrblm/nox/pull/736. Also https://github.com/scientific-python/cookie/pull/299, which I've been mentioning whenever I make a PR moving to ruff-format.

- style: remove --preview from black
- chore: move to ruff-format


FollowTheProcess (approved):
Knew this would be coming! Done this on most of my repos too 👍🏻

PR #749, opened 2023-11-16
Fixed Improper Method Call: Replaced `NotImplementedError`

fazledyn-or:
## Details
While triaging your project, our bug fixing tool generated the following message(s)-
> In file: [_parametrize.py](https://github.com/wntrblm/nox/blob/main/nox/_parametrize.py#L81), class: Param, there is a special method `__eq__` that raises a NotImplementedError. If a special method supporting a binary operation is not implemented it should return NotImplemented. On the other hand, NotImplementedError should be raised from abstract methods inside user defined base classes to indicate that derived classes should override those methods. iCR suggested that the special method `__eq__` should return NotImplemented instead of raising an exception. An example of how NotImplemented helps the interpreter support a binary operation is [here](https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations).

### Related Documentation
- [docs.python.org - NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented)
- [docs.python.org - NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError)

## Changes
- Replaced `NotImplementedError` with `NotImplemented`

## Previously Found & Fixed
- https://www.github.com/SciTools/iris/pull/5544
- https://www.github.com/cupy/cupy/pull/7900
- https://www.github.com/ethereum/web3.py/pull/3080

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/nox/_parametrize.py
+++ b/nox/_parametrize.py
@@ -78,7 +78,7 @@ class Param:
         elif isinstance(other, dict):
             return dict(zip(self.arg_names, self.args)) == other
 
-        raise NotImplementedError
+        return NotImplemented
 
 
 def _apply_param_specs(param_specs: Iterable[Param], f: Any) -> Any:


nox/_parametrize.py
```diff
@@ -78,7 +78,7 @@ def __eq__(self, other: object) -> bool:
         elif isinstance(other, dict):
             return dict(zip(self.arg_names, self.args)) == other
 
-        raise NotImplementedError
+        return NotImplemented
```
johnthagen, line 81: This doesn't seem correct. If nothing else, this doesn't type check correctly with `-> bool` return type for the function?

johnthagen (changes requested):

nox/_parametrize.py
edgarrmondragon: @johnthagen I think that's ok. Mypy accepts `NotImplemented` as a return value for functions annotated as `-> bool`. It's in fact a little too liberal: https://github.com/python/mypy/issues/363.

That said, @fazledyn-or you probably want to update the `tests.tests__parametrize.test_param_eq_fail` test to use `assert _parametrize.Param() != "a"` instead of `pytest.raises`.

fazledyn-or: @edgarrmondragon Thanks for the heads up.
Seems like you fixed it. Is there anything I can do ?

edgarrmondragon (approved):

johnthagen (approved):

Issue #751, opened 2023-12-08
Allow passing callable `venv_backend` for creation of session virtualenv


wpk-nist-gov:
### How would this feature be useful?

There are are number of open issues regarding specific details of creating session virtual environments (e.g., #680, #662, #260).  For example, the creation of conda virtual environments is often intimately linked to an `environment.yaml` file.  While dependencies can often be installed after the virtual environment creation, this can be limiting.  This is especially true for locked conda environments (e.g. from [conda-lock]) which combines creating the environment and installing all dependencies.  If a callable could be passed as a `venv_backend`, you could do something like the following:

```python
def create_conda_lock_env(
    location: str,
    interpreter: str | None = None,
    reuse_existing: bool = False,
    venv_params: Any = None,
) -> CondaEnv:
    venv = CondaEnv(
        location=location,
        interpreter=interpreter,
        reuse_existing=reuse_existing,
        venv_params=venv_params,
    )

    # assume venv_params contains conda-lock.yml file
    if isinstance(venv_params, str) and os.path.exists(venv_params):
        lock_file = venv_params
    else:
        raise ValueError("pass `venv_params={conda-lock-file}`")

    # Custom creating (based on CondaEnv.create)
    if not venv._clean_location():
        logger.debug(f"Re-using existing conda env at {venv.location_name}.")
        venv._reused = True

    else:
        cmd = ["conda-lock", "install", "--prefix", venv.location, lock_file]

        logger.info(
            f"Creating conda env in {venv.location_name} with conda-lock {lock_file}"
        )
        nox.command.run(cmd, silent=True, log=nox.options.verbose or False)

    return venv


# Note that it's on the end user/custom backend to make sure passing python=.... makes sense.
@nox.session(
    python="3.11",
    venv_backend=create_conda_lock_env,
    venv_params="py311-test-conda-lock.yml",
)
def test(session: nox.Session) -> None:
    ...

```

Similarly, you could use this approach to:

* Create environments using micromamba
* Customize python search path for python interpreters
* Use `conda env create -f environment.yaml`
* Make creation of development environment trivial

For the last one, you could do the following:

```python
from nox.virtualenv import VirtualEnv
def create_venv_override_location(
    location: str,
    interpreter: str | None = None,
    reuse_existing: bool = False,
    venv_params: Any = None,
) -> None:
    assert isinstance(venv_params, str), "supply location with venv_params"

    location = venv_params
    venv = VirtualEnv(
        location=location,
[...]

Issue #680, opened 2023-01-03
support micromamba as a backend


yarnabrina:
### How would this feature be useful?

I would like to be able to use `micromamba` along with `mamba`/`conda`/`venv`. It's a lot smaller/faster than `mamba` (that's why we use it in our Gitlab CI flows), and to my knowledge lacks no functionality that is required for `nox` backend purposes.

mforbes:
Here is a hacky strategy to do this now. The idea is to link your `micromamba` executable to `conda` somewhere on `PATH`.  Here I effectively do

```bash
ln <micromamba> build/bin/conda
export PATH="./build/bin;${PATH}"
```

Here is an example which uses `micromamba` to install the `fftw` libraries, then pip installs `pyfftw` and runs a test.

```python
# noxfile.py
import os
import subprocess

import nox

try:
    MICROMAMBA = subprocess.check_output(["type", "-p", "micromamba"]).strip().decode()
except:
    MICROMAMBA = None

if MICROMAMBA:
    BIN_DIR = os.path.join("build", "bin")
    CONDA = os.path.join(BIN_DIR, "conda")
    if not os.path.exists(CONDA):
        os.makedirs(BIN_DIR, exist_ok=True)
        os.link(MICROMAMBA, CONDA)
    os.environ["PATH"] = os.pathsep.join([BIN_DIR, os.environ["PATH"]])

# If on Mac OS X Silicon and you need to use Rosetta
os.environ["CONDA_SUBDIR"] = "osx-64"

args = dict(python=["3.7", "3.8", "3.9", "3.10", "3.11"])


@nox.session(venv_backend="conda", venv_params=["-c", "defaults"], **args)
def test(session):
    # The following fails with micromamba -- gets quoted to '"fftw>=0.3.10"'
    #session.conda_install("fftw>=0.3.10", channel="conda-forge")
    session.conda_install("fftw", channel="conda-forge")
    session.install("pyfftw>=0.13.1", "pytest")
    session.run("pytest", "tests")
```

I noticed one bug: I could not specify the version of `fftw`:

```
    session.conda_install("fftw>=0.3.10", channel="conda-forge")
```

This gets quoted as `'"fftw>=0.3.10"'` by `_dblquote_pkg_install_args`, and then `micromamba` thinks this is a package name.  It should instead be quoted to `"fftw>=0.3.10"`.  This will need to be fixed.

# Desired Features

[...]

PR #753, opened 2023-12-11
Allow passing callable venv_backend for creation of session virtualenv

wpk-nist-gov:
This addresses #751

This PR allows for callable `venv_backend` to `nox.session`.  This functionality can be used to, for example,

* Create conda environments from `environment.yaml` files during creation
* Easily override the virtual environment location

This is a general solution that should be application to a number of open issues.  

Main changes:

* Update `nox.sessions.SessionRunner._create_venv` to look for a callable.
* Included a section in `Customized virtual environment creation` in `docs/tutorial.rst`  and example usage 
in `docs/cookbook.rst`. 
* Changed the `conda_tests` session of `noxfile.py` to use this functionality. This now creates conda environments using
`conda env create -f ....` with environment files under `environment/py{py_version}-conda-test.yaml`.  These files are are created from `requirements-conda-test.txt`.  There is also a session to use conda-lock.
* Added (simple) tests of this functionality to `tests_session.py`.   Nothing too deep, but there is coverage.

Open issues:

* The conda examples refer to private variables (`SessionRunner._reused` and `SessionRunner._clean_location`).  Does it make sense to make these public? I think that this is "advanced" functionality, so I'm happy to go with it.
* The callable has the signature 

```python
def venv_backend(
    location: str,
    interpreter: str | None,
    reuse_existing: bool,
    venv_params: Any,
    runner: SessionRunner,
) -> CondaEnv/VirtualEnv/..."
...
```

The parameters where copied from the previous `SessionRunner._create_env` calls to `VirtualEnv`, etc, with the addition of `runner`.  I added this if someone wants to access the calling runner instance.  This might be seen as repetitive, as all these parameters are derived from `runner`.   

Please let me know how I can improve the PR.  Thank you!


wpk-nist-gov:
Any chance someone could take a look at this?

jayqi:
It was mentioned in #751 that this could be a solution for supporting micromamba (#680). It would be great if there were an example of that in the cookbook.

wpk-nist-gov:
Thank you for the suggestion @jayqi!  I added sessions `micromamba_tests` to `noxfile.py` and updated the cookbook example to use both `conda env create` and `micromamba create`.

cjolowicz:
I think this merits broader discussion because you can't use the callable without accessing undocumented Nox internals.

henryiii:
(I was thinking that, was getting worried about how much of nox's internals would be expected to be touched by users. #762 changed the internals a bit, and we might not have been able to do that if this was public)

wpk-nist-gov:
Thank you for taking a look at this PR! I agree that using private/undocumented internals is less than ideal. But I still think functionality like this would be awesome. Before #762, I was myself playing with using uv via the callable backend. I'd like to take a look at #762 to get a feel of what internals have changed. 

Would it make sense to propose making a minimal set of the private internals touched here public? I'd be happy to take a crack at that. 

Thanks again!

henryiii:
Let's wait till after the next release, then it might be interesting to discuss. A few more environment related changes are going in.

wpk-nist-gov:
I think I have a better solution, that will require less interaction with nox internals.  There are two things I'm trying to override with the callback.

1. venv location
2. venv creation command.

For (1), the obvious use case is using nox to manage development environments.  Something like the following:

```python

@nox.session(venv_location=".venv", ....)
def dev(session: nox.Session) -> None:
    ....
```

For (2), the use case is customizing the creation of conda environments.  Instead of trying to hook into the virtualenv creation, you could instead just ask for the `venv` class to do everything (check for env reuse, delete previous venv, etc) but **skip** actually creating the virtualenv.  In this case, it would be on the user to create the virtualenv in the session function.  Something like:

```python
@nox.session(venv_skip_create=True)
def test(session: nox.Session) -> None:

    if not session.virtualenv._reused:  # I'd like to make the "reused property public"
        session.run_always(
            "conda",
            "env",
            "create",
            "-f",
            f"py{session.python}-environment.yaml",
            "--prefix",
            session.virtualenv.location,  # might add session.location, session.location_name?
        )

    ...
```
where as noted, I'd like to make the `reused` property public, and perhaps expose some properties in `session.virtualenv` to `session` level.

So my question:  Should I make a new PR with the above functionality?  If so, should I split into 2 PR's or just one?

henryiii:
IMO, venv location makes sense. Being able to set it to something other than `.nox/{name}` would be useful. Maybe that would be a good thing to start with?

wpk-nist-gov:
Excellent!  I'll put a new PR together with that functionality.

PR #756, opened 2023-12-31
Allow `--force-python` on unparametrized sessions

### LINKED ISSUE #623: Allow running a session with a different Python interpreter

I normally run `nox -s test` to run my tests using the default Python version (the one nox is installed under, I guess?) This is my most common activity, and I'd like it to be straightforward. However, I sometimes want to run the tests for a specific Python version. I can do this by parametrising the Python version for the test session, but if I do that, `nox -s test` now means "run the test session for *all* Python versions.

Furthermore, I don't want to have to hard-code the list of Python versions. Doing that means I need to update the noxfile every time a new Python version comes out.

If a session does not have the Python version parametrised, `nox -s session --use-python 3.9` would run the given session with Python 3.9. Having `-P` as a short form for `--use-python` would be useful.

* Using `@session(python=None, "3.8", "3.9", "3.10")` looks like it might work, as it creates a session with no suffix. But that still runs all of the suffixed sessions, rather than just running the one session.
* `--force-pythons` sounds like it would work, but it doesn't, it just says "Python version selection caused no sessions to be selected".

### Anything else?

_No response_

chrysle:
Also adds the shorthand `-P` for the `--force-python` option.

Closes #623

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -433,6 +433,7 @@ options.add_options(
     ),
     _option_set.Option(
         "force_pythons",
+        "-P",
         "--force-pythons",
         "--force-python",
         group=options.groups["python"],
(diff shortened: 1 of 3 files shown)


theacodes (approved):

chrysle:
Thank you for reviewing!

PR #763, opened 2024-02-17
Added a virtualenv backend for using uv

alex:
This also causes session.install to use uv instead of pip.


alex:
Aaand I see that between when I started this and when I put the PR up, @henryiii did it too. Closing in favor of his PR (#762).

@henryiii you may find some of the test fixes or such useful.

PR #768, opened 2024-02-17
feat(action): do not depend on pipx in Nox action

mayeut:
pipx is being re-added in https://github.com/actions/runner-images/pull/9474

blocked by https://github.com/actions/runner-images/issues/9256 (`temp: don't use pipx` commit)
blocked by https://github.com/actions/setup-python/issues/808 (`temp: python 3.8 & 3.9 missing on macos-14` commit)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -72,4 +72,4 @@ runs:
         with open(os.environ["GITHUB_PATH"], "at") as f:
             f.write(f"{nox_bin_path}\n")
         print("::endgroup::")
-      shell: "${{ steps.localpython.outputs.python-path }}" -u
+      shell: "${{ steps.localpython.outputs.python-path }} -u"
--- a/action.yml
+++ b/action.yml
@@ -31,5 +31,48 @@ runs:
         allow-prereleases: true
 
     - name: "Install nox"
-      run: pipx install --python "${{ steps.allpython.outputs.python-path }}" '${{ github.action_path }}'
+      run: |
+        # Install nox
+        "${{ steps.allpython.outputs.python-path }}" -u << "EOF"
+        import os
+        import shutil
+        import sys
+        import venv
+
+        from pathlib import Path
+        from subprocess import run
+
+
+        class EnvBuilder(venv.EnvBuilder):
+            def __init__(self):
+                super().__init__()
+
+            def setup_scripts(self, context):
+                pass
+
+            def post_setup(self, context):
+                super().post_setup(context)
+                self.bin_path = Path(context.env_exe).parent
+                run([sys.executable, "-m", "pip", "--python", context.env_exe, "install", r"${{ github.action_path }}"], check=True)
+
+
+        print("::group::Install nox")
+        nox_bin_path = Path(r"${{ runner.temp }}") / "nox"
+        if nox_bin_path.exists():
+            shutil.rmtree(nox_bin_path)
+        venv_path = Path(r"${{ runner.temp }}") / "nox-venv"
+        if venv_path.exists():
+            shutil.rmtree(venv_path)
+        builder = EnvBuilder()
+        builder.create(venv_path)
+        nox_path = [path for path in builder.bin_path.glob("nox*") if path.stem == "nox"][0]
+        nox_bin_path.mkdir()
+        try:
+            os.symlink(nox_path, nox_bin_path / nox_path.name)
+        except OSError:
+            shutil.copyfile(nox_path, nox_bin_path / nox_path.name)
+        with open(os.environ["GITHUB_PATH"], "at") as f:
+            f.write(f"{nox_bin_path}\n")
+        print("::endgroup::")
+        EOF
       shell: bash
--- a/action.yml
+++ b/action.yml
@@ -72,4 +72,4 @@ runs:
         with open(os.environ["GITHUB_PATH"], "at") as f:
             f.write(f"{nox_bin_path}\n")
         print("::endgroup::")
-      shell: "${{ steps.localpython.outputs.python-path }} -u"
+      shell: "${{ steps.localpython.outputs.python-path }} -u {0}"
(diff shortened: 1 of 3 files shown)


henryiii:
Do we know if pipx is going to be re-added? We could use this mechanism (both here and in cibuildwheel) until it is. And I wouldn't be shocked if 3.8 was never added to GHA macOS AS, Python didn't officially support AS at that point, it was very experimental. The 3.8 that ships on (at least previous versions of) macOS AS was an Apple patched fork. I'm not sure why 3.9 is missing though, as that had pretty solid support for AS.

henryiii:
The matching PR was merged and released in cibuildwheel. @mayeut can we take this out of draft?

mayeut:
I marked this PR as ready for review as requested by @henryiii 

pipx is going to be re-added however this PR also deals with self-hosted runners on which pipx might not be available.
As for python 3.8/3.9, there's no timeline from upstream with issue/PR for a couple months.

cjolowicz (approved):
LGTM. Can we change the title to something like "do not depend on pipx on Nox action"?

henryiii:
AFAICT pipx is just awaiting deployment. I'm not nearly as worried about not using pipx if it's available on all the official runners - a self hosted runner could just add pipx. Maybe we should just add the tests when pipx becomes available? What do others think? I'm fine either way.

action.yml
```diff
@@ -31,5 +31,48 @@ runs:
...
+        nox_bin_path.mkdir()
+        try:
+            os.symlink(nox_path, nox_bin_path / nox_path.name)
+        except OSError:
+            shutil.copyfile(nox_path, nox_bin_path / nox_path.name)
+        with open(os.environ["GITHUB_PATH"], "at") as f:
+            f.write(f"{nox_bin_path}\n")
+        print("::endgroup::")
+        EOF
       shell: bash
```
henryiii, line 78: Does `"${{ steps.allpython.outputs.python-path }}" -u` work for the shell specification?

henryiii: And this should be `localpython`

action.yml
```diff
@@ -31,5 +31,45 @@ runs:
...
+        nox_path = [path for path in builder.bin_path.glob("nox*") if path.stem == "nox"][0]
+        nox_bin_path.mkdir()
+        try:
+            os.symlink(nox_path, nox_bin_path / nox_path.name)
+        except OSError:
+            shutil.copyfile(nox_path, nox_bin_path / nox_path.name)
+        with open(os.environ["GITHUB_PATH"], "at") as f:
+            f.write(f"{nox_bin_path}\n")
+        print("::endgroup::")
+      shell: "${{ steps.localpython.outputs.python-path }}" -u
```
henryiii, line 75:
```suggestion
      shell: "${{ steps.localpython.outputs.python-path }} -u"
```

action.yml
```diff
@@ -31,5 +31,45 @@ runs:
...
+        nox_path = [path for path in builder.bin_path.glob("nox*") if path.stem == "nox"][0]
+        nox_bin_path.mkdir()
+        try:
+            os.symlink(nox_path, nox_bin_path / nox_path.name)
+        except OSError:
+            shutil.copyfile(nox_path, nox_bin_path / nox_path.name)
+        with open(os.environ["GITHUB_PATH"], "at") as f:
+            f.write(f"{nox_bin_path}\n")
+        print("::endgroup::")
+      shell: "${{ steps.localpython.outputs.python-path }} -u"
```
henryiii, line 75:
```suggestion
      shell: "${{ steps.localpython.outputs.python-path }} -u {0}"
```

henryiii (approved):

action.yml
mayeut: I used the shell trick because you can't quote the python path (`shell: "'${{ steps.localpython.outputs.python-path }}' -u"`). Maybe I was being too paranoid over self-hosted runners getting spaces in their path and that should not really matter.

action.yml
```diff
@@ -31,5 +31,45 @@ runs:
...
+        nox_path = [path for path in builder.bin_path.glob("nox*") if path.stem == "nox"][0]
+        nox_bin_path.mkdir()
+        try:
+            os.symlink(nox_path, nox_bin_path / nox_path.name)
+        except OSError:
+            shutil.copyfile(nox_path, nox_bin_path / nox_path.name)
+        with open(os.environ["GITHUB_PATH"], "at") as f:
+            f.write(f"{nox_bin_path}\n")
+        print("::endgroup::")
+      shell: "${{ steps.localpython.outputs.python-path }} -u {0}"
```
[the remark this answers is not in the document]
mayeut: @henryiii, the last activated one is always 3.11:
https://github.com/wntrblm/nox/blob/efc17bb8f5db0948d960e5c9a2efc4f1d69f332d/.github/action_helper.py#L58-L66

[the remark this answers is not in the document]
henryiii: Couldn’t we use localpython?

action.yml
henryiii: Oh, okay. Worth reverting? Why can’t it be quoted, also?

action.yml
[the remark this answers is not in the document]
mayeut: at some point, and that's probably still the case, setup-python was allowed to delete some python install to install a more specific one: e.g. you have the latest 3.11.y pre-installed, but specify 3.11.x in the nox action then, 3.11.y is deleted and localpython won't work anymore.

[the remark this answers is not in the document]
henryiii: Ah, makes sense. Should we bump to 3.12 now?

action.yml
mayeut: It's not worth reverting. I'll try to link/open a relevant issue in the runner implementation. I guess they just split by space but do not use any shlex.split resulting in the quoted string being used as the real path and fail to run the executable.

PR #770, opened 2024-02-19
feat: rename run_always to run_install

### LINKED ISSUE #761: Support indicating that a command *is* an install command

`nox` supports `--no-install` and `-R` options for not running installation commands, usually for use when reusing existing environments, and to speed up development where the install process is often the long pole.

But there doesn't seem to be a good way to indicate when something run via `session.run` *is* an install command -- specifically, if you're installing something *not* via `pip` or `conda` etc, which you may want to skip in the same scenarios as the above.

Concretely -- given a Javascript project, one might run `(p)npm install` to install all the project dependencies, followed by some test runner, e.g. `jest`, all under a noxenv called `ui-tests`.

In this scenario it would be nice if `nox -s ui-tests -R` was able to *not* run the install command, but the existing `-R` of course has no way to know which command to skip.

Even doing this externally (in the noxfile) seems nontrivial, as it seems the way `session.install` does this is by accessing private internals, e.g.:

https://github.com/wntrblm/nox/blob/5c82dc553bd04ee017784fc16193da0b35a44ab6/nox/sessions.py#L629-L630

Add a `session.run_install(...)` command which runs a subprocess but assumes that subprocess is an installation command which will be skipped when using `-R` / `--no-install`.

Refactoring `session.install` and `session.conda_install` to call this to do their `no_install` check seems also like a nice minor internal refactor.

I personally don't like boolean flags (as a rule), but `session.run(..., is_install=True)` is an alternative for those who are less averse.

A further alternative is to make more things public API (e.g. whether no_install is active) so that someone can do this themselves in their noxfile, but given that `nox` itself has 2 places where it itself wants this, it seems like something generally useful to me.

### Anything else?

A concrete `noxfile` where I wanted this functionality is [here](https://github.com/bowtie-json-schema/bowtie/blob/5cdb622567b556815a54db59b4276e82e89aa1ce/noxfile.py#L333-L379).

henryiii:
Fix #761 by renaming `run_always` to `run_install`, with back-compat alias.

CI will fail until conda-forge's setuptools gets patched with the correct metadata.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -186,7 +186,7 @@ The ``--no-install`` option causes the following session methods to return early
 
 - :func:`session.install <nox.sessions.Session.install>`
 - :func:`session.conda_install <nox.sessions.Session.conda_install>`
-- :func:`session.run_always <nox.sessions.Session.run_always>`
+- :func:`session.run_instsall <nox.sessions.Session.run_install>`
 
 This option has no effect if the virtualenv is not being reused.
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -143,20 +143,20 @@ If your project is a Python package and you want to install it:
 
 In some cases such as Python binary extensions, your package may depend on
 code compiled outside of the Python ecosystem. To make sure a low-level
-dependency (e.g. ``libfoo``) is available during installation
+dependency (e.g. ``libfoo``) is available during installation:
 
 .. code-block:: python
 
     @nox.session
     def tests(session):
         ...
-        session.run_always(
+        session.run_install(
             "cmake", "-DCMAKE_BUILD_TYPE=Debug",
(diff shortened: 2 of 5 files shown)


docs/usage.rst
```diff
@@ -186,7 +186,7 @@ The ``--no-install`` option causes the following session methods to return early
 
 - :func:`session.install <nox.sessions.Session.install>`
 - :func:`session.conda_install <nox.sessions.Session.conda_install>`
-- :func:`session.run_always <nox.sessions.Session.run_always>`
+- :func:`session.run_instsall <nox.sessions.Session.run_install>`
```
henryiii, line 189:
```suggestion
- :func:`session.run_install <nox.sessions.Session.run_install>`
```

cjolowicz (approved):

PR #771, opened 2024-02-20
ci: fix coverage

henryiii:
Pulling out of #762. Updates CI.


henryiii:
I think rebasing on #778 will be good. This saves at least a minute per previously pair of runs.

theacodes (approved):

PR #772, opened 2024-02-20
ci: update to artifacts v2

henryiii:
This updates to artifacts v2 (upload/download v4).

Following suggestions from `pipx run sp-repo-review[cli] .` (available in WASM here: https://learn.scientific-python.org/development/guides/repo-review/?repo=wntrblm%2Fnox&branch=main).


henryiii:
I think rebasing on #778 will fix this. I think theres a data corruption bug if you combine files with the same name.

theacodes (approved):
Approved, pending CI passing.

PR #775, opened 2024-02-21
tests: support running when the Python launcher for UNIX is present

henryiii:
I can't run the tests locally and see coverage because it fails on one test - if you have `py` (the Python launcher for UNIX) installed, it breaks the assumptions of this test. Fixed and used print instead of noqa to make it easier to debug in the future.


FollowTheProcess (approved):
Thanks @henryiii 🙂

Issue #441, opened 2021-06-08
Issues with virtual environment reuse and multiple Python versions installed through `pyenv`


PureFunctor:
**Describe the bug**
G'day, Nox 2021.6.6 seems to fail reusing environments when it finds multiple Python interpreters installed by `pyenv`. Perhaps this is related to #428 ?

**How to reproduce**
Install multiple versions with `pyenv`
```sh
$ pyenv install 3.8.9
$ pyenv install 3.9.5
$ pyenv global 3.9.5 3.8.9
```
This populates `~/.pyenv/shims/` with the following with executable scripts e.g. `python`, `python3.9`, `python3.8`, etc. This folder is typically added to `PATH`.

Running the following session breaks virtual environment reuse:
```py
@nox.session(python="3.9")
def test(session):
    session.install("pytest", ".")
    session.run("pytest")
```

Installing/running `nox` using `poetry` or a manually-created `virtualenv` environment in a project seems to resolve this, however.

**Expected behavior**
Virtual environments are reused in the presence of `pyenv`-installed Python versions.

cjolowicz:
Can you verify that ˋvirtualenv -p. 3.9ˋ (or whichever Python version you were using with Nox) gives you the expected full Python version?

I have seen similar behavior, though my situation was slightly different from the one you described. To my knowledge, virtualenv does not invalidate its cache when you install a newer point release (X.Y.Z) of Python using pyenv. Its cache invalidation code only checks the mtime of the pyenv shim, which does not change.

Nox will refuse to reuse the environment because the interpreter does not have the expected version. (That's not to say that Nox can't support this case better.)

cjolowicz:
See https://github.com/pypa/virtualenv/issues/2130

PureFunctor:
> Can you verify that `virtualenv -p 3.9` (or whichever Python version you were using for your Nox session) gives you the expected full Python version?

I'm not entirely sure what steps to take but when I did `virtualenv -p 3.9` and inspected the generated `pyvenv.cfg` file inside of the environment directory, I found:
```
home = /usr
implementation = CPython
version_info = 3.9.5.final.0
virtualenv = 20.4.7
include-system-site-packages = false
base-prefix = /usr
base-exec-prefix = /usr
base-executable = /usr/bin/python3.9
```

As opposed to using `virtualenv -p 3.8` which generates:
```
home = /home/pure/.pyenv/versions/3.8.9
implementation = CPython
version_info = 3.8.9.final.0
[...]

PR #781, opened 2024-02-23
fix: rebuild env if making an incompatible change

henryiii:
This is a followup to #762. Switching environments is likely to be a bit more common, so let's handle it correctly. This has a key path comparison fix that I think was stopping this from being the default before, and instead available via a secret environment variable used in tests.

This is also now "smart" about switching between compatible backends. Switching from venv to virtualenv or back won't trigger a rebuild.

@slafs, could you check this?

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -36,7 +36,7 @@ RAISE_ERROR = "RAISE_ERROR"
 VIRTUALENV_VERSION = virtualenv.__version__
 
 has_uv = pytest.mark.skipif(not HAS_UV, reason="Missing uv command.")
-has_conda = pytest.mark.skipif(not HAS_UV, reason="Missing conda command.")
+has_conda = pytest.mark.skipif(not HAS_CONDA, reason="Missing conda command.")
 
 
 class TextProcessResult(NamedTuple):
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -474,6 +474,7 @@ def test_create_reuse_venv_environment(make_one):
     assert reused
 
 
+@pytest.mark.skipif(IS_WINDOWS, reason="Avoid 'No pyvenv.cfg file' error on Windows.")
 def test_create_reuse_oldstyle_virtualenv_environment(make_one):
     venv, location = make_one(reuse_existing=True)
     venv.create()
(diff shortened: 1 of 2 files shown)


henryiii:
Okay, enjoy! I think there are enough tests to be pretty safe if everything passes.

cjolowicz:
Didn't that environment variable also control whether we check the interpreter version before reuse? IIRC That logic was subtly broken, which is why we put it behind the environment variable in the first place, until someone could come fix it.

henryiii:
Yes, that’s the samefile fix above.

cjolowicz:
> Yes, that’s the os.path.samefile fix above. 

I'm not sure I understand how that addresses the issues mentioned in the comment linked above? Specifically, these two:

- #441
- https://github.com/pypa/virtualenv/issues/2130

These issues result in environments never being reused because Nox and virtualenv disagree on which interpreter should be selected. Nox does a `PATH` lookup, while virtualenv favors the interpreter it was installed on, and sometimes uses stale entries from its cache because it doesn't understand pyenv's shims.

henryiii:
> These issues result in environments never being reused because Nox and virtualenv disagree on which interpreter should be selected

I’ll check, but what I was seeing was that the simple comparison was always failing due to one being resolved differently than the other, but they were the same file. So it was always recreating for me, and the fix above fixed it. I have pyenv but not using it here, so might have missed that case.

We do require virtualenv 20+ now IIRC, so we could follow the suggestion there and use its discovery system, I think?

cjolowicz:
> (I can also just keep the interp check behind a flag)

Yeah, this way we can land the environment type check right away and profit :)

I'm all for tackling the interpreter check as well, but I feel it needs a bit of thought.

henryiii:
Unrelated, but we have a logging issue in the tests. On a test failure (possibly only in the correct spot) we get tons of logging file closed errors.

tests/test_virtualenv.py
```diff
@@ -35,6 +35,9 @@
 RAISE_ERROR = "RAISE_ERROR"
 VIRTUALENV_VERSION = virtualenv.__version__
 
+has_uv = pytest.mark.skipif(not HAS_UV, reason="Missing uv command.")
+has_conda = pytest.mark.skipif(not HAS_UV, reason="Missing conda command.")
```
cjolowicz, line 39:
```suggestion
has_conda = pytest.mark.skipif(not HAS_CONDA, reason="Missing conda command.")
```

cjolowicz (dismissed):
Besides the `has_conda` (see comment), this LGTM!

henryiii:
@cjolowicz added your suggestion, feel free to change the "changes requested". :)

PR #782, opened 2024-02-23
fix: always pull versions from metadata

### LINKED ISSUE #779: 'virtualenv.version' has no attribute 'version' in test_virtualenv.py
### Current Behavior

running the tests for 2023.4.22, I get this error:
```
______________________________________________________________ ERROR collecting tests/test_virtualenv.py ______________________________________________________________
tests/test_virtualenv.py:35: in <module>
    VIRTUALENV_VERSION = virtualenv.version.version
E   AttributeError: module 'virtualenv.version' has no attribute 'version'
```

virtualenv.version doesn't have 'version', but does have '\_\_version\_\_'

### Expected Behavior

_No response_

### Steps To Reproduce

```python
$ /sw/bin/python3.10
Python 3.10.4 (main, Feb 18 2023, 04:48:29) [Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import virtualenv
>>> print(virtualenv.version.version)
  File "<stdin>", line 1, in <module>
AttributeError: module 'virtualenv.version' has no attribute 'version'
>>> print(virtualenv.version.__version__)
20.13.0
>>> 
```

### Environment

```Markdown
- OS: macOS 10.14.6
- Python: 3.8 and 3.10
- Nox: 2023.4.22
```

### Anything else?

_No response_

henryiii:
Close #779. A random attribute can move around, but we can always ask for the metadata version.

### DIFF
--- a/nox/tox_to_nox.py
+++ b/nox/tox_to_nox.py
@@ -20,6 +20,7 @@ import argparse
 import os
 import pkgutil
 import re
+import sys
 from configparser import ConfigParser
 from pathlib import Path
 from subprocess import check_output
@@ -27,9 +28,15 @@ from typing import Any, Iterable
 
 import jinja2
 import tox.config
-from tox import __version__ as TOX_VERSION
 
-TOX4 = TOX_VERSION[0] == "4"
+if sys.version_info < (3, 8):
+    import importlib_metadata as metadata
+else:
+    from importlib import metadata
+
+TOX_VERSION = metadata.version("tox")
+
+TOX4 = int(TOX_VERSION.split(".")[0]) >= 4
 
 if TOX4:
     _TEMPLATE = jinja2.Template(
--- a/tests/test_tox_to_nox.py
+++ b/tests/test_tox_to_nox.py
@@ -18,11 +18,10 @@ import sys
 import textwrap
 
 import pytest
-from tox import __version__ as TOX_VERSION
 
 tox_to_nox = pytest.importorskip("nox.tox_to_nox")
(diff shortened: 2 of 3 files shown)


chrysle:
Have you already considered adding a compat module?

FollowTheProcess (approved):

henryiii:
Of course. https://learn.scientific-python.org/development/patterns/backports/

I’ll probably do that in the future.

henryiii:
@chrysle though, since I expect to see if we are okay dropping 3.7 after the next release, and basically every comparison I see is based on < 3.8, it might not be worth adding a compat module. I'll probably wait to see after next release.

chrysle:
> I'll probably wait to see after next release.

Sounds good. If you need some help, I'm available for hire ;-)

> https://learn.scientific-python.org/development/patterns/backports/

Good article, by the way.

edgarrmondragon:
fwiw there's still ~23% of Nox downloads on 3.7, though I guess users could just use an older Nox if they're stuck on an EOL Python



(Disregard the August downloads, they're incomplete)

henryiii:
> though I guess users could just use an older Nox if they're stuck on an EOL Python

Yep. Though that’s why I’d like to get this release out supporting 3.7, since we have already put the effort in. In general, it’s not always helpful to keep supporting EoL Pythons, as you can easily break something that would have worked if you just dropped the older Python. But the last release should be a good one. :)

PR #783, opened 2024-02-23
chore: cleanup Ruff a bit

henryiii:
Just a few minor cleanups. Ruff doesn't need the target version if you use PEP 621 metadata. Added the pylint checks (minus a few). Did just a bit of cleanup.


FollowTheProcess (approved):
LGTM, thanks @henryiii

PR #785, opened 2024-02-27
feat: added venv_location option to specify virtualenv location

wpk-nist-gov:
This PR adds the option `venv_location` to specify the location for virtualenv creation and a per session basis, and is a follow up to #753.

 The use case is quickly defining development environments:

```python
@nox.session(venv_location=".venv")
def dev(session):
    ....
```

### DIFF
--- a/nox/registry.py
+++ b/nox/registry.py
@@ -21,7 +21,7 @@ from collections.abc import Sequence
 from typing import Any, Callable, TypeVar, overload
 
 from ._decorators import Func
-from ._typing import Python
+from ._typing import Python, StrPath
 
 F = TypeVar("F", bound=Callable[..., Any])
 
@@ -42,7 +42,7 @@ def session_decorator(
     name: str | None = ...,
     venv_backend: Any | None = ...,
     venv_params: Any | None = ...,
-    venv_location: str | None = ...,
+    venv_location: StrPath | None = ...,
     tags: Sequence[str] | None = ...,
 ) -> Callable[[F], F]:
     ...
@@ -56,7 +56,7 @@ def session_decorator(
     name: str | None = None,
     venv_backend: Any | None = None,
     venv_params: Any | None = None,
-    venv_location: str | None = None,
+    venv_location: StrPath | None = None,
     tags: Sequence[str] | None = None,
 ) -> F | Callable[[F], F]:
     """Designate the decorated function as a session."""
--- a/nox/registry.py
+++ b/nox/registry.py
@@ -42,6 +42,7 @@ def session_decorator(
     name: str | None = ...,
     venv_backend: Any | None = ...,
     venv_params: Any | None = ...,
+    venv_location: str | None = ...,
     tags: Sequence[str] | None = ...,
 ) -> Callable[[F], F]:
     ...
@@ -55,6 +56,7 @@ def session_decorator(
     name: str | None = None,
     venv_backend: Any | None = None,
     venv_params: Any | None = None,
+    venv_location: str | None = None,
     tags: Sequence[str] | None = None,
 ) -> F | Callable[[F], F]:
     """Designate the decorated function as a session."""
@@ -74,6 +76,7 @@ def session_decorator(
             name=name,
             venv_backend=venv_backend,
             venv_params=venv_params,
+            venv_location=venv_location,
             tags=tags,
         )
 
@@ -88,7 +91,14 @@ def session_decorator(
 
     final_name = name or func.__name__
     fn = Func(
-        func, python, reuse_venv, final_name, venv_backend, venv_params, tags=tags
+        func,
+        python,
+        reuse_venv,
+        final_name,
+        venv_backend,
+        venv_params,
+        venv_location=venv_location,
+        tags=tags,
     )
     _REGISTRY[final_name] = fn
     return fn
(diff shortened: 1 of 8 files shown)


henryiii:
This could also be used by power users to have multiple sessions share a single venv. Nicely combines with better logic on venv reuse. :)

wpk-nist-gov:
Yes, I was thinking the same thing!   I've used a similar (not supported) feature in tox in the past.  I'll play around with this possible functionality...

Dr. William P. Krekelberg
National Institute of Standards and Technology
***@***.***
(301) 975-2457
________________________________

cjolowicz:
Does this allow arbitrary path traversal?

nox/registry.py
```diff
@@ -55,6 +56,7 @@ def session_decorator(
     name: str | None = None,
     venv_backend: Any | None = None,
     venv_params: Any | None = None,
+    venv_location: str | None = None,
```
henryiii, line 59: These should support Path too, I think.

henryiii:
> [wrap venv_location with os.path.expanuser](https://github.com/wntrblm/nox/pull/785/commits/ea8e444367e411d64f79dab459ba559c4a026375)

Do we do that anywhere else? Any reason not to let a user make it explicit? `Path.home()` would work, for example.

wpk-nist-gov:
The call to 'expanduser' was just to catch an edge case. Not necessarily needed. But it also will covert pathlib.Path objects to strings correctly. Most other paths in the code at some point get passed through 'os.path.join' or similar which typically ensures they are path-like strings.

cjolowicz:
That is what I meant, thanks. Absolute paths won't be very portable and may be surprising behavior when you run Nox on somebody else's project. Why do you think this should be possible?

How does the `venv_location` feature work for parameterized sessions including sessions with multiple Pythons?

cjolowicz:
Are we concerned about users trying something like `venv_location="tests"`? They might expect this to land under `.nox` but it would create files within their test suite.

henryiii:
How would this interact with parametrization? I could see it potentially being very useful to allow a parametrized test to reuse the same environment, but some cases that would be bad, especially if Python was parametrized over (the most common one!).

wpk-nist-gov:
> That is what I meant, thanks. Absolute paths won't be very portable and may be surprising behavior when you run Nox on somebody else's project. Why do you think this should be possible?

I'm fine restricting `venv_location` to be below current directory, if that's the consensus?  

> How does the `venv_location` feature work for parameterized sessions including sessions with multiple Pythons?

Combining with the comment from @henryiii, I like the idea of using `.format`, so you could use something like `venv_location="my-env-{python}"`.  But I'm hesitant to add this functionality in general.  In my opinion, you should use the default nox behavior for these cases. Maybe for now, raise an error if using `venv_location` with parametrize? I'm not sure how to do this...

> Are we concerned about users trying something like venv_location="tests"? They might expect this to land under .nox but it would create files within their test suite.

Yes, I agree this could be an issue.  How about a check like: if `venv_location` exists, it must contains `{venv_location}/pyvenv.cfg` or `{venv_location}/conda-meta`?

henryiii:
Does `{venv_location}` contain `pyvenv.cfg` on Windows? I'm not sure it does, for at least one of venv/virtualenv.

wpk-nist-gov:
> Does `{venv_location}` contain `pyvenv.cfg` on Windows? I'm not sure it does, for at least one of venv/virtualenv.

No clue. Is there a general way to test if a directory is venv/virtualenv/conda-env?

cjolowicz:
A pyvenv.cfg with a home key is always present in a virtual environment, CPython needs it to build sys.path. It doesn't depend on the tool that creates it.

henryiii:
That's what I thought, but some of the tests are disabled on Windows looking for it (and virutalenv < 20 didn't add one it seems from comments).

wpk-nist-gov:
> A pyvenv.cfg with a home key is always present in a virtual environment, CPython needs it to build sys.path. It doesn't depend on the tool that creates it or the platform.

In this case, should I add a check that, if `venv_location` exists, that it is a virtualenv?

cjolowicz:
> In this case, should I add a check that, if venv_location exists, that it is a virtualenv?

Yes, I think this is a good safeguard.

> I'm fine restricting venv_location to be below the current directory, if that's the consensus? I don't have strong feelings on the matter.

Either that or just requiring a relative path. I'm not opposed to relaxing this restriction later if people come up with good use cases.

> Maybe for now, raise an error if using venv_location with parametrize? I'm not sure how to do this...

Yes, I'd prefer this solution, also for sessions with multiple Pythons. Let's see some real world usage of this feature before we add something like templating to it. One approach would be to check this in `Manifest.make_session` (similar to how we warn about `backend="none"` but resulting in a hard error instead of a warning). Another approach would be to error out when we create the environment. I'm not sure which is best.

wpk-nist-gov:
The latest commit raises an error if trying to pass multiple pythons to a session with `venv_location`.  It also ignores (with warning) any `--extra-python` or `--force-python` flags, although I'm thinking it should handle the case where passing a single value of `--force-python`.  

TODO
- [ ] Make `venv_location` always relative.  And hopefully fix the bug in windows CI...
- [ ] Add check for venv before overwriting directory
- [ ] Proper handling of `--force-python` if single valued.
- [ ] Possible addition.  Add `--devenv` flag that the could be used to direct *any* session to a user defined `venv_location`

Issue #746, opened 2023-11-09
`session.run` does not fail gracefully when receiving a list as command to run


stefanv:
### Current Behavior

`session.run(["pytests", "mypkg"])` results in a traceback.

### Expected Behavior

One of the following options:

(a) The error is caught and a helpful message displayed.
(b) The list is correctly intepreted as a command.

### Steps To Reproduce

```python
import nox


@nox.session
def tests(session: nox.Session) -> None:
    session.run(["pytest", "foo"])
```

### Environment

```Markdown
- OS: Linux
- Python: 3.11
- Nox: 2023.4.22
```

### Anything else?

I would be happy to submit a PR to address the issue, once I know which solution is preferred.

henryiii:
IMO a) is fine, it keeps the syntax and type hints simple, and you simply need to add a single character (`*`) to convert if you already have a list.

PR #786, opened 2024-02-28
fix: warn user when first argument to session.run is a list

stefanv:
They probably meant to unpack the list.

Closes https://github.com/wntrblm/nox/issues/746

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -379,6 +379,10 @@ class Session:
         if not args:
             raise ValueError("At least one argument required to run().")
 
+        if len(args) == 1 and isinstance(args[0], list):
+            raise ValueError("First argument to `session.run` is a list. "
+                               "Did you mean to use `session.run(*args)`?")
+
         if self._runner.global_config.install_only:
             logger.info(f"Skipping {args[0]} run, as --install-only is set.")
             return None
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -379,7 +379,7 @@ class Session:
         if not args:
             raise ValueError("At least one argument required to run().")
 
-        if len(args) == 1 and isinstance(args[0], list):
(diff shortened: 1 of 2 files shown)


nox/sessions.py
```diff
@@ -379,6 +379,10 @@ def run(
         if not args:
             raise ValueError("At least one argument required to run().")
 
+        if len(args) == 1 and isinstance(args[0], list):
+            raise ValueError("First argument to `session.run` is a list. "
+                               "Did you mean to use `session.run(*args)`?")
```
henryiii, line 384:
```suggestion
            msg = "First argument to `session.run` is a list. Did you mean to use `session.run(*args)`?"
            raise ValueError(msg)
```

nox/sessions.py
```diff
@@ -379,6 +379,10 @@ def run(
         if not args:
             raise ValueError("At least one argument required to run().")
 
+        if len(args) == 1 and isinstance(args[0], list):
```
henryiii, line 382:
```suggestion
        if len(args) == 1 and isinstance(args[0], (list, tuple)):
```

I think it might make sense to check for tuple too?

nox/sessions.py
stefanv: Sure! Is the error message still OK as-is? That's why I originally removed tuple, because I didn't want to get into the weeds

henryiii: I thought of that, thought it probably was okay to leave. You could use Sequence if you wanted to be a little more general - "first argument to `session.run` is a Sequence. Did you mean...

stefanv: I accepted your suggestion as-is. Happy to make another update, if you prefer.

henryiii (approved):
Using "list" is fine.

stefanv:
Thank you @henryiii and @cjolowicz

PR #787, opened 2024-02-28
feat: venv backend fallback

henryiii:
This allows a user to specify `uv|virtualenv` to use uv only if it is available. Also works for `conda` and `mamba`. A few notes:

* A non-optional backend cannot be followed by an optional one; `virtualenv|uv` is invalid, for example.
* This is supported everywhere, but it's not really useful for the command line, so I didn't add it to the option system. It's most useful in a config file or maybe an environment variable.
* This also puts the environment selection into a dict, making it more reusable, and could allow power users (@wpk-nist-gov) to fiddle with adding new environment types. It is not considered supported/public, though.


theacodes (approved):

theacodes:
Looks great.

PR #795, opened 2024-03-05
fix: look for uv next to python if it's not on PATH

### LINKED ISSUE #791: uv executable not found when installed with `pipx install nox[uv]`
### Current Behavior

The following exception is raised when a global `uv` installation is not available.

```python
ValueError: No backends present, looked for ['uv'].
```

https://github.com/wntrblm/nox/blob/4c8e89164d90718312e4819a48243797924f55a9/nox/sessions.py#L790

The underlying cause seems to be `shutil.which("uv")` not being able to find uv installed alongside nox.

https://github.com/wntrblm/nox/blob/4c8e89164d90718312e4819a48243797924f55a9/nox/virtualenv.py#L563

### Expected Behavior

Installing `nox[uv]` is sufficient to use uv as a backend, without relying on a global uv installation.

### Steps To Reproduce

1. Install `nox[uv]`
2. Configure `uv` as a venv backend, for example by setting `nox.options.default_venv_backend = "uv"` in the Noxfile
3. Run any session

### Environment

```Markdown
- OS: macOS
- Python: 3.12
- Nox: 2024.3.2
```

### Anything else?

_No response_

cjolowicz:
Fixes #791

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -39,6 +39,24 @@ _BLACKLISTED_ENV_VARS = frozenset(
 _SYSTEM = platform.system()
 
 
+def find_uv() -> tuple[bool, str]:
+    # Look for uv in Nox's environment, to handle `pipx install nox[uv]`.
+    with contextlib.suppress(ImportError, FileNotFoundError):
+        from uv import find_uv_bin
+
+        return True, find_uv_bin()
+
+    # Fall back to PATH.
+    uv = shutil.which("uv")
+    if uv is not None:
+        return True, uv
+
+    return False, "uv"
+
+
+HAS_UV, UV = find_uv()
+
+
 class InterpreterNotFound(OSError):
     def __init__(self, interpreter: str) -> None:
         super().__init__(f"Python interpreter {interpreter} not found")
@@ -325,7 +343,7 @@ class VirtualEnv(ProcessEnv):
     """
 
     is_sandboxed = True
-    allowed_globals = ("uv",)
+    allowed_globals = (UV,)
 
     def __init__(
         self,
@@ -524,7 +542,7 @@ class VirtualEnv(ProcessEnv):
                 cmd.extend(["-p", self._resolved_interpreter])
         elif self.venv_backend == "uv":
             cmd = [
-                "uv",
+                UV,
                 "venv",
                 "-p",
                 self._resolved_interpreter if self.interpreter else sys.executable,
@@ -560,5 +578,5 @@ ALL_VENVS: dict[str, Callable[..., ProcessEnv]] = {
 OPTIONAL_VENVS = {
     "conda": shutil.which("conda") is not None,
     "mamba": shutil.which("mamba") is not None,
-    "uv": shutil.which("uv") is not None,
+    "uv": HAS_UV,
 }
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -19,6 +19,7 @@ import re
 import shutil
 import subprocess
 import sys
+import types
 from textwrap import dedent
 from typing import NamedTuple
 from unittest import mock
@@ -536,6 +537,32 @@ def test_create_reuse_uv_environment(make_one):
 
     # The environment is reused because it looks like a uv environment
     assert reused
+
+
+UV_IN_PIPX_VENV = "/home/user/.local/pipx/venvs/nox/bin/uv"
+
+
+@pytest.mark.parametrize(
+    ["which_result", "find_uv_bin_result", "expected"],
+    [
+        ("/usr/bin/uv", UV_IN_PIPX_VENV,   (True,  UV_IN_PIPX_VENV)),
+        ("/usr/bin/uv", FileNotFoundError, (True,  "/usr/bin/uv")),
+        (None,          UV_IN_PIPX_VENV,   (True,  UV_IN_PIPX_VENV)),
+        (None,          FileNotFoundError, (False, "uv")),
+    ],
+)  # fmt: skip
+def test_find_uv(monkeypatch, which_result, find_uv_bin_result, expected):
+    def find_uv_bin():
+        if find_uv_bin_result is FileNotFoundError:
+            raise FileNotFoundError
+        return find_uv_bin_result
+
+    monkeypatch.setattr(shutil, "which", lambda _: which_result)
+    monkeypatch.setitem(
+        sys.modules, "uv", types.SimpleNamespace(find_uv_bin=find_uv_bin)
+    )
+
+    assert nox.virtualenv.find_uv() == expected
 
 
 def test_create_reuse_venv_environment(make_one, monkeypatch):
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,6 +40,7 @@ repos:
           - jinja2
           - packaging
           - importlib_metadata
+          - uv
 
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -41,7 +41,7 @@ import nox.command
 import nox.virtualenv
 from nox._decorators import Func
 from nox.logger import logger
-from nox.virtualenv import CondaEnv, PassthroughEnv, ProcessEnv, VirtualEnv
+from nox.virtualenv import UV, CondaEnv, PassthroughEnv, ProcessEnv, VirtualEnv
 
 if TYPE_CHECKING:
     from nox.manifest import Manifest
@@ -656,7 +656,7 @@ class Session:
             kwargs["silent"] = True
 
         if isinstance(venv, VirtualEnv) and venv.venv_backend == "uv":
-            self._run("uv", "pip", "install", *args, external="error", **kwargs)
+            self._run(UV, "pip", "install", *args, external="error", **kwargs)
         else:
             self._run(
                 "python", "-m", "pip", "install", *args, external="error", **kwargs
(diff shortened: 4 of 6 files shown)


nox/virtualenv.py
```diff
@@ -39,6 +39,21 @@
 _SYSTEM = platform.system()
 
 
+def find_uv() -> str | None:
+    if uv := shutil.which("uv"):
```
cjolowicz, line 43: Can we drop Python 3.7? 😅

FollowTheProcess: I don't see why not IMO? It's well past EOL

nox/virtualenv.py
```diff
@@ -39,6 +39,22 @@
 _SYSTEM = platform.system()
 
 
+def find_uv() -> str | None:
+    uv = shutil.which("uv")
+    if uv is not None:
+        return uv
+
+    # Handle `pipx install nox[uv]` by looking for uv next to python.
+    uv = os.path.join(os.path.dirname(sys.executable), "uv")
```
layday, line 48: I guess this needs `.exe` on Windows. Can also just use `shutil.which("uv", path=os.path.dirname(sys.executable))` which tries a bunch of different suffixes on Windows. [This](https://github.com/astral-sh/uv/blob/0bc047866d35c3944c27e71a185fae44e7ce3ae1/python/uv/__init__.py#L8-L30) is what uv itself does.

nox/virtualenv.py
```diff
@@ -39,6 +39,22 @@
...
+    if os.path.exists(uv):
+        return uv
+
+    return None
+
+
+UV = find_uv()
```
henryiii, line 55: Should we make sure this path is on the path? I'd expect `run("uv", ...)` to always work if the uv backend is used?

cjolowicz: Yeah, I also wondered about this...

My first intuition here is that we don't want this. If the user wants to run `uv` as a non-external command, they can `session.install("uv")` and it'll just work.

The problem is you can't add uv to the PATH without also adding every other command from the environment in which Nox and uv are installed. That seems prone to cause issues.

nox/virtualenv.py
cjolowicz: Using `find_uv_bin` sounds perfect, thanks!

nox/virtualenv.py
layday: Ideally, the env shouldn't contain any Python dependencies not explicitly `session.install`ed into it, and that includes uv's Python distribution.

tests/test_virtualenv.py
```diff
@@ -538,6 +539,28 @@ def test_create_reuse_uv_environment(make_one):
     assert reused
 
 
+UV_IN_PIPX_VENV = "/home/user/.local/pipx/venvs/nox/bin/uv"
+
+
+@pytest.mark.parametrize(
+    ["which_result", "find_uv_bin_result", "expected"],
+    [
+        ("/usr/bin/uv", UV_IN_PIPX_VENV, "/usr/bin/uv"),
```
cjolowicz, line 548: Is this the desired behavior?

You might argue that Nox should prefer the `uv` in its own environment.

Or you might argue that we should prefer `PATH` lookup because it's the standard mechanism, and looking in the environment is just a fallback for an edge case.

We're taking the second stance here, but I don't have a strong opinion fwiw.

FollowTheProcess: I'd say principal of least surprise suggests the second is best IMO

henryiii: In scikit-build-core, I try to import CMake/ninja first before falling back on path search. For a Python library, using something coinstalled I think makes more sense. Otherwise you can’t pip install uv to get the latest uv if you have a system uv.

layday: Provided that the venv is activated, uv from the venv's bin would take precedence, so `pip install uv` would work. pipx is a bit of an edge case, and an argument could probably be made that it should expose its venv/bin folder to its exported scripts.

henryiii: Not sure why it should be assumed that it's active? Usually environments are designed to be used without being active (though there are some exceptions).

Right now, it might not seem to make much difference, but as uv gets older and pre-installed in more places, it's likely to be much more important to be able to pip install a modern uv. That's really important for cmake/ninja (more for cmake since it changes more).

cjolowicz:
Converting back to draft because this also requires handling in `session.install`.

tests/test_virtualenv.py
cjolowicz: I'm fine to reverse the order.

How do you actually upgrade uv in Nox's pipx environment? AFAIU `pipx upgrade nox` wouldn't touch dependencies, and `--include-injected` doesn't apply. So would you need to `pipx runpip nox install --upgrade uv`? This seems sufficiently obscure that most users would never upgrade uv. We may want to document this, if there's no better way.

cjolowicz: I just checked, and the only way I found to upgrade uv was the `runpip` method.

To be honest, I think we should drop the extra. I fear that most users would end up with an outdated uv. They're better off installing uv explicitly, like with `pipx install nox uv`. Then `pipx upgrade-all` will just work.

layday: If uv isn't declared as a dependency of nox, there shouldn't be an expectation that nox would find ux from anywhere that's not on `$PATH` - could simply drop the import?

cjolowicz: To be clear, here's what I'm proposing:

- close this PR (which means that nox will only look for uv on PATH)
- open a new PR that removes the `nox[uv]` extra

henryiii: We can place the minimum version we require in the extra. Then it will get upgraded if it’s too old for us. You can also use pipx’s reinstall all.

The extra is useful to make this work in the action.

nox/virtualenv.py
```diff
@@ -39,6 +39,23 @@
...
+def find_uv() -> tuple[bool, str]:
+    uv = shutil.which("uv")
+    if uv is not None:
+        return True, uv
+
+    # Look for uv in Nox's environment as well, to handle `pipx install nox[uv]`.
+    with contextlib.suppress(ImportError, FileNotFoundError):
+        from uv import find_uv_bin
+
+        return True, find_uv_bin()
```
henryiii, line 51:
```suggestion
    # Look for uv in Nox's environment first, such as for `pipx install nox[uv]`.
    with contextlib.suppress(ImportError, FileNotFoundError):
        from uv import find_uv_bin

        return True, find_uv_bin()

    # Otherwise, see if there's a copy on the path
    uv = shutil.which("uv")
    if uv is not None:
        return True, uv
```

This is the part that needs reversing?

nox/virtualenv.py
```diff
@@ -39,6 +39,23 @@
...
+
+    return False, "uv"
+
+
+HAS_UV, UV = find_uv()
```
henryiii, line 56: We could use a single variable, `UV`, and set it to None if UV wasn't found.

tests/test_virtualenv.py
cjolowicz: > We can place the minimum version we require in the extra. Then it will get upgraded if it’s too old for us.

I'm not worried about our requirements, but about those of our users when calling `session.install`. Also, Nox releases less frequently than uv, by far.

> You can also use pipx reinstall nox[uv].

That's not a good experience for users. People will use `pipx install nox[uv]` and `pipx upgrade-all`, and it shouldn't leave them with an outdated uv.

> Pipx run (which is almost exclusively what I use) reinstalls after a week.

I don't think the majority of users uses `pipx run nox`. For those injecting packages like `nox-poetry` that's not an option.

> The extra is useful to make this work in the action.

If there's another way to make this work in the action that doesn't cause problems for our users and support overhead for us, I'd say that's worth investigating.

nox/virtualenv.py
cjolowicz: I had this in an earlier version, but it's awkward because you have to use `UV or "uv"` in the subprocess calls.

nox/virtualenv.py
```diff
@@ -39,6 +39,23 @@
...
 
 
+def find_uv() -> tuple[bool, str]:
+    uv = shutil.which("uv")
+    if uv is not None:
+        return True, uv
+
+    # Look for uv in Nox's environment as well, to handle `pipx install nox[uv]`.
+    with contextlib.suppress(ImportError, FileNotFoundError):
+        from uv import find_uv_bin
```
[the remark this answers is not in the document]
cjolowicz: Fixed

nox/virtualenv.py
cjolowicz: Yes. Holding back on this one while we're discussing the "stale uv" issue.

tests/test_virtualenv.py
layday: Related: https://github.com/pypa/pipx/issues/925

layday: It looks like this might be very easily fixable from pipx's side (see linked issue).

nox/virtualenv.py
henryiii: Ahh, so UV is always a string, and set to "uv" if not found? Okay. I though it was already None in that case.

nox/virtualenv.py
cjolowicz: You have convinced me :) Let's hope this does get fixed in pipx.

tests/test_virtualenv.py
cjolowicz: Ok I'm fine to go ahead with this here.

henryiii (approved):

PR #799, opened 2024-03-06
feat: extras for the action

### LINKED ISSUE #797: Support extras in the action

Currently, the action doesn't install uv. There's not a way to inject it, either.

I think the most general solution would be to support extras. So you could specify `extras: uv` or `extras: tox-to-nox` in the action inputs.

I thought about a `uv: true` setting, but I think the extras method is more general, also supports the other existing extra, and could be used in the future.

I also thought about adding uv by default, but that would limit the action to only binary-supported platforms (which is all the default runners, admittedly, but might not cover some self-hosted runners, perhaps? It's also a tiny extra install for users who aren't using it.

### Anything else?

Might be best to make sure this is in #768, as I think we need that for the action. This will very likely need the fix for #791 too.

henryiii:
Close #797. Requires #795.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -42,4 +42,5 @@ jobs:
       - uses: ./
         with:
           python-versions: "3.4, 3.5, 3.6, 3.7, 3.8, 3.9, 3.10, 3.11, 3.12, pypy-2.7, pypy-3.7, pypy-3.8, pypy-3.9, pypy-3.10"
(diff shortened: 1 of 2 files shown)


henryiii:
IMO, once can use:

```yaml
    - name: Setup uv
      uses: yezz123/setup-uv@v4
```

And get pip, so for now, I think that's probably fine, keeps the action simple.

PR #805, opened 2024-03-29
ci: use uv

henryiii:
Use uv for most (but not all) the jobs in CI.


cjolowicz (approved):
LGTM. Any particular reason to keep the lint job on pip?

henryiii:
uv doesn't support targeting Python 3.7 officially, and an warning will be produced in the future it seems: https://github.com/astral-sh/uv/pull/2757, so probably should disable uv entirely for the 3.7 jobs.

henryiii:
For the lint job, pre-commit will never support uv, so I went ahead and left it alone (also to have some non-uv jobs, 3.7 will be removed at some point).

PR #807, opened 2024-04-04
feat: add micromamba support

### LINKED ISSUE #680: support micromamba as a backend

henryiii:
Adds micromamba support alongside the existing conda and mamba support. Micromamba is a version of mamba using C++ completely, not requiring Python. It's very useful for CI systems and anywhere a single binary is easier to work with.

Fix #680.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -554,7 +554,9 @@ class Session:
 
         You can specify a conda channel using `channel=`; a falsey value will
         not change the current channels. You can specify a list of channels if
-        needed.
+        needed. It is highly recommended to specify this; micromamba does not
+        set default channels, and default channels vary for conda. Note that
+        "defaults" is also not permissivly licenced like "conda-forge" is.
 
         Additional keyword args are the same as for :meth:`run`.
 
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -236,7 +236,7 @@ class CondaEnv(ProcessEnv):
     """
 
     is_sandboxed = True
-    allowed_globals = ("conda", "mamba")
+    allowed_globals = ("conda", "mamba", "micromamba")
 
     def __init__(
         self,
@@ -305,6 +305,9 @@ class CondaEnv(ProcessEnv):
             return False
 
         cmd = [self.conda_cmd, "create", "--yes", "--prefix", self.location]
+        if self.conda_cmd == "micromamba":
+            # Micromamba doesn't have any default channels
+            cmd.append("--channel=conda-forge")
 
         cmd.extend(self.venv_params)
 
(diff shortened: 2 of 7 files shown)


tests/test_virtualenv.py
```diff
@@ -483,6 +483,24 @@ def test_reuse_conda_environment(make_one):
     assert reused
 
 
+@has_conda
+def test_micromamba_environment(make_one, monkeypatch):
+    conda_path = shutil.which("conda")
+    which = shutil.which
+    monkeypatch.setattr(
```
cjolowicz, line 490: Are we doing this so you can run the test if you don't have micromamba installed?

cjolowicz (approved):

tests/test_virtualenv.py
henryiii: Yes. It's easy to install but this is even easier to get proper coverage.

(Might be nice for more of the tests to just check the commands issued instead of requiring installs)

cjolowicz: Thanks for clarifying

henryiii: Added a comment.

nox/sessions.py
```diff
@@ -554,7 +554,9 @@ def conda_install(
 
         You can specify a conda channel using `channel=`; a falsey value will
         not change the current channels. You can specify a list of channels if
-        needed.
+        needed. It is highly recommended to specify this; micromamba does not
+        set default channels, and default channels vary for conda. Note that
+        "defaults" is also not permissivly licenced like "conda-forge" is.
```
jezdez, line 559: @henryiii This sounds ominous, Anaconda's `defaults` channel is commercially licensed and is free to use for educational users and companies with fewer than 200 employees.

nox/virtualenv.py
```diff
@@ -305,6 +305,12 @@ def create(self) -> bool:
             return False
 
         cmd = [self.conda_cmd, "create", "--yes", "--prefix", self.location]
+        if self.conda_cmd == "micromamba" and not any(
+            v.startswith(("--channel=", "-c")) or v == "--channel"
+            for v in self.venv_params
+        ):
+            # Micromamba doesn't have any default channels
+            cmd.append("--channel=conda-forge")
```
jezdez, line 313: It might be worth mentioning this default in the docs as it is not quite the same as relying on PyPI (which doesn't re-package software from 3rd party authors as conda-forge does). Linking to the [conda-forge security considerations](https://conda-forge.org/docs/maintainer/knowledge_base/#security-considerations-for-conda-forge-builds) would be even better to inform users about that.

Issue #808, opened 2024-04-06
Support for PEP 723


henryiii:
### How would this feature be useful?

I'd love to have good support for PEP 723 in nox. As a reminder, scripts can start with this:

```python
# /// script
# dependencies = ["requests"]
# requires-python = ">=3.10"
# ///
```

(I wrote a post about it a while back, see [here](https://peps.python.org/pep-0723/). I've started adding this to helper scripts; but most of the time I also have nox sessions that run the helper scripts, and it's nice to have then all in once place and show up with `nox -l`, and I'd like to keep the canonical dependency list inside the script.

Now, being written in Python, you _could_ implement this yourself in a noxfile - except if nox runs on Python < 3.11, no TOML library is installed (this hits me a lot - since I also often need to pull dependency lists from pyproject.toml, too, which can't be done since nox doesn't provide `tomli` on Python < 3.11 - I've really wanted it to for a while now).

### Describe the solution you'd like

I thought about a lot of ways, and there is a simple solution that could be built upon later, but provides a lot of flexibility and utility for now. A new function could be added, `nox.project.load_toml ` (happy to bike shed on name). This would take a path to either a toml file (which it would load) or a python file, which it would load the `/// script` block. The example above then would be:

```python
@nox.session
def myscript(session: nox.Session) -> None:
    deps = nox.project.load_toml("myscript.py")["dependencies"]
    nox.install(*deps)
    nox.run("python", "myscript.py", *session.posargs)
```

And you could use it to for example install a project's dependencies without building the project:

```python
@nox.session
def prepare(session: nox.Session) -> None:
    deps = nox.project.load_toml("pyproject.toml")["project"]["dependencies"]
    nox.install(*deps)
    ...
```

(you could modify this for Poetry, as well) You _can't_ do this today unless you assume nox will always run on Python 3.11 or above.

You could also access `requires-python` too, though nox can't do much with it yet. Might be a future addition, though, to allow Python to be specified as a SpecifierSet.

### Describe alternatives you've considered

I've thought about a lot of different ways, and finally arrived on the one above. A few of them were:

---

We could provide a way to turn a script into a session. This might look like this:

```python
nox.script_to_session("myscript.py", name="my-script")
# Docstring could become the docs for the session, too!
```

The problem is that this is quite rigid; if you want to add anything, like logging or something, you can't easily make this a normal session without something like proposed above. This could come later, but wouldn't be a good starting point.

---

[...]

PR #811, opened 2024-04-07
feat: support PEP 723 with a toml load function

henryiii:
This is the suggestion from #808. As detailed there, I think this is a flexible first step that will help noxfile authors support PEP 723 in scripts, and could have a followup later that provides a more integrated solution.

This also allows noxfile authors to finally integrate with pyproject.toml on all versions of Python, rather than requiring nox be installed on Python 3.11+, by ensuring that a toml library is always present. You might need to read the version, etc. from `pyproject.toml` in your noxfile, and now you can.

I made all the errors ValueError because that's what the PEP did in the example implementation. Happy to make them something else if there's something better.

### DIFF
--- a/nox/__init__.py
+++ b/nox/__init__.py
@@ -14,7 +14,7 @@
 
 from __future__ import annotations
 
-from nox import toml
+from nox import project
 from nox._options import noxfile_options as options
 from nox._parametrize import Param as param
 from nox._parametrize import parametrize_decorator as parametrize
@@ -30,5 +30,5 @@ __all__ = [
     "session",
     "options",
     "Session",
-    "toml",
+    "project",
 ]
--- a/nox/__init__.py
+++ b/nox/__init__.py
@@ -17,9 +17,18 @@ from __future__ import annotations
 from nox._options import noxfile_options as options
 from nox._parametrize import Param as param
 from nox._parametrize import parametrize_decorator as parametrize
+from nox._toml import load_toml
 from nox.registry import session_decorator as session
 from nox.sessions import Session
 
 needs_version: str | None = None
 
-__all__ = ["needs_version", "parametrize", "param", "session", "options", "Session"]
+__all__ = [
+    "needs_version",
+    "parametrize",
+    "param",
+    "session",
+    "options",
+    "Session",
+    "load_toml",
+]
--- a/nox/__init__.py
+++ b/nox/__init__.py
@@ -14,10 +14,10 @@
 
 from __future__ import annotations
 
+from nox import toml
 from nox._options import noxfile_options as options
 from nox._parametrize import Param as param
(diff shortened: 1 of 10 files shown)


cjolowicz:
Should we fail the session instead of raising ValueError? Nox can be more opinionated about the error handling than say a library, and it would avoid the traceback.

henryiii:
Hmm, we could, but most of the improperly configured calls seem to also raise ValueError, such as 

https://github.com/wntrblm/nox/blob/956f10c5691b4f4d1a7e2d648fa2ca2eb4b27ab5/nox/sessions.py#L219

and

https://github.com/wntrblm/nox/blob/956f10c5691b4f4d1a7e2d648fa2ca2eb4b27ab5/nox/sessions.py#L388-L393

nox/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+from nox import toml
```
henryiii, line 17: Should we support `nox.toml.load` out of the box, or should it require `import nox.toml`? The latter is a bit more efficient, since you only load it if you need it (but we could make this lazy). `rich` and `textual` go this way.

cjolowicz: I'd go with convenience over efficiency here, I don't think the extra import will hurt anyone in the context of a Nox run.

henryiii:
Current name (from the issue) is `nox.project.load_toml`. `nox.project` is imported by default.

cjolowicz (approved):
Looks great!

Do we want to mark this as a provisional API in some way?

It feels like new territory. We may want to adapt to evolving packaging standards and move around or rename things without going through deprecation.

henryiii:
> Do we want to mark this as a provisional API in some way? ...

I think what we've added here is pretty conservative. `nox.project` is a place for project related tools. `load_toml` does exactly that, it loads toml and nothing more. (It does support PEP 723, but that's accepted so it isn't allowed to change significantly without a new PEP). We've left it up to the user to handle converting pyproject.toml's and script blocks into useable information.

If we add more things here, then possibly. If we have a `session_from_script` helper or something, that could be provisional.

(But happy to call this provisional too if you prefer)

requirements-test.txt
```diff
@@ -1,4 +1,4 @@
-coverage[toml]>=5.3
+coverage[toml]>=7.2
```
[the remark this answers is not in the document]
henryiii: Oops, doesn't support Python 3.7.

requirements-test.txt
```diff
@@ -1,4 +1,4 @@
-coverage[toml]>=5.3
+coverage[toml]>=7.3
```
henryiii, line 1:
```suggestion
coverage[toml]>=7.2
```

PR #812, opened 2024-04-09
feat: support None to remove envvars

henryiii:
Followup to #804. Setting `env={"NOCOLOR": None}` will unset the value if it exists in the outer environment.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -283,7 +283,7 @@ class Session:
     def run(
         self,
         *args: str | os.PathLike[str],
-        env: Mapping[str, str] | None = None,
+        env: Mapping[str, str | None] | None = None,
         include_outer_env: bool = True,
         **kwargs: Any,
     ) -> Any | None:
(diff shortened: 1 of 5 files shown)


henryiii:
I got a thumbs up, anyone want to do a review? Nice to have an approval. ;)

cjolowicz (approved):

edgarrmondragon:
Should there be a documentation update for this?

henryiii:
I did add it to the docs. It's only present in nox/sessions.py, but that's the only place `env=` is documented. It shows up in the docs page, though.

nox/sessions.py
```diff
@@ -350,7 +350,9 @@ def run(
             print("Current Git commit is", out.strip())
 
         :param env: A dictionary of environment variables to expose to the
-            command. By default, all environment variables are passed.
+            command. By default, all environment variables are passed. You
+            can block an environment variable from the outer environment by
+            setting it to None.
```
henryiii, line 355: Docs update here.

edgarrmondragon: Ah, I do see it in https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.run. Sorry about the noise.

PR #818, opened 2024-04-12
fix: ensure 'uv' always works in a uv venv

henryiii:
It's necessary sometimes to use the backend tool (pip, uv, conda, mamba, micromamba) inside a virtual environment. We already have a bit of special casing for it, making sure it's not considered "external". This PR just makes sure that `.run("uv", ...)` is always supported if you are using the uv backend. All the other backends are already gaurenteed to be on the path if they are active, so no change is needed (yet, anyway - this mechanism could also enable using `pip` from outside instead of always installing it, might be worth investigating in the future).

Followup to and generalizes #795.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -557,8 +557,12 @@ class Session:
             return self._run_func(args[0], args[1:])  # type: ignore[unreachable]
 
         # Using `"uv"` when `uv` is the backend is guaranteed to work, even if it was co-installed with nox.
-        if self.virtualenv.venv_backend == "uv" and args[0] == "uv" and nox.virtualenv.UV != "uv":
-            args = (nox.virualenv.UV, *args[1:])
+        if (
+            self.virtualenv.venv_backend == "uv"
+            and args[0] == "uv"
+            and nox.virtualenv.UV != "uv"
+        ):
+            args = (nox.virtualenv.UV, *args[1:])
 
         # Combine the env argument with our virtualenv's env vars.
         if include_outer_env:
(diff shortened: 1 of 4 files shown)


edgarrmondragon (approved):

cjolowicz (approved):

henryiii:
Added handling for one more corner case: if a user installs uv into a uv environment, they'd probably expect the new uv to be used after they installed it, just like pip is for pip environments. Tested locally with:

```python
@nox.session(venv_backend="uv")
def uv(session):
    session.install("uv")
    session.install("pip")
```

cjolowicz:
Can you add a source code comment to explain this edge case?

It would also be good to have a comment for the `Path.samefile` call in `find_uv`, and to add both branches to the test matrix in `test_find_uv`.

henryiii:
Added comments and added to tests.

edgarrmondragon:
@henryiii wdyt of also not warning when `uvx` is used in uv venv?

edgarrmondragon:
> Seems good. Though you can always use `"uv", "tool", "run"` instead. I can see if I can make a PR.

Ah `uv tool run` works in the meantime, though I hate typing 8 more characters! 😅 

> Could you check my other two open PRs? I'd like a set of eyes on them before merging, and I'd like to move toward a release soonish. :)

Done they LGTM, though mind you I don't have write access here...

henryiii:
Yes, the approval shows up grey, that's fine, it's not a hard requirement and I do have write access. :)

PR #822, opened 2024-04-26
chore: drop Python 3.7

henryiii:
Dropping support for running with Python 3.7, which is well past EoL.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -40,17 +40,13 @@ if shutil.which("micromamba"):
     nox.options.sessions.append("micromamba_tests")
 
 
-# Because there is a dependency conflict between argcomplete and the latest tox
-# (both depend on a different version of importlibmetadata for Py 3.7) pip
-# installs tox 3 as the latest one for Py 3.7.
 @nox.session
 @nox.parametrize(
     "python, tox_version",
     [
         (python, tox_version)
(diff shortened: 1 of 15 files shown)


noxfile.py
```diff
@@ -202,13 +198,11 @@ def github_actions_default_tests(session: nox.Session) -> None:
...
 @nox.session(
     python=[
-        "3.7",
         "3.8",
         "3.9",
         "3.10",
         "3.11",
         "3.12",
-        "pypy3.7",
         "pypy3.8",
```
edgarrmondragon, line 206: This is also EOL fwiw. Last released was on 2022-12-29.

henryiii: Ahh, I thought I got PyPy everywhere. It's a bit tricky to grep since we support targeting older Pythons.

Technically, pypy3.8 is EoL too. pypy2.7, pypy3.9, and pypy3.10 are the only supported ones. But I think matching CPython is fine.

henryiii: Ah, you mean the line below, pypy3.8. Yes, but we've always just matched CPython. It's not extra effort to support (in fact, you can't split them up in requires-python IIRC)

cjolowicz (approved):
🎉

Issue #302, opened 2020-03-28
Dry run for nox


maciej-lech:
Hello.

I have been using `nox` for a while now and I would love to see a possibility to dry run `nox`; the feature which exists in many different shell tools e.g. `rsync`, `git clean`, etc.

**How would this feature be useful?**

For me personally, this future will be useful mostly in two areas:
- During the development of `noxfile.py`, especially when the same `noxfile.py` is used for many different python distributions and the logic is more complex.
- For automated verification of CI/CD workflows using `nox` with e.g. dynamically created `nox` parameter to filter the sessions. It will allow us to create tests that can verify if `nox` works as expected without really need to run it.

**Describe the solution you'd like**
Add a new parameter `-n, --dry-run` to the `nox`.

```
$ nox --dry-run -s tests-3.8
nox > Would run session tests-3.8
nox > Would create virtual environment (virtualenv) using python3.8 in .nox/tests-3-8
nox > pip install -r requirements-test.txt
nox > pip install -e .[tox_to_nox]
nox > pytest --cov=nox --cov-config .coveragerc --cov-report= tests/
nox > Would run session cover
nox > Would create virtual environment (virtualenv) using python3 in .nox/cover
nox > pip install coverage
nox > Would run multiple sessions:
nox > * tests-3.8
nox > * cover
```
The exit code should be always 0.

From the source code we could detect dry run by checking the session property:
```python
@nox.session(python="3.8")
def tests(session):
    """Run test suite with pytest."""
    if session.dry_run:
        ...
    else:
        ...
```

The solution described above has one problem. We can not prevent people from having e.g. `shutil.rmtree(...)` in the function without first checking `session.dry_run`, so whether it is really dry run depends on how `noxfile.py` is written. We can only vouch for `nox` API, not the logic inside the `noxfile.py`. Even though it is still very useful functionality, however above limitation should be pointed out in the documentation and help message.

**Describe alternatives you've considered**
Another alternative that does not have the same problem is something similar to `--collect-only` from `pytest`. In that case, we do not need to execute the session functions, just collect them.

```
$ nox --collect-only -s tests-3.8
nox > Would run session tests-3.8
```
The exit code should be always 0.

**Contributing**
I can start working on implementing this if I have a green light.

Best regards,
Maciej

theacodes:
I like this idea! A couple of notes:

> nox > Would run session tests-3.8

[...]

PR #823, opened 2024-04-27
Add dry run option, test, and documentation

jklott:
This PR addresses Issue https://github.com/wntrblm/nox/issues/302 which requests for a dry run option to be added. This option displays to the console what steps would be taken when the command is run but does not actually run them.

As part of this PR I added a test and updated the docs accordingly. Please let me know what changes need to be made still and I am happy to fix up the PR.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -405,6 +405,10 @@ class Session:
             msg = "First argument to `session.run` is a list. Did you mean to use `session.run(*args)`?"
             raise ValueError(msg)
 
+        if self._runner.global_config.dry_run:
+            logger.info(f"Would run: {args}, but --dry-run is set.")
+            return None
+
         if self._runner.global_config.install_only:
             logger.info(f"Skipping {args[0]} run, as --install-only is set.")
             return None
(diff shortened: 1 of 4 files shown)


nox/sessions.py
```diff
@@ -405,6 +405,10 @@ def run(
             msg = "First argument to `session.run` is a list. Did you mean to use `session.run(*args)`?"
             raise ValueError(msg)
 
+        if self._runner.global_config.dry_run:
```
henryiii, line 408: Shouldn't these be moved down? It would not run {args} even if --dry-run was not set, because --install-only was set too.

tests/test_sessions.py
```diff
@@ -287,6 +287,14 @@ def test_run_install_only_should_install(self):
             **run_with_defaults(paths=mock.ANY, silent=True, env={}, external="error"),
         )
 
+    def test_run_dry_run(self):
+        session, _ = self.make_session_and_runner()
+
+        with mock.patch("nox.command.run", autospec=True) as run:
+            session.run(sys.executable, "--version", dry_run=True)
```
henryiii, line 294: This wasn't added. And I don't think it shovel be, though, I think is should be a command line flag only. Probably monkeypatch the value here instead.

PR #826, opened 2024-05-04
Add option to limit the logged arguments

FredStober:
When running nox from within pre-commit, the argument list can be quite large and fill several screens.
In most cases, the interesting thing is not the whole used command line argument but just the first few options.
Therefore, I've added a new option 'max_log_args' that allows to limit the number of arguments that are shown in the logging output.


FredStober:
Sorry, I forgot to commit a line to tests/test_sessions.py

cjolowicz:
Does this need to be a built in Nox feature? I suppose you're passing the files as session arguments (session.posargs) so in your session function you have full control over what gets logged. You can pass the silent flag to session.run and produce the log line yourself.

FredStober:
I tried to implement it outside with silence and log=False, but in case of an error, these flags do not apply and my command line with 500 file names get dumped - and I need to scroll past them to see the error.

Spacetown:
I agree with @cjolowicz, I want to see the full command on failure. BTW ruff supports pre-commit directly without the need of using a box session.

PR #829, opened 2024-05-19
fix: support noxfile being a symlink

### LINKED ISSUE #828: Wrong starting directory when noxfile.py is a symlink
I have a project where I want the noxfille to be a symlink to a different shared file.  When I do this, it throws off the paths b/c nox follows the symlink and starts in that directory instead of starting in the directory where the symlink is.

```
 ❯ ls -l noxfile.py 
lrwxrwxrwx 1 rsyring rsyring 21 May 15 23:09 noxfile.py -> ./template/noxfile.py

 ❯ nox -s cwd
nox > Running session cwd
nox > Creating virtual environment (uv) using python3 in .nox/cwd
cwd is: /home/rsyring/projects/copier-py-package/template
```

Was this an intentional design decision?  If not, how about starting where the symlink is instead (i.e. don't follow it)?

Thanks for your work on this project and consideration.

henryiii:
Fix #828.

If you need the non-symlinked file, it's easy to get with `__file__`, as can be seen in the tests.

### DIFF
--- a/nox/tasks.py
+++ b/nox/tasks.py
@@ -83,13 +83,15 @@ def load_nox_module(global_config: Namespace) -> types.ModuleType | int:
     # Be sure to expand variables
     global_config_noxfile = os.path.expandvars(global_config.noxfile)
 
+    # Make sure we only expand the parent dir just in case the noxfile is a symlink
+    noxfile_parent_dir = os.path.realpath(os.path.dirname(global_config_noxfile))
+
     # Save the absolute path to the Noxfile.
     # This will inoculate it if Nox changes paths because of an implicit
     # or explicit chdir (like the one below).
-    global_config.noxfile = os.path.realpath(global_config_noxfile)
-
-    # Make sure we only expand the parent dir just in case the noxfile is a symlink
-    noxfile_parent_dir = os.path.realpath(os.path.dirname(global_config.noxfile))
+    global_config.noxfile = os.path.join(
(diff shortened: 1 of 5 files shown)


nox/tasks.py
```diff
@@ -79,16 +79,21 @@ def load_nox_module(global_config: Namespace) -> types.ModuleType | int:
...
-            # Be sure to expand variables
-            os.path.expandvars(global_config.noxfile)
-        )
-        noxfile_parent_dir = os.path.realpath(os.path.dirname(global_config.noxfile))
 
+    # Be sure to expand variables
+    global_config_noxfile = os.path.expandvars(global_config.noxfile)
+
+    # Make sure we only expand the parent dir just in case the noxfile is a symlink
+    noxfile_parent_dir = os.path.realpath(os.path.dirname(global_config.noxfile))
```
henryiii, line 87:
```suggestion
    noxfile_parent_dir = os.path.realpath(os.path.dirname(global_config_noxfile))
```

jamesharris-garmin:
I actually use the existing behavior (see https://github.com/wntrblm/nox/issues/828#issuecomment-2147640069) but I can probably work around a change if this is more natural. I'll follow this so I know if this comes in and breaks me.

henryiii:
Going to get it in for the next version; as seen in the tests, you can work around the change if you need the old behavior using `__file__`.

PR #831, opened 2024-05-22
chore: use uv if available in action

henryiii:
If uv has been installed before the nox action, use it to speed up the install of nox. Should save ~10 seconds or so.


cjolowicz (approved):

henryiii:
Also moved one of the action tests to using uv, nice and snappy as expected. :)

PR #834, opened 2024-06-25
ci: Test with Python 3.13

edgarrmondragon:


.github/workflows/ci.yml
```diff
@@ -50,6 +58,7 @@ jobs:
         with:
           name: coverage-${{ github.job }}-${{ strategy.job-index }}
           path: .coverage.*
+          include-hidden-files: true
```
henryiii, line 61:
```suggestion
          include-hidden-files: true
          if-no-files-found: error
```

.github/workflows/ci.yml
```diff
@@ -39,6 +46,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
           miniforge-variant: Mambaforge
           use-mamba: true
+          channels: conda-forge/label/python_rc,conda-forge
```
edgarrmondragon, line 49: Nice, TIL @henryiii!

henryiii: I figured it out today too. :) I've been seeing my feedstocks being rebuilt for 3.13, so I knew it had to be somewhere.

henryiii:
This also contains a fix for upload-artifact v4.4 now avoiding hidden files by default.

henryiii:
Looks like the action tests are only running on main. Those also need 3.13.

PR #835, opened 2024-06-25
Update type annotation of `stderr` parameter to make it optional

edgarrmondragon:
This is used, for example, by nox-poetry: https://github.com/cjolowicz/nox-poetry/blob/fb20b69892d394e8a4b97ae4d1f1d06cf811d32f/src/nox_poetry/poetry.py#L91


cjolowicz (approved):
LGTM. If you want to get the command output using the silent flag, passing stderr as None prevents mingling stdout and stderr.

PR #839, opened 2024-08-03
Fix typos

ember91:


cjolowicz:
Thanks!

One isn't really a typo but fair enough :)

https://en.m.wiktionary.org/wiki/moar

cjolowicz (approved):

ember91:
Merge this when you want

PR #847, opened 2024-09-10
feat: install_and_run_script (PEP 721)

henryiii:
See https://github.com/wntrblm/nox/issues/808#issuecomment-210613849. This makes using PEP 723 easier and more directly supported.


theacodes:
Looks good to me. :)

PR #851, opened 2024-09-29
chore: include 3.13 in classifiers

henryiii:
Already in tests.

### DIFF
--- a/nox/manifest.py
+++ b/nox/manifest.py
@@ -319,7 +319,7 @@ class Manifest:
         # Locate the session in the list of all sessions, and place it at
         # the end of the queue.
         for s in self._all_sessions:
-            if s == session or s.name == session or session in s.signatures:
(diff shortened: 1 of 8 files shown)


nox/manifest.py
```diff
@@ -319,7 +319,7 @@ def notify(
         # Locate the session in the list of all sessions, and place it at
         # the end of the queue.
         for s in self._all_sessions:
-            if s == session or s.name == session or session in s.signatures:
+            if s == session or s.name == session or session in s.signatures:  # noqa: PLR1714
```
henryiii, line 322: Probably should report this one to Ruff. It's not actually correct, since we are relying on short-circuiting here.

henryiii: Actually, maybe it's fine. I think s.name has to be valid, and session is either a string or a session.

Issue #848, opened 2024-09-10
Requesting dependencies in a noxfile


henryiii:
### How would this feature be useful?

This is just a wild idea, nothing might come of it, so take it with a grain of salt!

I’d like to have the ability to define and use packages inside the noxfile itself. Since nox knows how to make venvs and install packages, it seems like it should be able to request packages and even new nox versions. This would be a key prerequisite for supporting plugins, and 80% of the functionality of plugins would come from just being able to add packages; since noxfiles are just Python, you could write a package that adds common sessions, for example, and just use it in the noxfile, no need for a plugin system. An actual plugin system could be for things like new venv backends, locking, etc, and is not addressed yet by this proposal.

### Describe the solution you'd like

The core idea here is that nox could read PEP 723 data. This is `dependencies`, `requires-python`, and `tool.nox`. For the initial proposal, I'm suggesting using the standard `dependencies` and `requires-python`.

Nox would evaluate these two fields before it reads the `noxfile.py`. If the conditions are not met, it would try to make a special “_nox_self” environment (session) with the requirements and then run nox from there. This would only happen once - this new run would not be allowed to do this again (though it should always pass since you installed those requirements). `requires-python` would be respected (though this could be implemented in a follow-up). Due to the way this works, you kind of get `nox.needs-version` for free!

```python
# /// script
# dependencies = [“nox>=2024.12.31”, “nox-some-standard-sessions”, "pyyaml"]
# requires-python = ">=3.10"
# ///
```

If any of the packages are not installed, then the requirements list doesn’t pass and a new environment with a new nox is created, then called.

There are two downsides to not using our own `tool.nox` section above:

* You would not be able to specify this in pyproject.toml. Don’t think that’s too important, I don't think these need to be specified in pyproject.toml.
* This is not runnable by standardized runners, but only by nox.

For the second point, I think we could actually provide a way to make a noxfile runnable via these tools. I played with it, and since we require nox above in the dependencies, and you can nearly fake it now with a `if __name__ == “__main__”: nox.__main__.main()` block. We could probably add a way to do this and make noxfiles (with a minor user addition) valid PEP 723 files. The real question would be, though, is that required, or can we just use it in this way for now? If the answer is no, anything with dependencies/requires-python must be runnable by standardized runners (which I don't think is the case), we can replace this with `tool.nox.*` settings instead.

### Describe alternatives you've considered

[...]

PR #856, opened 2024-10-05
fix: remove add_timestamp from noxfile.options

Wurstnase:
logging is setup before all tasks,
so it's not possible to use it as noxfile.option.

we could add it after tasks.merge_noxfile_options, but in this case the logger wouldn't setup early for tasks.load_nox_module

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -407,7 +407,6 @@ options.add_options(
         group=options.groups["reporting"],
         action="store_true",
         help="Adds a timestamp to logged output.",
-        noxfile=True,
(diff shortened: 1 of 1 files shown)


henryiii (approved):
This might be a good candidate for a tool item in PEP 721-style options (#848).

Wurstnase:
> This might be a good candidate for a tool item in PEP 721-style options (#848).

Sounds reasonable.

theacodes:
Wonderful, thank you for doing this. :)

PR #857, opened 2024-10-06
feat: add session sync for uv

Wurstnase:
This is a draft how `uv synv` could be added to `nox[uv]`.

```python
@nox.session(venv_backend="uv")
def test(session):
    session.sync(extras="all")  # inexact by default
    session.install("pytest")
    session.run(*"pytest tests".split())
```


henryiii:
This doesn't feel right to me yet. You can already do:

```python
@nox.session(venv_backend="uv")
def test(session):
    session.run_install("uv", "sync", "--inexact", "--frozen", "--all-extras", env={"UV_PROJECT_ENVIRONMENT": session.venv.location})
    session.install("pytest")
    session.run("pytest", "tests")
```

Which is more explicit and more flexible - the UV_PROJECT_ENVIRONMENT is inconvenient, but I'm not sure it warrants a custom command - a recipe for this would be better for now, I thinl. The benefit of a custom command would be to provide the same API for multiple backends (which isn't here yet). If we had a plugin system (#848), and users could add support for Poetry and whatnot, that might be more interesting. Or maybe if we integrate with `.venv`; the custom environments feature (https://github.com/wntrblm/nox/issues/167#issuecomment-2040967680) which I'd like to work on in the next month or two would also potentially provide a better place to put this; we could require that users use Env's if they want to use locking. We also don't know what the multi-environment support for uv will look like. So good idea, and great idea for a cookbook example, but maybe let's hold off on it for a bit?

Wurstnase:
Thanks for your feedback.

I start with `run_install` but I need to repeat this part again and again. In between I patched the `@nox.session` to return a `CustomSession` which has something like the `sync` method.

However, I see your point and I think it's worth to wait for [PEP 751](https://peps.python.org/pep-0751/).

johnthagen:
Related

- https://github.com/dantebben/nox-uv

PR #873, opened 2024-10-27
chore: use dependency-groups

henryiii:
As a side affect of using dependency-groups, `uv run pytest` works out of the box! I've split the test requirements into test and dev.

Also, the old dev requirements was including the conda-test file, which is actually a list of conda packages, not a PyPI list.

I've also modernized a few other things, like using `default=`, the python versions are read from the classifiers, author list fixed, and pypy3.9 is no longer supported by upstream.


henryiii:
@theacodes, @dhermes, @cjolowicz, or @DiddiLeija, could one of you enable PR builds on readthedocs? I've modernized the readthedocs config, should be much faster, but I'd like to see if it builds correctly in the PR.

theacodes:
@henryiii done!

henryiii:
There still seems to be something missing, like a webhook. Could you re-sync the webhook, perhaps?

https://docs.readthedocs.io/en/stable/guides/pull-requests.html

theacodes:
Alright, I resynced the webhook. Give it another shot?

henryiii:
That got it, thanks!

henryiii:
Cuts the docs time to less than half of that it was before. :)

Issue #868, opened 2024-10-15
`include_outer_env=False` will include outer_env when `env` is not set


Wurstnase:
### Current Behavior

On `include_outer_env=False`, all outer environment variable will be used, when `env` is not set or `None`.

### Expected Behavior

When `include_outer_env=False`, no environment variables should be set. Also when no `env` is set.
Current workaround is `env={}`

### Steps To Reproduce

```python
import nox

@nox.session
def env(session: Session):
    session.run("printenv", include_outer_env=False)
    session.run("printenv", include_outer_env=False, env={})
```

### Environment

```Markdown
- OS: MacOS 15
- Python: 3.12.7
- Nox: 2024.4.15
```

henryiii:
Actually, I think there's a second bug. I believe VIRTUAL_ENV and CONDA_PREFIX don't get set if you use this option, as well.

Wurstnase:
`tox`[^1] do a lot of magic to pass the correct values.

Should `nox` stay this clean way?

Maybe some more configurable?

```
include_outer_env: Literal["all", "recommended", "none"]
```
while `recommended` will pass a similar list of environment variables like in tox.

e.g. on my setup I need to set `HOME` again, otherwise it wouldn't find my .netrc with the credentials for a private pypi repository.

[^1]: https://github.com/tox-dev/tox/blob/08f2ac55d67b1ca91e56e61fb09d6f7d7034ef2a/src/tox/tox_env/api.py#L204

PR #874, opened 2024-10-27
fix: outer_env issues fixed

henryiii:
Fix https://github.com/wntrblm/nox/issues/868. Two issues:

* Setting `include_outer_env=False` does nothing at all if `env=` is not passed. The outer env is still included.
* Required environment variables set by nox itself, like `VIRTUAL_ENV` (required for `uv`), and `CONDA_PREFIX` (required for conda-family) would not be included if you do set both.

Fixing this was a bit intrusive: now `.env` only is the local environment
settings, and removals are added to it using None. The type has changed to match `.command`, with the None possibility. And then the final environment is built in the `_run`, where it can correctly process things like adding the bin dir and filtering the None's.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -618,9 +618,10 @@ class Session:
             args = (nox.virtualenv.UV, *args[1:])
 
         # Combine the env argument with our virtualenv's env vars.
+        env = env or {}
+        env = {**self.env, **env}
         if include_outer_env:
-            overlay_env = env or {}
-            env = {**self.env, **overlay_env}
+            env = {**self.virtualenv.outer_env, **env}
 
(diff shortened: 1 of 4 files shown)


DiddiLeija (approved):
This looks fine to me, as it seems to fix the issue properly!
However for this one I'd like to hear what other maintainers think ;)

henryiii:
> However for this one I'd like to hear what other maintainers think

Not sure about that, it's been open for nearly three months. ;)

DiddiLeija:
Yup, that's true, I was just a little concerned about some spots, but I've re-checked and everything is ok!
To me, you're free to merge :)

henryiii:
Thanks!

Issue #845, opened 2024-08-30
Enable reading dependency specifiers from PEP 621 `pyproject.toml`


paduszyk:
### How would this feature be useful?

Let us assume I develop a Python package. In my PEP 621 `pyproject.toml` I have:

```toml
[project]
# ...
dependencies = [
   "django >= 3.2, < 5.2",
   "typing-extensions >= 4.12.2, < 5",
]

# ...

[project.optional-dependencies]
# dev = [ ... ]
lint = [
  "ruff >= 0.6.3, < 1",
  "django-stubs[compatible-mypy] >= 5.0.4, < 6",
]
# test = [ ... ]
```

Then, I would like to have separate sessions for checking and formatting my codebase with `ruff` and type-checking it with `mypy`. It would go like this:

```python
@nox.session(tags=["lint"])
@nox.parametrize(
    "command",
    [
        "check",
        "format",
    ],
)
def ruff(session: nox.Session, command: str) -> None:
    session.install("-e", ".[lint]")

    session.run("ruff", command)


@nox.session(tags=["lint"])
def mypy(session: nox.Session) -> None:
    session.install("-e", ".[lint]")
 
    session.run("mypy", ".")
```

It works, but:

- actually, `ruff` session does not need the package as well as any other `lint` dependencies to be installed &mdash; we need `ruff` ONLY;
- `mypy` does not need `ruff` &mdash; nevertheless, it's installed because it's in the same dependency group.

A nice feature would be to tell Nox which dependencies should be installed, with version specifiers automatically read from `pyproject.toml`.

### Describe the solution you'd like

This is a draft for my `noxfile.py` implementing a potential solution:

```python
from __future__ import annotations

__all__ = [
    "mypy",
    "ruff",
]

import re
from functools import partial
from pathlib import Path
from typing import cast

import nox

PYPROJECT_TOML_PATH = Path(__file__).resolve().parent / "pyproject.toml"

def get_dependency_specifiers(pattern: str, *, group: str | None = None) -> list[str]:
    project = nox.project.load_toml(PYPROJECT_TOML_PATH).get("project")

    if not project:
        msg = (
            f"{PYPROJECT_TOML_PATH} is not a valid PEP 621 metadata file as "
            f"it does not contain a [project] table"
        )

        raise LookupError(msg)

    if group is None:
        dependencies = project.get("dependencies", [])
    else:
        try:
            dependencies = project.get("optional-dependencies", {})[group]
        except KeyError as e:
[...]

PR #876, opened 2024-10-28
feat: helper to get dependency-groups

henryiii:
Starting work on https://github.com/wntrblm/nox/issues/845.

~~I was going to use @sirosen's dependency-groups, but it only supports 3.9+ - even the first version was 3.9+. I don't think that was a great idea for a foundational utility, but it's what it is. So we can't use it yet. And this is pretty easy, I think we can just write our own support as long as it's tested.~~

~~I've improved on the example implementation several ways: `setdefault` is a safer than `defaultdict`. I've used generators to remove the appends and extends. And I've used `*args` to simplify the past_groups handling.~~

I think we can start with a function like this, and maybe eventually come up with a nicer interface for things in pyproject.toml. But this is simple and could be used by the fancier interface later.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
         files: ^nox/
         args: []
         additional_dependencies:
-          - dependency-groups>=1.1
+          - dependency-groups>=1.2
           - jinja2
           - packaging
           - importlib_metadata
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,6 +37,7 @@ repos:
         files: ^nox/
         args: []
         additional_dependencies:
+          - dependency-groups>=1.1
           - jinja2
           - packaging
           - importlib_metadata
--- a/nox/project.py
+++ b/nox/project.py
@@ -131,7 +131,7 @@ def python_versions(
     return [f"3.{v}" for v in range(min_minor_version, max_minor_version + 1)]
 
 
-def dependency_groups(pyproject: dict[str, Any], *groups: str) -> list[str]:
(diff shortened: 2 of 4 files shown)


sirosen:
> I was going to use @sirosen's dependency-groups, but it only supports 3.9+ - even the first version was 3.9+. I don't think that was a great idea for a foundational utility, but it's what it is.

Oh, I'm happy to add 3.8 support! In fact, rather than delaying, I just did in v1.1.0 .

I'm not sure why I wasn't more conservative about the supported versions when creating the library, but I agree that was a mistake.

henryiii:
Great, thanks! FWIW, I don't care if the latest version is 3.9+ (or 3.13+), I just need a working 3.8 version that gets pulled by 3.8. :)

henryiii:
FYI, the only thing left is rebasing after #877 and adding docs.

.pre-commit-config.yaml
```diff
@@ -37,6 +37,7 @@ repos:
         files: ^nox/
         args: []
         additional_dependencies:
+          - dependency-groups>=1.1
```
henryiii, line 40:
```suggestion
          - dependency-groups>=1.2
```

nox/project.py
```diff
@@ -127,3 +129,20 @@ def python_versions(
...
+
+    .. code-block:: python
+
+        @nox.session
+        def test(session):
+            pyproject = nox.project.load_toml("pyproject.toml")
+            session.install(*nox.project.dependency_groups(pyproject, "dev"))
+    """
+    dep_groups = pyproject["dependency-groups"]
+    return [item for g in groups for item in resolve(dep_groups, g)]
```
henryiii, line 148:
```suggestion
    return resolve(dep_groups, *groups)
```

cjolowicz (approved):

PR #877, opened 2024-10-28
feat: add helper to get the Python listing

### LINKED ISSUE #855: Reading Python classifiers from pyproject.toml

We could convert python classifiers to a Python version list.

Here's the current way to do it:

```python
ALL_PYTHONS = [
    c.split()[-1]
    for c in nox.project.load_toml("pyproject.toml")["project"]["classifiers"]
    if c.startswith("Programming Language :: Python :: 3.")
]
```

_No response_

### Anything else?

_No response_

henryiii:
Close #855.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -466,8 +466,8 @@ class.
     :members:
     :undoc-members:
 
-The project helpers
--------------------
+The pyproject.toml helpers
+--------------------------
 
 Nox provides helpers for ``pyproject.toml`` projects in the ``nox.project`` namespace.
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -466,6 +466,13 @@ class.
     :members:
     :undoc-members:
 
+The project helpers
+-------------------
+
+Nox provides helpers for ``pyproject.toml`` projects in the ``nox.project`` namespace.
+
+.. automodule:: nox.project
+   :members:
 
 Modifying Nox's behavior in the Noxfile
 ---------------------------------------
--- a/nox/project.py
+++ b/nox/project.py
@@ -6,6 +6,8 @@ import sys
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+import packaging.specifiers
+
 if TYPE_CHECKING:
     from typing import Any
 
@@ -15,7 +17,7 @@ else:
     import tomllib
 
 
-__all__ = ["load_toml"]
+__all__ = ["load_toml", "python_list"]
 
 
 def __dir__() -> list[str]:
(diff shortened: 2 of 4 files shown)


nox/project.py
```diff
@@ -67,3 +78,52 @@ def _load_script_block(filepath: Path) -> dict[str, Any]:
...
+        if from_classifiers:
+            return from_classifiers
+        raise ValueError('No Python version classifiers found in "project.classifiers"')
+
+    requires_python_str = pyproject.get("project", {}).get("requires-python", "")
+    if not requires_python_str:
+        raise ValueError('No "project.requires-python" value set')
+
+    for spec in packaging.specifiers.SpecifierSet(requires_python_str):
+        if spec.operator in {">", ">="}:
```
henryiii, line 121:
```suggestion
        if spec.operator in {">", ">=", "~="}:
```

docs/config.rst
```diff
@@ -466,6 +466,13 @@ class.
     :members:
     :undoc-members:
 
+The project helpers
```
theacodes, line 469: maybe "The project.toml helpers"? "project" is a bit generic on its own

nox/project.py
```diff
@@ -67,3 +78,52 @@ def _load_script_block(filepath: Path) -> dict[str, Any]:
         for line in matches[0].group("content").splitlines(keepends=True)
     )
     return tomllib.loads(content)
+
+
+def python_list(
```
theacodes, line 83: Maybe `python_versions`?

theacodes (commented):
Mostly looks good, just some nitpicks about naming

cjolowicz:
I was surprised that `max_version` determines from where we read the Python versions.

cjolowicz:
Makes sense. Another approach would be to always look at Trove classifiers first, and fall back to `requires_python` if no classifiers are found. The `max_version` parameter would then just do what it says on the tin: apply an upper bound. (If it's not specified and we can't determine the upper bound from pyproject.toml, we could either bail out or apply our notion of what Python's current version is.) I don't feel strongly about this, though. Happy for this to go in 👍

henryiii:
That's exactly what I originally tried, actually, but then I would have had to implement filtering, and why would you ever want to filter the classifiers using this parameter? What would be more likely to happen is someone would forget to update this bound when adding the classifier. Also, having it be "auto" means you'd have to pick what to do if classifiers and requires-python didn't match.

ews-ffarella:
Hi!
Would that work with `Programming Language :: Python :: 3.13t`.
Or is this a bad classifiers?

henryiii:
You can add to this list or process it, though. Including adding PyPy, etc.

PR #878, opened 2024-10-28
feat: expose main as nox.main

henryiii:
See https://github.com/wntrblm/nox/issues/848. This just makes using PEP 723 a bit nicer, but doesn't implement native support in nox yet.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -559,7 +559,7 @@ the tags, so all three sessions:
     * flake8
 
 
-Running without nox
+Running without the nox command
 -------------------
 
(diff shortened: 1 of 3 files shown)


docs/tutorial.rst
```diff
@@ -559,6 +559,28 @@ the tags, so all three sessions:
     * flake8
 
 
+Running without nox
```
theacodes, line 562:
```suggestion
Running without the nox command
```

theacodes (approved):
Looks good with minor suggestion

cjolowicz (approved):

Issue #803, opened 2024-03-23
2024.03.02: pytest fails in 4 units


kloczek:
### Current Behavior

<details>
<summary>Here is pytest summary output:</summary>

```console
=========================== short test summary info ============================
SKIPPED [1] tests/test_command.py:62: Needs git
SKIPPED [1] tests/test_command.py:362: Only run this test on Windows.
SKIPPED [1] tests/test_command.py:384: Only run this test on Windows.
SKIPPED [1] tests/test_tox_to_nox.py:284: Not supported in tox 4.
SKIPPED [1] tests/test_virtualenv.py:154: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:183: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:195: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:209: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:239: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:245: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:262: Missing uv command.
SKIPPED [1] tests/test_virtualenv.py:423: Missing uv command.
SKIPPED [1] tests/test_virtualenv.py:455: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:468: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:479: Missing uv command.
SKIPPED [1] tests/test_virtualenv.py:479: Missing conda command.
SKIPPED [1] tests/test_virtualenv.py:500: Missing uv command.
SKIPPED [1] tests/test_virtualenv.py:526: Missing uv command.
SKIPPED [1] tests/test_virtualenv.py:599: Python 2.7 unsupported for virtualenv>=20.22.0
XPASS tests/test_logger.py::test_no_color_timestamp[color]
FAILED tests/test_sessions.py::TestSessionRunner::test__create_venv_options[nox.virtualenv.CondaEnv.create-conda-CondaEnv]
FAILED tests/test_virtualenv.py::test_stale_environment[venv-virtualenv-True]
FAILED tests/test_virtualenv.py::test_create_reuse_venv_environment - nox.com...
FAILED tests/test_virtualenv.py::test_create_venv_backend - nox.command.Comma...
============ 4 failed, 520 passed, 19 skipped, 1 xpassed in 16.41s =============
```
</details>

[...]

PR #879, opened 2024-10-28
fix: don't add a handler if there's already one

henryiii:
This might help the logger bug we see when nox runs itself, like in https://github.com/wntrblm/nox/issues/803.


henryiii:
I think this can still occur, will have to continue to investigate later (happy to take suggestions!)

PR #871, opened 2024-10-26
fix: error with helpful message if invalid option is set via nox.options

henryiii:
Fix part of #869 - make sure only valid options are set.

The most recent commit also adds static typing for `nox.options`.

PR #880, opened 2024-10-29
fix: validate entries in nox.config, too, using attrs

henryiii:
This is a followup to #871. It also reduces the need for #872, as I think it makes it harder or impossible to incorrectly set this via normal means.

Attrs is a well respected, zero dependency, small library by a CPython core dev, but it is an extra dependency. If we did adopt it, though, I would start moving over a few other things, which could improve our internal typing too.

<details>
<summary>Incomplete list for "NoxCmdOptions":</summary>

```python
class NoxCmdOptions(NoxOptions):
    version: bool
    list_sessions: bool
    json: bool
    posargs: list[str]
    add_timestamp: bool
    no_venv: bool
    R: bool
    noxfile: str
    envdir: None | str
    extra_pythons: list[str]
    force_pythons: list[str]
    install_only: bool
    no_install: bool
    report: None | str
    no_reuse_existing_virtualenvs: bool  # backcompat
    non_interactive: bool
    nocolor: bool
    forcecolor: bool
    color: bool
    invoked_from: str
```

</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,6 +37,7 @@ repos:
         files: ^nox/
         args: []
         additional_dependencies:
+          - attrs
           - colorlog
           - dependency-groups>=1.2
           - jinja2
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -20,7 +20,6 @@ from __future__ import annotations
 
 import argparse
 import collections
-import dataclasses
 import functools
 from argparse import ArgumentError as ArgumentError  # noqa: PLC0414
 from argparse import ArgumentParser, Namespace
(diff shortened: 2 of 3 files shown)


DiddiLeija (commented):
Looks like a very handy fix to me! However I'd like to see what other mantainers think of adding a dependency before giving the approval :)

henryiii:
It's been a couple of weeks without any complaints, and I'd like to build on this, so I think I'll merge unless there are any last objections!

Milad335t (approved):

Milad335t (approved):

Milad335t (approved):

Issue #414, opened 2021-04-09
Have noxfile.py install dependencies for itself


mikeholler:
**How would this feature be useful?**

Nox has sessions that create virtualenvs, but nox itself executes using the environment available to the system. It would be nice if we could have as predictable an environment for the nox buildscript as we have for each session.

**Describe the solution you'd like**

Have a second file, `noxfile-settings.py` that nox executes if present to create an environment for `noxfile.py` to run in. The contents of the file might look something like this:

```
from nox.settings import noxfile_session

noxfile_settings.install("GitPython==3.1.14")
```

Then, in `noxfile.py`, a user can have a script that calls `from git import Repo` and feel confident that they are using the same version of GitPython as other developers.

Note: The `settings.gradle` files and `buildscript { ... }` blocks from Gradle are examples of another build system doing this kind of thing.

**Describe alternatives you've considered**

As far as I know there are no workarounds or allowances for this.

henryiii:
I'd probably put something like this in a non-Python file, say `pyproject.toml` under `[tool.nox] requires=["GitPython==3.1.14"]`, etc. Sounds like it would be a rather large change, though, since nox would need to run nox in an venv. You'd probably want to control/customize that venv, etc...

crwilcox:
Is there a scenario where `requirements-dev.txt` would be different from the version of `GitPython` you would need for nox?

As a pattern, I tend to have a venv per project I work on, so nox already is in a venv, but just shared with the project I work on.

I think there may be a more particular corner case worth describing here?

Issue #814, opened 2024-04-09
Support version specifiers in `nox`'s Python definition


chrysle:
### How would this feature be useful?

I think it would really spare some work to not have to hard-code the Python versions to use in `nox`'s session decorator or elsewhere. Version specifiers are a clean and efficient way to specify version ranges and exclude versions. `packaging`  could handle the parsing of the specifier set, as it's also already a dependency of `nox`. This might also be useful regarding what #811 proposes.

### Describe the solution you'd like

Being able to specify something like the following in the end:

```python
@nox.session(python=">=3.8<=3.12!=3.9")
def tests(session):
    pass
```

### Describe alternatives you've considered

Working with the Python specification syntax that exists currently, which isn't too tedious.

henryiii:
Yes, it would. The biggest issue would be parametrization; how would you parametrize over Pythons using this? I don't like Python parametrization most of the time (the new force-python support for unparameterized session is _fantastic_!), but some people do. A followup might be to add some sort of generator function that would turn `>=3.10` into `3.10`, `3.11`, and `3.12`.

I'm not sure how this would interact with PyPy.

Also, great care should be taken not to promote upper caps. These are not supported properly in the Python ecosystem and packages that add them (numba, numpy, etc) quickly have had to remove them due to breakage and them not behaving as you might expect - they are used to back solve, and back solving is never correct for an upper cap, you are not more likely to support a newer Python version in the past. See https://iscinumpy.dev/post/bound-version-constraints/ for example. I _think_ this would only be tempting for parametrization, so maybe we could ignore upper caps and require a user to specify an upper cap in the helper function, something like `nox.project.supported_pythons(">=3.10", max="3.12")`.

I've been eyeing https://pypi.org/project/findpython and wondering if we could use that to find Python's. I'm not sure if it supports Specifiers, it calls it a `version_spec`, but there are no examples with one, and `pipx run findpython ">=3.10"` doesn't work.

chrysle:
> (the new force-python support for unparameterized session is _fantastic_!)

Thank you!
> A followup might be to add some sort of generator function that would turn `>=3.10` into `3.10`, `3.11`, and `3.12`.

I'm in favour.

> I'm not sure how this would interact with PyPy.

Could you elaborate on your concerns?

>  I _think_ this would only be tempting for parametrization, so maybe we could ignore upper caps and require a user to specify an upper cap in the helper function, something like `nox.project.supported_pythons(">=3.10", max="3.12")`.

[...]

PR #881, opened 2024-10-31
feat: support PEP 723 noxfiles

henryiii:
I've started https://github.com/wntrblm/nox/issues/848. This allows you to write a noxfile like this:

```python
# /// script
# dependencies = ["nox>=1.2.3", "numpy"]
# ///

import nox
import numpy

@nox.session
def example(session: nox.Session) -> None:
    print(numpy.__file__)
```

Any dependencies, including version ranges and nox, will be checked. If not satisfied, nox will make a new `.nox/_nox_script_mode` environment, set it up with the above dependencies, and then pass control to the new nox environment. Requires-python is not respected yet, might need https://github.com/wntrblm/nox/issues/814. That can be added later, not a target for this PR.

I've allowed this to be passed via flag and envvar, but envvar always comes first, and overrides the flag. This ensures the second run will never trigger a third run of nox. If someone has other ideas for ways to ensure this, I'm open. Also, the envvar is needed if you request an older nox, since the flag won't be valid in the "inner" run. `--script-mode=fresh` allows you to enforce a fresh environment.

It supports both pip and uv; since this is a new feature, defaulted to uv if available, and fallback on pip. ~~Maybe should be configurable; currently it's locked on the "default" is `uv|virtualenv`.~~ I've made it configurable with `tool.nox.script-venv-backend`/`--script-venv-backend`/`NOX_SCRIPT_VENV_BACKEND`.

It's a bit less clever on Windows following pipx's example; pipx doesn't use process replacement there; the original PR said it "wasn't reliable" for pipx. Looks like path quoting is not handled by this method, and it doesn't do anything special on Windows. https://github.com/pypa/pipx/pull/531

Note this is available in tox as `requires = ...`, but before PEP 723 there wasn't a clean way to do this for nox.

Fix https://github.com/wntrblm/nox/issues/414.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -597,8 +597,8 @@ the tags, so all three sessions:
     * flake8
 
 
-Running without the nox command
--------------------------------
+Running without the nox command or adding dependencies
+------------------------------------------------------
 
 With a few small additions to your noxfile, you can support running using only
 a generalized Python runner, such as ``pipx run noxfile.py``, ``uv run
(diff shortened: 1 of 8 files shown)


henryiii:
It's been exactly three months today, so going to go ahead and merge, we can revert or change if there are any post-merge concerns!

PR #891, opened 2024-11-12
fix: include encoding for consistent behavior (default in Python 3.15+)

henryiii:
Noticed when checking some Ruff preview stuff.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -60,6 +60,7 @@ def tests(session: nox.Session, tox_version: str) -> None:
     session.install("-e.[tox_to_nox]")
     if tox_version != "latest":
         session.install(f"tox{tox_version}")
+    extra_env = {} if tox_version != "latest" else {"PYTHONWARNDEFAULTENCODING": "1"}
     session.run(
         "pytest",
         "--cov",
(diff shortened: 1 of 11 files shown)


noxfile.py
```diff
@@ -60,6 +60,7 @@ def tests(session: nox.Session, tox_version: str) -> None:
     session.install("-e.[tox_to_nox]")
     if tox_version != "latest":
         session.install(f"tox{tox_version}")
+    extra_env = {} if tox_version != "latest" else {"PYTHONWARNDEFAULTENCODING": "1"}
```
henryiii, line 63:
```suggestion
    extra_env = {"PYTHONWARNDEFAULTENCODING": "1"} if tox_version == "latest" else {}
```

pyproject.toml
```diff
@@ -137,7 +137,9 @@ max_supported_python = "3.13"
 minversion = "7.0"
 addopts = [ "-ra", "--strict-markers", "--strict-config" ]
 xfail_strict = true
-filterwarnings = [ "error" ]
+filterwarnings = [
+  "error",
+]
```
henryiii, line 142:
```suggestion
filterwarnings = [ "error" ]
```

PR #893, opened 2024-11-12
chore: add more Ruff checks and fixes

henryiii:
Adding more checks. First commit is stable checks, and the second commit is some from preview (not activated, just fixed the issue if it looked useful).

- **chore: add more Ruff checks and fix issues**
- **chore: add some preview check fixes**

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -14,7 +14,6 @@
 
 from __future__ import annotations
 
-import argparse
 import contextlib
 import enum
 import hashlib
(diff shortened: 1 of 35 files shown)


nox/sessions.py
```diff
@@ -278,6 +280,7 @@ def install_and_run_script(
             silent=silent,
             success_codes=success_codes,
             external=None,
+            log=log,
```
henryiii, line 283: This is a bug fix, it wasn't being passed through.

henryiii: If it was passed, it was ignored before, now it will work as intended. It just affects logging, so it shouldn't break anyone. And this is a very new method, it was added last release.

henryiii:
Thanks! I've rebased and also bumped to Ruff 0.8, which gets rid of one of our ignores.

PR #900, opened 2024-11-28
Use `uv python install` only with uv backend

### LINKED ISSUE #899: Nox installs Python using `uv` with non-uv backend
### Current Behavior

Nox installs Python using `uv` with non-uv backend. If uv is installed on the system.

### Expected Behavior

Don't use `uv` to install Python with non-uv backend.

### Steps To Reproduce

noxfile.py:

```python
import nox

@nox.session(python="3.8")
def lint(session):
    ...
```

Invoke `nox -s lint`.

If Python 3.8 is **not installed globally**:

```plaintext
nox > Running session lint
Installed Python 3.8.20 in 2m 20s
 + cpython-3.8.20-windows-x86_64-none
nox > Creating virtual environment (virtualenv) using python3.8 in .nox\lint
nox > Command ...\.venv\Scripts\python.exe -m virtualenv '...\.nox\lint' -p python3.8 failed with exit code 1:
RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.8'
nox > Session lint failed.
```

If Python 3.8 is **installed globally**.

```plaintext
nox > Running session lint
Installed Python 3.8.20 in 3m 55s
 + cpython-3.8.20-windows-x86_64-none
nox > Creating virtual environment (virtualenv) using python3.8 in .nox\lint
nox > Session lint was successful.
```

Nox installed Python using `uv`. But the Nox session virtual environment uses the globally installed version of Python. Because `virtualenv` is not compatible with `uv`.

### Environment

```Markdown
- OS: Windows 10
- Python: 3.11
- Nox: 2024.10.9
```

### Anything else?

_No response_

oliversen:
fix #899

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -598,7 +598,9 @@ class VirtualEnv(ProcessEnv):
             self._resolved = cleaned_interpreter
             return self._resolved
 
-        if self.venv_backend == "uv" and HAS_UV and UV_PYTHON_SUPPORT:
+        if (
+            self.venv_backend == "uv" and HAS_UV and UV_PYTHON_SUPPORT
+        ):  # pragma: nocover
             uv_python_success = uv_install_python(cleaned_interpreter)
             if uv_python_success:
                 self._resolved = cleaned_interpreter
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -229,6 +229,7 @@ def locate_via_py(version: str) -> str | None:
             check=False,
             text=True,
             capture_output=True,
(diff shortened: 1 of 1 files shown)


henryiii (approved):

henryiii:
I think a test needs updating too.

DiddiLeija (approved):
I guess some coverage is required for the new behavior, but that's all what it's needed.

henryiii:
I've just marked this bit as nocover, since installing uv Python versions is a global operation and I don't think we want to require it in our tests.

PR #901, opened 2024-12-04
correct virtualenv bin dir under mingw python

### LINKED ISSUE #630: `session.install` fails on MSYS2 virtualenv
### Current Behavior

Install `nox` using MSYS2 Python. (Use setup-msys2 action to create an environment on Github actions.)

`session.install(<anything>)` will fail, with errors such as the following:

 ```
nox > Error: python is not installed into the virtualenv, it is located at D:/a/_temp/msys64/mingw64/bin/python.EXE. Pass external=True into run() to explicitly allow this.
```

I tried to use `--no-error-on-external-run` flag to workaround this, but it looks like the implementation of `session.install` hard forces `external="error"` so this makes it a hard blocker.

### Expected Behavior

`nox` is able to install required dependencies into the virtualenv.

### Steps To Reproduce

See https://github.com/davidhewitt/msys2-nox-repro for a minimal repro to set up an environment in Github actions which reproduces this failure.

Failing run is e.g. https://github.com/davidhewitt/msys2-nox-repro/runs/7296705086?check_suite_focus=true

### Environment

```Markdown
- OS: Windows 10
- Python: 3.10 (MinGW / MSYS2)
- Nox: 2022.1.7
```

### Anything else?

_No response_

davidhewitt:
I believe this fixes #630 

... at least, I was able to remove the workaround I had in https://github.com/PyO3/setuptools-rust/pull/501

Open to alternative ideas / suggestions how you'd like this tested.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -24,6 +24,7 @@ import re
 import shutil
 import subprocess
 import sys
+import sysconfig
 from pathlib import Path
 from socket import gethostbyname
 from typing import TYPE_CHECKING, Any, ClassVar
@@ -75,6 +76,7 @@ _BLACKLISTED_ENV_VARS = frozenset(
     ]
 )
 _SYSTEM = platform.system()
+_IS_MINGW = sysconfig.get_platform().startswith("mingw")
 
 
 def find_uv() -> tuple[bool, str]:
@@ -630,7 +632,7 @@ class VirtualEnv(ProcessEnv):
     @property
     def bin_paths(self) -> list[str]:
         """Returns the location of the virtualenv's bin folder."""
-        if _SYSTEM == "Windows":
+        if _SYSTEM == "Windows" and not _IS_MINGW:
             return [os.path.join(self.location, "Scripts")]
         return [os.path.join(self.location, "bin")]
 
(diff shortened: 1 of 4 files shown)


henryiii:
The idiomatic thing is to use `sys.platform.startswith`, which also happens to be correct here, I think: https://stackoverflow.com/questions/446209/possible-values-from-sys-platform

henryiii:
We might need to be careful mocking (see #902), but I'm curious to see if this runs first.

henryiii:
Could you make sure this also fixes your issue? I think it's a better fix and should handle Cygwin and such too.

davidhewitt:
Sorry for the delay, I finally reran on the `setuptools-rust` CI and it looks like this branch didn't work, I'll try to debug when I get the chance.

davidhewitt:
Ok so I had a debugging session and it seems like `sys.platform` is still just `win32`, so that seems to be insufficient to detect msys2. The MSYS2 portability docs seem to explicitly point to using `sysconfig.get_platform()` in https://www.msys2.org/docs/python/

I restored the `_IS_MINGW` check against `sysconfig.get_platform()`, and it's fixed the workflow again. Would you like me to push a (mocked) test?

davidhewitt:
I found a brief moment to push a test, hopefully looks suitable. Happy New Year!

PR #909, opened 2025-01-09
chore: bump to ruff 0.9 and mypy 1.14

henryiii:
2025 style.


tests/test_tox_to_nox.py
```diff
@@ -383,7 +382,7 @@ def test_commands_with_requirements(makeconfig: Callable[[str], str]) -> None:
 
     @nox.session(python='python3.13')
     def aiohttp(session):
-        session.install('pytest', 'pytest-cov', '-r', 'requirements/aiohttp.txt')
+        session.install('pytest', 'pytest-cov', '-rrequirements/aiohttp.txt')
```
DiddiLeija, line 385: I guess this is the reason of the failing checks :)

```suggestion
        session.install('pytest', 'pytest-cov', '-r', 'requirements/aiohttp.txt')
```

.pre-commit-config.yaml
```diff
@@ -24,14 +24,14 @@ repos:
       - id: pyproject-fmt
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.0
+    rev: v0.9.0
```
DiddiLeija, line 27: Ruff 0.9.1 came out with some fixes, maybe it's fine to add it :)

```suggestion
    rev: v0.9.1
```

DiddiLeija (commented):
LGTM! Though I found a small mistake :)

tests/test_tox_to_nox.py
henryiii: The problem is that the older tox is now repairing this as one joined string. I think we just need to make the tests work in either case, as they are identical.

henryiii:
How do we feel about dropping support for tox 3? :) AFAICT, the tox 4 implementation is independent of tox, so it should "just work" even if tox is not installed.

tests/test_tox_to_nox.py
```diff
@@ -327,14 +327,17 @@ def test_commands_with_requirements(makeconfig: Callable[[str], str]) -> None:
...
         deps =
             pytest
             pytest-cov
             aiohttp: -r requirements/aiohttp.txt
     """)
     )
 
+    # This is equivalent
+    result = result.replace("'-rrequirements", "'-r', 'requirements")
+
```
henryiii, line 340:
```suggestion
```

tests/test_tox_to_nox.py
```diff
@@ -327,7 +327,7 @@ def test_commands_with_requirements(makeconfig: Callable[[str], str]) -> None:
         envlist = aiohttp
 
         [testenv]
-        use_develop = true
+        usedevelop = true
```
henryiii, line 330:
```suggestion
        use_develop = true
```

PR #913, opened 2025-01-22
fix: allow pypy- to be used (matching GHA)

### LINKED ISSUE #861: pypy-3.10 doesn't work as an interpreter name
### Current Behavior

GitHub Actions called PyPy interpreters `pypy-3.10`. I think we call them `pypy3.10`. Should we support the GHA form too?

henryiii:
Fix #861.


henryiii:
Guess I'll count a couple thumbs up as a passing review. ;)

PR #916, opened 2025-01-28
fix(gha): drop PyPy from the default list for the GitHub Action

### LINKED ISSUE #915: ARM runners don't have PyPy
Our action default includes `pypy-3.10`, which is missing for builds like `ubuntu-arm-24.04`. It's also probably not present on the other images, which means it has to be downloaded, making the action slower. I think we could just remove pypy from the default, like other projects are doing (cibuildwheel 3, for example). Someone who needs it could ask for it (as they already need to do for older pypy). Then our default would also work on the ARM images. I'd assume they might eventually build it for ARM, but this seems pretty reasonable overall?

henryiii:
Close #915. Add an ARM runner, drop PyPy from the default list, which should also speed up non-ARM runs.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -21,6 +21,7 @@ jobs:
         os:
           - ubuntu-22.04
           - ubuntu-24.04
+          - ubuntu-24.04-arm
           - windows-2019
           - windows-2022
           - macos-13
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
Makes sense after reading https://github.com/wntrblm/nox/issues/915 👍

Issue #843, opened 2024-08-30
Default value for `filename` in `nox.project.load_toml`


paduszyk:
### How would this feature be useful?

I think most people will use the `nox.project.load_toml` function (added in #811) with their `pyproject.toml`. In fact, the name of the function itself suggests loading from a TOML.

### Describe the solution you'd like

I would update the function's signature:

https://github.com/wntrblm/nox/blob/e0868b6c092c80837f86463a54d34379851d43d7/nox/project.py#L33

with:

```python
def load_toml(filename: os.PathLike[str] | str = "pyproject.toml") -> dict[str, Any]: 
```

or set the default to `None` and evaluate the file location from `pwd`:

```python
def load_toml(filename: os.PathLike[str] | str | None = None) -> dict[str, Any]: 
    if filename is None:
        filename = Path.cwd() / "pyproject.toml"

    # ...
```    

### Describe alternatives you've considered

Alternatively, the path to the `pyproject.toml` can be set via `nox.options`.

CAM-Gerlach:
Not a Nox maintainer or anything, but NB, you can just do `Path("pyproject.toml")` rather than the unnecessary `Path.cwd() / "pyproject.toml"`, since `filename` doesn't require absolute paths (as the example itself shows).

henryiii:
The function was actually added to make loading PEP 723 files easier, since it's a pretty trivial wrapper over tomllib/tomli otherwise. But defaulting to pyprojec.toml doesn't seem unreasonable.

CAM-Gerlach:
Well, as I _am_ a PEP editor, I _can_ say that despite the widespread misconception to the contrary, PEP 20 (the Zen of Python) was never intended to be taken seriously as a set of rules for the language, as opposed to one core dev's tongue in cheek ideas :)

PR #917, opened 2025-01-28
fix: add a pyproject.toml default for load_toml

henryiii:
Close https://github.com/wntrblm/nox/issues/843. I was worried about nox's handling of the CWD but I think this is actually safe.

### DIFF
--- a/nox/project.py
+++ b/nox/project.py
@@ -34,13 +34,13 @@ REGEX = re.compile(
 )
 
 
-def load_toml(filename: os.PathLike[str] | str) -> dict[str, Any]:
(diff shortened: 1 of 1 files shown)


paduszyk:
@henryiii Actually, aliasing the `load_toml` function to `load_pyproject` (e.g. via `functools.partial`) could also be an idea...

henryiii:
I'd rather keep the API surface minimal, I think? If the default is not obvious, we can also leave it, since `.load_toml("pyproject.toml")` is only marginally longer and doesn't require increasing the API to learn / use.

edgarrmondragon (approved):
I see no harm in doing this! 👍

PR #920, opened 2025-01-31
fix: handle uvx like uv

henryiii:
See https://github.com/wntrblm/nox/pull/818#issuecomment-2628118612.

### DIFF
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -608,14 +608,17 @@ class Session:
         if callable(args[0]):
             return self._run_func(args[0], args[1:])  # type: ignore[unreachable]
 
-        # Using `"uv"` when `uv` is the backend is guaranteed to work, even if it was co-installed with nox.
-        if (
-            self.virtualenv.venv_backend == "uv"
-            and args[0] == "uv"
-            and nox.virtualenv.UV != "uv"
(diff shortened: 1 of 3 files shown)


edgarrmondragon (commented):
I'm still seeing the warning

```
nox > Warning: uvx is not installed into the virtualenv, it is located at /Users/me/.local/bin/uvx. This might cause issues! Pass external=True into run() to silence this message.
```

I used `uvx --with git+https://github.com/wntrblm/nox.git@refs/pull/920/head nox -rs <session>`

PS: I also had to change `nox.options.sessions` to be a `list` since it was erroring with a tuple, so we should call it out in the future release notes.

henryiii:
Oops, I thought about that when I started, but forgot about it by the time I added a test. I'll add it.

henryiii:
Where was `nox.options.sessions` a tuple?

edgarrmondragon:
> Where was `nox.options.sessions` a tuple?

In my `noxfile.py`. I see the examples in https://nox.thea.codes/en/stable/config.html all use lists, so maybe only I was in the wrong 😅

henryiii:
Can you try now? I'm looking into the over-validtion issue.

edgarrmondragon:
> Can you try now? I'm looking into the over-validtion issue.

Aha, that did it!

PR #921, opened 2025-01-31
fix: over validation of lists

henryiii:
Lists were being over-validated when we added validation (see #920), `nox.options.sessions = ("a", "b")` should be allowed. I don't want to allow `nox.options.sessions = "ab"`, though, since that's surprising and would only work for one-letter sessions. ~~I went with just validating both lists and tuples~~, though now that I think of it, maybe sets should be supported too. There's also the option of just validating for Sequence + not(str), but that can't be represented in static types (first commit), as far as I know. Edit: reverted to the original Sequence idea, `set` and `frozenset` (at least) should be supported too.

### DIFF
--- a/nox/_option_set.py
+++ b/nox/_option_set.py
@@ -30,7 +30,7 @@ import attrs
 import attrs.validators as av
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable
+    from collections.abc import Callable, Iterable, Sequence
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


henryiii:
@edgarrmondragon Does this look fine?

edgarrmondragon (approved):

edgarrmondragon:
> @edgarrmondragon Does this look fine?

Yup, this works!

PR #922, opened 2025-01-31
fix: support broken uv (via pyenv)

### LINKED ISSUE #919: nox fails if uv in PATH is a pyenv shim
### Current Behavior

[pyenv]() uses shim scripts that live in a directory in PATH. These scripts either trigger the executable for the loaded python version or raise an error if the script is not installed in the "activated" python version.

This leads to broken nox builds, since nox automatically uses `uv` if it finds it in your PATH, even if the uv is a shim script. It would be nice to detect this because venv builds run fine. I worked around this by ufing `-fb venv`.

### Expected Behavior

Nox somehow infers that the uv in the user's PATH is broken and doesn't use it.

### Steps To Reproduce

```
$ pyenv global 3.12.8
$ pip install uv
$ pyenv global 3.13.1
$ nox
nox > Running session tests
nox > Re-using existing virtual environment at .nox/tests.
nox > uv pip install coverage
nox > Command uv pip install coverage failed with exit code 127:
pyenv: uv: command not found

The `uv' command exists in these Python versions:
  3.12.8

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
```

### Environment

```Markdown
- OS: MacOS
- Python: Any two python versions installed via PyEnv with uv installed via pip in one environment
- Nox: 2024.10.9
```

### Anything else?

_No response_

henryiii:
Fix #919 by moving the `uv` version check up to the discovery phase, and skip discovery if the version step returns an error code. If we know what the minimum version of uv we support is, we could add that here.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -44,7 +44,7 @@ __all__ = [
     "HAS_UV",
     "OPTIONAL_VENVS",
     "UV",
-    "UV_PYTHON_SUPPORT",
+    "UV_VERSION",
     "CondaEnv",
     "InterpreterNotFound",
     "PassthroughEnv",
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Thanks! This fixes using `nox` with no arguments in my environment with a broken `uv` in `$PATH`.

PR #924, opened 2025-02-04
ci: trusted publisher

henryiii:
This needs to be setup in PyPI as well. No need to make a `pypi` environment, it will be created automatically. Publishing a release triggers the deployment; you can add environment protection rules to requiressome number of maintainer approvals and such. Docs: https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/

You can download the SDist/wheel from any run, now, too.

Fields:

* Workflow name: `cd.yml`
* Environment name: `pypi`

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,49 @@
+name: CD
+on:
+  push:
+    branches: [main]
+  pull_request:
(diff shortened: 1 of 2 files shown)


theacodes (approved):
Alright, I got the PyPI side setup. Let's see how it goes.

PR #935, opened 2025-02-14
Never ignore URL dependencies in PEP 723 noxfiles

### LINKED ISSUE #934: URL dependencies are sometimes silently ignored in PEP 723 noxfiles
### Current Behavior

URL dependencies in a noxfile's PEP 723 `dependencies` are silently ignored in some cases.

### Expected Behavior

Nox should not ignore any PEP 723 dependencies.

### Steps To Reproduce

Minimal reproducer:

1.  `noxfile.py`:

    ```python
    # /// script
    # requires-python = ">= 3.9"
    # dependencies = [
    #     "nox @ git+https://github.com/wntrblm/nox.git@2024.10.09",
    # ]
    # ///

    import importlib.metadata

    import nox


    @nox.session(python=False)
    def a(session: nox.Session) -> None:
        print(importlib.metadata.version("nox"))


    if __name__ == "__main__":
        import nox.__main__

        nox.__main__.main()
    ```

2.  Run `nox`

    Behavior:

    ```console
    $ nox
    nox > Running session a
    2025.2.9
    nox > Session a was successful.
    ```

    Expected behavior:

    ```console
    $ nox
    nox > Running session a
    2024.10.09
    nox > Session a was successful.
    ```

    For comparison:

    ```console
    $ pipx run noxfile.py
    nox > Running session a
    2024.10.9
    nox > Session a was successful.
    ```

    ```console
    $ uv run -s noxfile.py
    nox > Running session a
    2024.10.9
    nox > Session a was successful.
    ```

### Environment

```Markdown
- OS: Arch Linux
- Python: CPython 3.13.2
- Nox: 2025.2.9
```

### Anything else?

Whether Nox ignores a requirement depends on both the requirement and on the environment of the calling `nox`. For example, if we modify the example above:

```diff
 # dependencies = [
-#     "nox @ git+https://github.com/wntrblm/nox.git@2024.10.09",
+#     "nox[uv] @ git+https://github.com/wntrblm/nox.git@2024.10.09",
 # ]
```

then now `nox` will no longer ignore the requirement _if_ `nox` is installed a venv without uv, but `nox` will still ignore the requirement if uv is installed in the same venv as it.

Another example:

```python
# /// script
# requires-python = ">= 3.9"
# dependencies = [
#     "nox >= 2025.2.9",
#     "httpx @ git+https://github.com/encode/httpx@0.28.0",
# ]
# ///

import importlib.metadata

import nox


@nox.session(python=False)
def a(session: nox.Session) -> None:
    print(importlib.metadata.version("httpx"))


if __name__ == "__main__":
    nox.main()
```

With this one, Nox will ignore the httpx requirement if (for example) `httpx==0.28.1` is installed in the same venv as `nox`.

gschaffner:
Fixes #934.

W.r.t. performance: On uv (the default, if available), uv caches based on the URL. On virtualenv however, the URL will get pulled and built/installed every time.

Ideally, for the virtualenv case, nox would do its own caching on the URL, similar to what the uv backend, `uv run -s noxfile.py` and `pipx run noxfile.py` all do. I have not included that here since I wanted to keep the scope of this PR to the correctness issue. Caching can be added later if needed. URL dependencies in PEP 723 noxfiles are perhaps not a very common case anyway, so perhaps it is TBD whether it's worth it for Nox to build a serialization format for doing said caching?

### DIFF
--- a/nox/_cli.py
+++ b/nox/_cli.py
@@ -106,6 +106,8 @@ def check_dependencies(dependencies: list[str]) -> bool:
             version = importlib.metadata.version(dep.name)
             if not dep.specifier.contains(version):
                 return False
+        if dep.url:
+            return False
 
     return True
 
--- a/nox/_cli.py
+++ b/nox/_cli.py
@@ -21,6 +21,7 @@ import os
 import shutil
 import subprocess
 import sys
+import urllib.parse
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NoReturn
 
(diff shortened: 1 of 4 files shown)


nox/_cli.py
```diff
@@ -106,6 +106,8 @@ def check_dependencies(dependencies: list[str]) -> bool:
             version = importlib.metadata.version(dep.name)
             if not dep.specifier.contains(version):
                 return False
+        if dep.url:
+            return False
```
henryiii, line 110: I think we could make this smarter, given the suggestion in build and importlib.metadata's `.origin`?

PR #941, opened 2025-03-11
feat: show skip reason by default

### LINKED ISSUE #940: Add option to display  reason for skipped sessions in the summary

In large projects, there can be multiple reasons why tests are skipped. For example, some tests might be skipped on Python 3.13 due to compatibility issues, while others could be skipped due to missing dependencies, platform-specific constraints, etc

Without a clear summary, developers often need to scroll through long test logs to find out why certain tests were skipped.

By adding a flag to display skip reasons in the final summary, developers can quickly understand why tests were skipped without digging through nox output. 

nox > Ran multiple sessions:
nox > * fail: failed
nox > * skip_3_13: skipped (Skip tests on python 3.13 due to compatibility issues)

_No response_

### Anything else?

_No response_

chirizxc:
/closes #940

### DIFF
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -643,6 +643,13 @@ options.add_options(
         hidden=True,
         finalizer_func=_color_finalizer,
     ),
+    _option_set.Option(
+        "skip_summary_details",
+        "--skip-summary-details",
+        group=options.groups["reporting"],
+        action="store_true",
+        help="Skip detailed summary output, such as reasons for skipped sessions.",
+    ),
     # Stores the original working directory that Nox was invoked from,
     # since it could be different from the Noxfile's directory.
     _option_set.Option(
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -643,13 +643,6 @@ options.add_options(
         hidden=True,
         finalizer_func=_color_finalizer,
     ),
-    _option_set.Option(
-        "skip_summary_details",
-        "--skip-summary-details",
-        group=options.groups["reporting"],
-        action="store_true",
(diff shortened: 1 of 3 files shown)


henryiii:
I don't know if this needs to be a flag. I think something like this would likely be fine:

```diff
-        result.log(f"* {name}: {status}")
+        if result.status is Status.SKIPPED and result.reason:
+            result.log(f"* {name}: {status} ({result.reason})")
+        else:
+            result.log(f"* {name}: {status}")
```

chirizxc:
> I don't know if this needs to be a flag. I think something like this would likely be fine:
> 
> ```diff
> -        result.log(f"* {name}: {status}")
> +        if result.status is Status.SKIPPED and result.reason:
> +            result.log(f"* {name}: {status} ({result.reason})")
> +        else:
> +            result.log(f"* {name}: {status}")
> ```



> I don't know if this needs to be a flag. I think something like this would likely be fine:
> 
> ```diff
> -        result.log(f"* {name}: {status}")
> +        if result.status is Status.SKIPPED and result.reason:
> +            result.log(f"* {name}: {status} ({result.reason})")
> +        else:
> +            result.log(f"* {name}: {status}")
> ```

maybe we can make this behaviour by default?

henryiii:
That's what I meant. If a reason is set, we could just always show it.

chirizxc:
idk why coverage failed, but here is successful: https://github.com/chirizxc/nox/actions/runs/14065702085/job/39389829080

henryiii (approved):

PR #942, opened 2025-03-12
docs: fix parametrized session tagging example

living180:
The final example in the "Assigning tags to parametrized sessions" section was using `nox.param()` incorrectly.  Fix by passing the parameter values as positional arguments (rather than as a tuple), and passing the tags as a keyword argument.


henryiii (approved):

henryiii:
Thanks!

PR #954, opened 2025-04-30
uv version is now uv self version

### LINKED ISSUE #953: uv support broken when recent release of uv 0.7.0
### Current Behavior

This session was working just earlier today, but with the fresh release of uv 0.7.0 it seems that the uv backend is broken, see error below.  This session is setup with `uv` as the only backend.  I have not had time to look into the changes in uv 0.7.0, but I did quickly notice that there are breaking changes in the [release notes](https://github.com/astral-sh/uv/releases/tag/0.7.0).

```python
$ nox --error-on-missing-interpreters --non-interactive --session build
nox > Running session build
nox > Session build raised exception ValueError("No backends present, looked for ('uv',).")
File "/usr/local/lib/python3.13/site-packages/nox/sessions.py", line 1097, in execute
self._create_venv()
~~~~~~~~~~~~~~~~~^^
File "/usr/local/lib/python3.13/site-packages/nox/sessions.py", line 1017, in _create_venv
self.venv = get_virtualenv(
~~~~~~~~~~~~~~^
*backends,
^^^^^^^^^^
...<3 lines>...
venv_params=self.func.venv_params,
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
)
^
File "/usr/local/lib/python3.13/site-packages/nox/virtualenv.py", line 755, in get_virtualenv
raise ValueError(msg)
ValueError: No backends present, looked for ('uv',).
nox > Session build failed.
```

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```Markdown
- OS:
- Python:
- Nox:
```

### Anything else?

_No response_

Spectre5:
Fixes #953

See https://github.com/astral-sh/uv/pull/12349

Note that I have not actually tested it on my previously failing environment from the issue.  Also, the test cases mock the call to uv version, so perhaps it would be worthwhile to add another test case without the mock to make sure uv is found?

Also, this will work with uv >= 0.7.0 but fail for versions before that, so this becomes a breaking change in nox.  Perhaps a maintainer would rather add in the logic to make nox work for uv versions both before and after 0.7.0.  If not, then `"uv>=0.1.6"` in the pyproject.toml file should be updated to `"uv>=0.7.0"`.

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -108,7 +108,7 @@ def uv_version(uv_bin: str) -> version.Version:
     """Returns uv's version defaulting to 0.0 if uv is not available"""
     try:
         ret = subprocess.run(
-            [uv_bin, "version", "--output-format", "json"],
+            [uv_bin, "self", "version", "--output-format", "json"],
             check=False,
             text=True,
             capture_output=True,
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -751,7 +751,7 @@ def test_find_uv(
 
     def mock_run(*args: object, **kwargs: object) -> subprocess.CompletedProcess[str]:
         return subprocess.CompletedProcess(
-            args=["uv", "version", "--output-format", "json"],
+            args=["uv", "self", "version", "--output-format", "json"],
             stdout=f'{{"version": "{vers}", "commit_info": null}}',
             returncode=vers_rc,
         )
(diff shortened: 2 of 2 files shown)


alfiedouglas0 (approved):
🙌

henryiii:
Ahh, sorry, didn't look for an existing PR before making #955! I did handle both cases, since the point of checking the version is to handle older versions.

Spectre5:
Closing, See #955

henryiii:
I cherry-picked your test changes, thanks! :)

PR #955, opened 2025-04-30
fix: uv version is now uv self version, support UV

henryiii:
We don't support uv 0.7. This fixes it and also supports the UV environment variable, which wasn't something UV had before. (This fixes nested uv calls, uv sets UV when it calls anything)


Spectre5:
@henryiii can we get a release for this?

henryiii:
Yes, tonight or tomorrow. Trying to quickly fix a few of the most recent open issues, specifically the "broken with newest condas" one. I've updated my conda to test it and now I need to it too :).

PR #957, opened 2025-04-30
fix: conda_install issue with newer conda

### LINKED ISSUE #951: session.conda_install fails when specifying min/max version with newer conda
### Current Behavior

nox > Running session flake8
nox > Creating conda env in .nox/flake8 with python=3.9
nox > conda install --yes --prefix /home/jeff/workspace/nox_test/.nox/flake8 '"flake8>=4.0.1"'
nox > Command conda install --yes --prefix /home/jeff/workspace/nox_test/.nox/flake8 '"flake8>=4.0.1"' failed with exit code 1:

InvalidMatchSpec: Invalid spec '"flake8>=4.0.1"': Invalid version '4.0.1"': invalid character(s)

nox > Session flake8 failed.

### Expected Behavior

I expect the package to be installed when specifying a min or max version

### Steps To Reproduce

noxfile.py:
```
import nox

@nox.session(python='3.9', venv_backend='conda')
def flake8(session):
    session.conda_install('flake8>=4.0.1')
    session.install('--no-deps', '-e', '.')
    session.run('flake8')
```

1. Install a new miniconda version (https://repo.anaconda.com/miniconda/Miniconda3-py312_25.1.1-2-Linux-x86_64.sh)
2. conda create -n nox python=3.11 nox
3. conda activate nox -y
4. nox

### Environment

```Markdown
- OS: Ubuntu 24.04.2
- Python: 3.12.9
- Nox: 2025.2.9
- Conda: 25.1.1
```

### Anything else?

Temporary workaround:
`nox.sessions._dblquote_pkg_install_args = lambda args: args`

You appear to routing this through [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#security-considerations) which states:

> Unlike some other popen functions, this library will not implicitly choose to call a system shell. This means that all characters, including shell metacharacters, can safely be passed to child processes.

I believe that this quoting can be removed completely: https://github.com/wntrblm/nox/blob/main/nox/sessions.py#L726-L727

Fails with:
- conda 25.1.1
- conda 24.11.1

Works with:
- conda 24.9.2

henryiii:
Fix #951.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,14 @@ jobs:
         os: [ubuntu-22.04, windows-latest, macos-13]
         python-version:
           - "3.8"
-          - "3.9"
           - "3.10"
-          - "3.11"
           - "3.12"
           - "3.13"
         include:
+          - os: ubuntu-24.04
+            python-version: "3.9"
+          - os: ubuntu-24.04
+            python-version: "3.11"
           - os: macos-14
             python-version: "3.12"
     steps:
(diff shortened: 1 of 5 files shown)


henryiii:
I don't understand why this fails on Windows. This is supposed to be supported on Windows without extra escaping as far as I can tell: `subprocess.run(["pip", "install", "numpy<2"])`, but it seems like that's pretty much what's failing here. And adding quotes causes modern Conda to error out due to the quotes making it through.

henryiii:
This is strictly better on Unix, and Windows just does the same thing (which works on older Conda's, and hopefully mambas).

PR #962, opened 2025-05-01
fix: correct minimum versions and test

### LINKED ISSUE #961: 2025.05.01 breaks with certain compatible versions of `attrs` installed
### Current Behavior

Receive the following error when invoking nox 2025.05.01: `AttributeError: module 'attrs.validators' has no attribute 'or_'` 
`attrs==23.1` is installed, which is compatible with nox dependency `>=23.1`
This is because as of [PR 932](https://github.com/wntrblm/nox/pull/932) nox uses `.or_` which was not introduced til attrs v24.1.0

Image

### Expected Behavior

Nox can be invoked with all versions consistent with dependency >=23.1 (including attrs < 24.1 which is affected by this issue), or bump attrs dependency

### Steps To Reproduce
Install attrs 23.1
Install nox 2025.05.01
Run nox

henryiii:
Fix #961. Test the minimums and update to fix tests.

### DIFF
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -1015,9 +1015,16 @@ def test_inner_functions_reusing_venv(
     venv, location = make_one(reuse_existing=True)
     venv.create()
 
+    txt = location.joinpath("pyvenv.cfg").read_text(encoding="utf-8")
+    cfg = {
+        (v := line.partition("="))[0].strip(): v[-1].strip()
+        for line in txt.splitlines()
+    }
+    home = cfg["home"]
+
     # Drop a venv-style pyvenv.cfg into the environment.
-    pyvenv_cfg = """\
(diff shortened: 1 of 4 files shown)


tests/test_virtualenv.py
```diff
@@ -1015,9 +1015,16 @@ def test_inner_functions_reusing_venv(
...
+        (v := line.partition("="))[0].strip(): v[-1].strip()
+        for line in txt.splitlines()
+    }
+    home = cfg["home"]
+
     # Drop a venv-style pyvenv.cfg into the environment.
-    pyvenv_cfg = """\
-    home = /usr/bin
+    pyvenv_cfg = f"""\
+    home = {home}
```
henryiii, line 1027: FWIW, this also fixes an issue I've seen a lot running locally.

PR #967, opened 2025-05-04
fix(types): tighten type for venv_backend

henryiii:
See https://github.com/dantebben/nox-uv/issues/23. We clearly assume it's a string and it seems like typing passes with it tightened.


johnthagen (approved):

dantebben:
I'll be sure that's in the next release.

PR #970, opened 2025-05-07
chore: add 3.14 beta 1 support

henryiii:
Let's see how it goes!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           - "3.8"
           - "3.10"
           - "3.13"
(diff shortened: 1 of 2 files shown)


henryiii:
Even though I got the notification, it seems setup-python isn't picking up 3.14b1 yet. We'll need to skip conda, of course, but I want to ensure we work with beta 1 before declaring support.

henryiii:
We support it but tox doesn't, so it breaks our tox tests. :)

henryiii:
Also, unrelated breakage from https://github.com/snowballstem/snowball/issues/229.

PR #971, opened 2025-05-12
docs: update badge from black to ruff

shenxianpeng:
Since black was removed in #745, it would be better to update the corresponding badge as well.


johnthagen:
Ruff has a badge if you want to replace it

- https://github.com/astral-sh/ruff?tab=readme-ov-file#ruff

PR #972, opened 2025-05-13
Recommend `--locked` over `--frozen` for `uv sync`

johnthagen:
This tracks upstream recommendations

- https://github.com/astral-sh/uv/pull/12818
- https://github.com/astral-sh/uv/pull/12819

To be honest, I actually think that we should put `--locked` into the default code example above this paragraph. An out-of-date lockfile is something you almost always want to catch in CI or locally with a tool like Nox.

@henryiii what do you think?

For now, this PR takes the more conservative change of simply recommending `--locked` over `--frozen`.

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -140,7 +140,7 @@ If you use a tool like ``uv`` to lock your dependencies, you can use that inside
         session.run("pytest", *session.posargs)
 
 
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):
This makes sense. I recently found that I was using `UV_PRERELEASE=allow` in CI but not locally.

henryiii:
What is the default? And what happens if there isn't a lockfile yet? I think the default is like `--frozen` except it creates the lockfile the first time it runs, which is probably useful if people are just using nox and not running uv directly.

Issue #409, opened 2021-03-25
Ability to get a separate log file for each session, containing the full detailed command logs (not the level from the logger)


smarie:
On projects with quite a number of nox sessions, it is a bit annoying if several sessions fail but several other succeed, to get the logs of all failed sessions. 

The same will definitely happen when sessions can run in parallel (See https://github.com/theacodes/nox/issues/198 )

In my projects I came up with a quick and dirty prototype where the full log (including stdout and stderr of all commands) of each session is written to an independent file. When the session ends, the file is renamed either as `SUCCESS_...` or `ERROR_...` as shown below:

![image](https://user-images.githubusercontent.com/3236794/112484656-6bdb6900-8d7a-11eb-9452-ccfa22a486a7.png)

This would be an increadibly useful feature in my opinion. To reach this I had to make a change in the `popen` routine in order to use `asyncio`. That way I have the possibility to create an os-independent `tee` of the stdout and stderr streams to both the log file and the general nox logger.

It works pretty well for six of my projects for now.

For example in [this nox file](https://github.com/smarie/python-pytest-cases/blob/main/noxfile.py)

Let me know if you think that such feature could be relevant and if so, how you would think of integrating it in `nox`.
Best, 
Sylvain

theacodes:
Oh I like this idea a lot.

smarie:
To make a step towards resolution, here is an attemps to specify the feature

 - a new option `nox.options.logsdir` with default value `None` meaning `".nox/_runlogs"` (note: maybe `.nox/` comes from another parameter, I need to check). Setting this option to `False` would disable log files. Setting this to a valid path string changes the folder. 
 
 - The folder does not need to exist, it is created automatically if it doesn't. However if it exists, it needs to be empty

 - If this option is enabled, a new `FileHandler` is added to the root nox logger (`logging.getLogger("nox")`). 

 - The stream open by the file handler is shared with `popen` (that is the underlying method used to execute all commands). This method is rewritten (as shown in my repository above) using asyncio in order to "tee" the stdout/stderr *both* to the log file stream, **and** to the string in memory for possible appending to the log (depending on `silent` and the level).
 
[...]

PR #974, opened 2025-05-15
fix: log Output of Failed Process

jbdyn:
Hi there :wave:,

I started using `nox` for matrix testing in [`elva`](https://github.com/innocampus/elva) and got a working `noxfile.py` in no time.
Very nice :heart: 

Since there are many sessions, I want to write session-specific logs.

However, there was no way to capture the output of a failed `pytest` run to see what went wrong.

So, here is my patch to overcome this.

Basically, the output of a failed process is no longer written to `sys.stderr` solely but instead by the session logger via `logger.error`.
Before, any output logging was prevented by a raised `CommandFailed` exception.

This whole story is somewhat related to issue #409.
The difference is that I didn't need to use `asyncio` in `popen`.

<details>
<summary>That is the current state of my <code>noxfile.py</code></summary>

```python
from itertools import product
from typing import Generator, Iterable
import logging
import sys
from pathlib import Path
from datetime import datetime

import nox

def sanitize_path(path):
    for char in ": =().":
        path = path.replace(char, "-")
    return path

BACKEND = "uv|virtualenv"
EDITABLE = ("-e", ".[dev,logo]")
TIMESTAMP = sanitize_path(datetime.now().isoformat(timespec="minutes"))
LOG_PATH = Path(__file__).parent / "logs" / "nox" / TIMESTAMP
LOG_PATH.mkdir(parents=True, exist_ok=True)

PROJECT = nox.project.load_toml("pyproject.toml")

# from classifiers and not `requires-python` entry;
# see https://nox.thea.codes/en/stable/config.html#nox.project.python_versions
PYTHON = nox.project.python_versions(PROJECT)

# versions to test by compatible release;
# check for every version adding new functionality or breaking the API
WEBSOCKETS = ("13.0.0", "13.1.0", "14.0.0", "14.1.0", "14.2.0", "15.0.0")
TEXTUAL = ("1.0.0", "2.0.0", "2.1.0", "3.0.0", "3.1.0", "3.2.0")


def parameters_excluding_last(*params: Iterable[Iterable[str]]) -> Generator[nox.param, None, None]:
    """Genererate the products of parameters except for the last one."""
    latest = tuple(param[-1] for param in params)

    for prod in product(*params):
        if prod != latest:
            yield nox.param(*prod)



def set_log_file(path):
    SESSION_HANDLER = "nox-session"
    logger = logging.getLogger()

    for handler in logger.handlers:
        if handler.name == SESSION_HANDLER:
            logger.removeHandler(handler)

    handler = logging.FileHandler(path)
    handler.name = SESSION_HANDLER
    logger.addHandler(handler)


@nox.session(
    venv_backend=BACKEND,
)
@nox.parametrize(
    # exclude newest since this environment configuration is covered by the `coverage` session below
    ("python", "websockets", "textual"), parameters_excluding_last(PYTHON, WEBSOCKETS, TEXTUAL)
)
def tests(session, websockets, textual):
    LOG_FILE = LOG_PATH / f"{sanitize_path(session.name)}.log"
    set_log_file(LOG_FILE)

    # idempotent
    session.notify("coverage")

    # install from `pyproject.toml`
    session.install(*EDITABLE)

    # overwrite with specific versions;
    # for compatible release specifier spec,
    # see https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release
    session.install(
        f"websockets~={websockets}",
        f"textual~={textual}",
    )

    # TODO: run across all tests
    session.run(
        "pytest",
        "--log-file",
        LOG_FILE,
        "--log-file-mode=a",
        "tests/test_component.py",
        silent=True,
    )


@nox.session(
    venv_backend=BACKEND,
)
def coverage(session):
    LOG_FILE = LOG_PATH / f"{sanitize_path(session.name)}.log"
    set_log_file(LOG_FILE)

    # install from `pyproject.toml`;
    # make sure to install the latest possible versions since `uv` won't update otherwise
    if session.venv_backend == "uv":
        session.install("--exact", "--reinstall", *EDITABLE)
    else:
        session.install(*EDITABLE)

    # TODO: run across all tests
    session.run(
        "coverage",
        "run",
        "-m",
        "pytest",
        "--log-file",
        LOG_FILE,
        "--log-file-mode=a",
        "tests/test_component.py",
        silent=False
    )

    # generate reports
    session.run("coverage", "combine", silent=True)
    session.run("coverage", "report", silent=True)
    session.run("coverage", "html", silent=True)
```
</details>

What do you think?

### DIFF
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -259,7 +259,7 @@ def test_fail_with_silent(
             ],
             silent=True,
         )
-    out, err = capsys.readouterr()
+    _out, err = capsys.readouterr()
     assert "out" not in err
     assert "err" not in err
     assert "out" in caplog.text
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -195,6 +195,13 @@ class TestSession:
 
         assert session.venv_backend == "none"
 
+    def test_virtualenv_directory(self) -> None:
+        session, runner = self.make_session_and_runner()
+
+        with tempfile.TemporaryDirectory() as root:
+            runner.global_config.envdir = root
+            assert session.env_dir == Path(runner.envdir)
+
     def test_interactive(self) -> None:
         session, _runner = self.make_session_and_runner()
 


jbdyn:
> Is `session.env_dir` the same as `Path(session.virtualenv.location)`?

Yes, it is:

- `Session.virtualenv` is equivalent to `SessionRunner.venv`
- `SessionRunner.venv` is created via `get_virtualenv` passing `SessionRunner.envdir` as `location` argument for `VirtualEnv`, so `SessionRunner.envdir` is equivalent to `SessionRunner.venv.location`
- `Session.env_dir` is equivalent to `SessionRunner.envdir`
- By logic, that means that `Session.env_dir` is equivalent to `Session.virtualenv.location`

My definition of `Session.env_dir` has the advantage of linking directly to `SessionRunner.envdir`, which seems to be always available.
`Session.virtualenv.location` is not, as `Session.virtualenv` aka `SessionRunner.venv` needs to exist, else a `ValueError` is raised in `Session.virtualenv`.

But, I don't whether this is relevant in practice.

jbdyn:
@henryiii Thank you, too, for reviewing and adding the last test as well!

PR #979, opened 2025-06-12
fix(action): simpler action, better default on Windows ARM

henryiii:
* Drops validation code, just rely on underlying action
* Custom default on Windows ARM (3.11 minimum)
* Use the final item in the list for nox


henryiii:
@mayeut IIRC you worked on this before? Thoughts?

mayeut:
The only thing that worries me with the simplification is the order of CPython vs other interpreters: PyPy and/or GraalPy could take precedence and CPython won't be tested. At least, that was the case some time ago.
If this seems a burden that can be moved to the end-users of the action then so be it but it should be checked and the doc updated to reflect that the list should start by alternative interpreters if any.
This order thing should probably be discussed in [setup-python](https://github.com/actions/setup-python) bug tracker.

PR #983, opened 2025-06-13
Percolate the `verbose` global option to the `silent` argument for session installation commands, and document it

### LINKED ISSUE #741: Add

Currently `nox` does not notify the user of what packages were installed by `pip` on the use of `session.install()` syntax. It is possible to do so by using the `--verbose` command-line flag. There is #666 that is (somewhat) related. Adding a way to make nox session arguments verbose by default would help evade the use of `--verbose` and therefore display verbose output by default.

For example

```python
@nox.session(name="tests")
def run_tests(session):
    """Run the unit tests."""
    session.install("-e", ".[test]")
    session.run("pytest")
```

can be called with

```bash
nox -s tests --verbose
```

to display `pip`'s installation output when invoking a `nox` session. However, this use-case is not limited to just this, and displaying verbose outputs at the time of virtual environment creation can also help for debugging purposes.

It would be great to have an option in this form

```python
# perhaps kept False by default
nox.options.verbose = True
```

which can then apply to all sessions invoked inside a `noxfile`. If verbosity is not required for a particular session, it can be overridden for said session by `@nox.session(verbose=False)` which takes precedence over `nox.options`.

Another solution is to just make `pip` display complete output in all cases (`tox` environments have always done this, with or without a verbose command-line flag or option).

The current workaround for this is to use 

```python
 session.run_always("pip", "install", "-e", ".[test]")
```

as an alias to `session.install()`, which for users means that

1. `conda` or `pip` inside this command has to be specified explicitly which 
2. This will make it run even in the presence of `--no-install`, which does not help in the case where speedups are required (not that `pip` output is required anyway when not installing prerequisites or other dependencies).

### Anything else?

N/A

agriyakhetarpal:
This PR documents the existing `nox.options.verbose` option and makes `verbose` a flag pair. Based on its value, it flips the `silent` parameter for `session.install(<...>)` and `conda_install(<...>)` to expose outputs to stdout/stderr.

The standard rule for flag pairs applies (as I understand it):  the `silent` parameter if set takes precedence over the value of `nox.options.verbose`, and the `--verbose`/`--no-verbose` CLI arguments will take precedence over both the values of `nox.options.verbose` and of `silent` whether set or not.

Please let me know if this is not desirable; in that case, I can close this PR.

Thank you for your time!

Closes #741

> [!NOTE]
> My original ask with the linked issue was to do both things: add a `nox.options.verbose` option to be able to specify that in a Noxfile, and also to not need to set `session.install(silent=False)` for every command; i.e., also print the output of installation commands (i.e., in order to adhere to the "all output of all commands run is shown, regardless of the silent argument" sentence in the docs by including these installation commands as a part of all commands).
>
> It appears that since then, the `verbose` option has been added; however, the installation commands were not taking this option's value into account, and this PR only addresses this part.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -26,7 +26,6 @@ jobs:
           - ubuntu-22.04
           - ubuntu-24.04
           - ubuntu-24.04-arm
-          - windows-2019
           - windows-2022
           - windows-2025
           - macos-13
@@ -48,8 +47,8 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: 3.12
       - uses: ./
         with:
-          python-versions: "3.4, 3.5, 3.6, 3.7, 3.8, 3.9, 3.10, 3.11, 3.12, 3.13, pypy-2.7, pypy-3.7, pypy-3.8, pypy-3.9, pypy-3.10, pypy-3.11"
+          python-versions: "pypy-2.7, pypy-3.8, pypy-3.9, pypy-3.10, pypy-3.11, 3.6, 3.7, 3.8, 3.9, 3.10, 3.11, 3.12, 3.13, 3.14"
       - run: nox --session github_actions_all_tests
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -147,7 +147,7 @@ By default, ``nox-uv`` also validates that the lockfile is up-to-date.
 
 .. code-block:: python
 
-    #!/usr/bin/env -S uv run -q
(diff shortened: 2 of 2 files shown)


agriyakhetarpal:
Thanks for triggering the CI! I've fixed the docs build.

docs/usage.rst
```diff
@@ -434,14 +434,18 @@ However, this will never output colorful logs:
     nox --nocolor
 
 
-.. _opt-report:
```
agriyakhetarpal, line 437: If I understand correctly, this link was placed at the wrong location and should have been at the "Outputting a machine-readable report" section, so I've moved it there.

Issue #986, opened 2025-06-18
nox installs Python using uv after setup-python in Windows


illia-v:
### Current Behavior

In urllib3's CI (GitHub Actions), we noticed that nox installs Python using uv in Windows even when the standard [setup-python](https://github.com/actions/setup-python) is used.

I can see that nox checks for `python3.13` using `shutil.which` when, for example, it has to use Python 3.13
https://github.com/wntrblm/nox/blob/50a740adcdf3db07d4fa2cda0d8549962d71a2ec/nox/virtualenv.py#L625-L633

But there is no `python3.13` executable in the Windows installation, so nox installs Python using uv.

<details>
<summary>Output of dir "%Python3_ROOT_DIR%"</summary>

```
Run dir "%Python3_ROOT_DIR%"
  dir "%Python3_ROOT_DIR%"
  for %%I in (python.exe) do mklink %%~dpIpython3.13.exe %%~dpIpython.exe
  shell: C:\Windows\system3[2](https://github.com/illia-v/urllib3/actions/runs/15720646707/job/44300706005#step:10:2)\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    FORCE_COLOR: 1
    pythonLocation: C:\hostedtoolcache\windows\Python\[3](https://github.com/illia-v/urllib3/actions/runs/15720646707/job/44300706005#step:10:3).13.5\x64
    PKG_CONFIG_PATH: C:\hostedtoolcache\windows\Python\3.13.5\x6[4](https://github.com/illia-v/urllib3/actions/runs/15720646707/job/44300706005#step:10:4)/lib/pkgconfig
    Python_ROOT_DIR: C:\hostedtoolcache\windows\Python\3.13.[5](https://github.com/illia-v/urllib3/actions/runs/15720646707/job/44300706005#step:10:5)\x64
    Python2_ROOT_DIR: C:\hostedtoolcache\windows\Python\3.13.5\x[6](https://github.com/illia-v/urllib3/actions/runs/15720646707/job/44300706005#step:10:6)4
    Python3_ROOT_DIR: C:\hostedtoolcache\windows\Python\3.13.5\x64
    UV_TOOL_BIN_DIR: C:\a\_temp\uv-tool-bin-dir
    UV_TOOL_DIR: C:\a\_temp\uv-tool-dir
    UV_CACHE_DIR: C:\a\_temp\setup-uv-cache
 Volume in drive C is Windows
[...]

Issue #944, opened 2025-03-24
Configure when `uv` should install Python


johnthagen:
### How would this feature be useful?

Support for `uv` installing Python was added in 

- https://github.com/wntrblm/nox/pull/842

But as discussed in https://github.com/wntrblm/nox/pull/842#issuecomment-2415903984 no configuration option was added to opt-out of this behaviour.

### Describe the solution you'd like

As mentioned in https://github.com/wntrblm/nox/pull/842#issuecomment-2384742104 add new configuration options:

```
--download-python=auto,always,never
```

Additionally, this should be configurable within a `noxfile.py`:

```py
import nox

nox.options.download_python = "never"  # type: Literal["auto", "always", "never"]
```

This feature should also be documented.

johnthagen:
CC @saucoide

rsyring:
Does [nox-uv](https://github.com/dantebben/nox-uv) have the functionality you are looking for?

PR #989, opened 2025-07-07
Add --download-python python option

saucoide:
Adding a --download-python / NOX_DOWNLOAD_PYTHON option to control the behavior of uv installing python, and introducing nox-installed pythons via pbs-installer for venv/virtualenv backends

The options are "auto" (default), "always" and "never".

- "auto" will first check for a pre-existing installation, and fall back to installing with pbs-installer (or uv in uv venv_backends)
- "always" will ignore system pythons and always try to install
- "never" will only use pre-existing python installs

this was discussed here https://github.com/wntrblm/nox/pull/842#issuecomment-2415903984 but i didnt have time until now, also this issue  https://github.com/wntrblm/nox/issues/944

I am not super confident this is a good idea in the end or not, the default "auto" behavior might be surprising if you use venv/virtualenv backend (could be considered breaking in some cases) like it happened when uv started installing python if missing and people did not expect it

and there is still an issue with windows that this would not fix but might help having the opt out https://github.com/wntrblm/nox/issues/986 

still todo: 
- ~~script-mode test breaks~~ - locally, works on ci?
- ~~coverage~~
- ~~documentation~~
- ~~a bunch more broken tests on ci~~
- ~~windows install paths are not handled correctly~~

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -125,7 +125,7 @@ When you provide a version number, Nox automatically prepends python to determin
     def tests(session):
         pass
 
-If the specified python interpreter is not found, Nox can automatically download it when ``--download-python`` is set to ``auto`` or ``always``.
+If the specified python interpreter is not found, Nox can automatically download it when ``--download-python`` is set to ``auto`` (the default) or ``always``. ``never`` avoids the download.
 
 When collecting your sessions, Nox will create a separate session for each interpreter. You can see these sessions when running ``nox --list``. For example this Noxfile:
 


saucoide:
@ dependencies, i'm not sure if to add the pbs-installer as an optional and have the whole behavior also depend on that, or if to add it by default

saucoide:
I'm also unsure about what to do with conda environments, right now they behave like if it was set to "always" and probably doesn't make sense to change it, but it'd be weird the option doesn't apply to them

saucoide:
no idea why this is producing an ungodly amount of logs, I haven't been able to reproduce the CI errors both on ubuntu & macos, pbs_installer fails in ubuntu when trying to install python 3.7, but it does not look anything like in ci :/

henryiii:
I was just mentioning this here at SciPy, thanks! :)

We have an issue with logs where it dumps logging closed messages. It's test specific, I think, possibly due to nox running nox. We've never managed to hunt down the root cause.

saucoide:
@henryiii do you think we should skip conda for this completely or try to have this toggle also control it? looks like it has always had its own separate behavior so i'm not sure

henryiii:
Let's make this optional; installing will be disabled if it's missing. Looks like `zstandard` doesn't ship PyPy or GraalPy wheels, too.

PR #991, opened 2025-07-20
CI ensure there are no duplicate log handlers

saucoide:
related: https://github.com/wntrblm/nox/pull/989#issuecomment-3054366659

nox tests run setup_logging() multiple times in a session and we end up with multiple duplicate handlers, some of which are already closed, fail and try to log the error, creating a cascade of logs

The change is making sure we get rid of our own handler before creating a new one


henryiii:
So that's it! I never considered duplicates when trying to debug this.

henryiii (approved):

Issue #949, opened 2025-04-07
Document support for the free-threaded build and add to tests


ngoldbaum:
### How would this feature be useful?

AFAIK, nox works fine on the free-threaded build, and I've been using it in the CI for several projects.

### Describe the solution you'd like

I notice that you aren't running tests using `3.13t`, it's probably worth adding that and seeing if anything is broken.

If that works, then it would be nice to officially document that you support the free-threaded build and that it should be referred to as e.g. `3.13t` with a `t` suffix.

henryiii:
This is the name it's trying to match: `'check_package_files-3.13(8.1.0)'`. We need to allow it to try to match without the `-3.13` if it's forced.

ngoldbaum:
@henryiii sorry - that PR looks unreated to this issue, did you really mean to close this issue?

henryiii:
Oops, must have got the wrong issue, thanks!

PR #999, opened 2025-09-27
correctly match free-threaded python versions

zzzeek:
in the case where nox is run given an explicit python version target to run on, nox locates Python interpreters by matching that interpreter string to a regex, that is then expanded to match the name of a Python executable by appending the numeric portions of the version to the word "python", its existence is verified with
shutil.which(), and is then passed off to virtualenv as a target to use.

However, if the explicit python version ends in "t" for a free-threaded python, such as "3.13t", the regex does not match, due to that "t" that's present which isn't included in the regex. 

This PR fixes the regex to also look for a "t" at the end for a free-threaded version name, and formats it into the interpreter name virtualenv will look for.

There's a bit of a secondary issue which is that if you are trying to select for "3.13" non-free-threaded, but you have a free-threaded 3.13t in your path as well, that "python3.13" name matches in the free-threaded bin folder as well, but virtualenv seems to know how to make the right choice in this case.

The only issue I could find regarding free-threading is #949 which vaguely claims that free-threading is working, however this patch is needed if you actually want to indicate the Python version.

Fixes: #949

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -622,10 +622,11 @@ class VirtualEnv(ProcessEnv):
 
         # If this is just a X, X.Y, or X.Y.Z string, extract just the X / X.Y
         # part and add Python to the front of it.
-        match = re.match(r"^(?P<xy_ver>\d(\.\d+)?)(\.\d+)?$", self.interpreter)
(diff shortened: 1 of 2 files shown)


zzzeek:
TIL I can also send the full executable name like `python3.13t` in, and that bypasses the regex issue here, but this still would be a good fix to have

Issue #1016, opened 2025-10-15
New `pbs-installer` dependency does not install well on PyPy


davidhewitt:
### Current Behavior

Latest release of `nox` (2025.10.14) includes a dependency of `pbs-installer`, which [itself depends on `zstandard`](https://github.com/frostming/pbs-installer/blob/main/pyproject.toml#L21).

Sadly `zstandard` is a native extension which does not ship PyPy or GraalPy wheels. We can probably work around this by introducing something like `setup-uv` with better caching for PyPy installs, however just wanted to flag this in case you didn't expect this downstream pain.

e.g. https://github.com/PyO3/pyo3/actions/runs/18522724388/job/52787311984?#step:5:58

```
Collecting zstandard>=0.21.0 (from pbs-installer>=2025.1.6->pbs-installer[all]>=2025.1.6->nox[uv])
  Downloading zstandard-0.25.0.tar.gz (711 kB)
     ---------------------------------------- 711.5/711.5 kB 14.0 MB/s  0:00:00
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [38 lines of output]
      Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35217 for x64
      Copyright (C) Microsoft Corporation.  All rights reserved.
      
      tmp2ghq70pz.h
      Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35217 for x64
      Copyright (C) Microsoft Corporation.  All rights reserved.
      
      tmpg1ewy050.h
      Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35217 for x64
      Copyright (C) Microsoft Corporation.  All rights reserved.
      
      tmpl9c2kjkt.h
      Traceback (most recent call last):
        File "C:\hostedtoolcache\windows\PyPy\3.9.19\x86\Lib\cffi\_shimmed_dist_utils.py", line 33, in <module>
          from distutils.msvc9compiler import MSVCCompiler
      ModuleNotFoundError: No module named 'distutils.msvc9compiler'
      
      The above exception was the direct cause of the following exception:
      
      Traceback (most recent call last):
[...]

PR #1017, opened 2025-10-15
fix: make pbs an optional dependency

henryiii:
Fix for #1016, followup to #989.


johnthagen (approved):
I like this. Especially since many users will already have and be using `uv`, so this isn't needed in their case.

PR #1062, opened 2026-02-10
Reapply "feat: drop Python 3.8 (#1004)" (#1060)

henryiii:
This reverts commit 90a8aef54122e188dbea80680552507517523ced.

(Except for coverage core, which isn't needed anymore, coverage changed the default.)


henryiii:
Seems quiet, so going back in.

PR #1086, opened 2026-04-16
feat: use python-discovery

henryiii:
This enables version ranges to be used. It also can find Pythons from a variety of sources that we didn't support before.

Two user facing changes:

1. The interpreter selected can sometimes be different.
2. Putting the name of an interpreter on the path there the discovery doesn't select won't pick it - full paths required for that case.

* We could check matching names first, which would shortcut the check and behave a bit more like previous (probably affecting 1 above, supporting 2).
* We should also support script python-requires using this

:robot: ~~Assisted-by: OpenCode:Kimi-K2.5~~ (First draft)

Assisted-by: ClaudeCode:claude-opus-4.8

### DIFF
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -34,6 +34,36 @@ import nox
 import nox.command
 from nox.logger import logger
 
+
+def _python_discovery_get_interpreter() -> Any:
+    """Lazy import of get_interpreter from python-discovery."""
+    import python_discovery  # noqa: PLC0415
+
+    return python_discovery.get_interpreter
+
+
+def _python_discovery_DiskCache() -> Any:
+    """Lazy import of DiskCache from python-discovery."""
+    import python_discovery  # noqa: PLC0415
+
+    return python_discovery.DiskCache
+
+
+def _get_python_discovery_cache() -> Any:
+    """Get or create the DiskCache for python-discovery.
+
+    Returns:
+        DiskCache instance or None if caching should not be used
+    """
+    cache_dir = Path.home() / ".cache" / "nox" / "python-discovery"
+    try:
+        disk_cache_cls = _python_discovery_DiskCache()
+        return disk_cache_cls(root=cache_dir)
+    except OSError:
+        # If cache directory can't be created, return None (no caching)
+        return None
+
+
 if TYPE_CHECKING:
     from collections.abc import Callable, Mapping, Sequence
 
@@ -111,10 +141,45 @@ def find_uv() -> tuple[bool, str, version.Version]:
     )
 
 
+def _is_version_range_spec(spec: str) -> bool:
+    """Check if the spec is a version range spec (e.g., '>=3.11,<3.13')."""
+    # Check for PEP 440 version comparison operators
+    return any(op in spec for op in [">=", ">", "<=", "<", "~=", "==", "!=", ","])
+
+
 def _find_python(interpreter: str, xy_ver: str) -> str | None:
-    """Find a python executable matching the requested interpreter"""
+    """Find a python executable matching the requested interpreter.
+
+    Uses traditional methods first for backwards compatibility, with python-discovery
+    as a fallback for advanced version specs or when traditional methods fail.
+    """
+    # For version range specs (e.g., ">=3.11,<3.13"), use python-discovery directly
+    if _is_version_range_spec(interpreter):
+        try:
+            cache = _get_python_discovery_cache()
+            get_interpreter = _python_discovery_get_interpreter()
+            result = get_interpreter(interpreter, cache=cache)
+            if result is not None:
+                return str(result.executable)
+        except (OSError, ValueError, RuntimeError):
+            # Fall back to traditional discovery on error
+            pass
+        return None
+
+    # Try traditional discovery first for backwards compatibility
     if shutil.which(interpreter):
         return interpreter
+
+    # Fall back to python-discovery for other interpreters that weren't found
+    try:
+        cache = _get_python_discovery_cache()
+        get_interpreter = _python_discovery_get_interpreter()
+        result = get_interpreter(interpreter, cache=cache)
+        if result is not None:
+            return str(result.executable)
+    except (OSError, ValueError, RuntimeError):
+        # Fall back to platform-specific discovery on error
+        pass
 
     # Windows only search for the executable
     if _PLATFORM.startswith("win"):
@@ -776,6 +841,14 @@ class VirtualEnv(ProcessEnv):
             xy_version = match.group("xy_ver")
             t = match.group("t")
             cleaned_interpreter = f"python{xy_version}{t}"
+
+        # For version range specs (e.g., ">=3.11,<3.13"), check directly
+        # This enables new functionality for specifying version ranges
+        if _is_version_range_spec(self.interpreter) and (
+            resolved := _find_python(self.interpreter, xy_version)
+        ):
+            self._resolved = resolved
+            return self._resolved
 
         match self.download_python, self.venv_backend:
             # never -> check for interpreters
--- a/nox/virtualenv.py
+++ b/nox/virtualenv.py
@@ -166,7 +166,17 @@ def _discover_interpreter(spec: str) -> PythonInfo | None:
     """
     from python_discovery import get_interpreter  # noqa: PLC0415
 
-    return get_interpreter(spec, cache=_get_python_discovery_cache())
+    cache = _get_python_discovery_cache()
+    if cache is None:
+        return get_interpreter(spec)
+    try:
+        return get_interpreter(spec, cache=cache)
+    except OSError:
+        # The cache dir exists but isn't writable (e.g. a read-only home):
+        # DiskCache touches the filesystem only lazily, so the failure surfaces
+        # here. Degrade to uncached discovery instead of aborting resolution.
+        logger.debug("python-discovery cache is not writable; discovering without it.")
+        return get_interpreter(spec)
 
 
 @functools.cache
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
   "dependency-groups>=1.1",
   "humanize>=4",
   "packaging>=22",
+  "python-discovery>=0.1",
```
felixfontein: Since the first release on PyPI is 1.0.0, shouldn't this be
```suggestion
  "python-discovery>=1.0.0",
```
?

felixfontein (commented):
Thanks for working on this! I'm really lloking forward to using this :) Is there anything we (community) can help with to bring this forward?

I looked at the CI faliures, but I don't know conda at all, so I don't know what exactly is going wrong in the two failed tests.

nox/virtualenv.py
```diff
@@ -34,6 +34,36 @@
...
+    return python_discovery.DiskCache
+
+
+def _get_python_discovery_cache() -> Any:
+    """Get or create the DiskCache for python-discovery.
+
+    Returns:
+        DiskCache instance or None if caching should not be used
+    """
+    cache_dir = Path.home() / ".cache" / "nox" / "python-discovery"
```
felixfontein: Regarding cache location prefix: I'd use `$XDG_CACHE_HOME / "nox"` first on Unix (https://specifications.freedesktop.org/basedir/latest/#basics), and then fall back to `~/Library/Caches/nox` on macOS and `~/.cache/nox` on other Unixes.

For Windows, one can get hold of the user cache directory from the registry. There's also [platformdirs](https://pypi.org/project/platformdirs/) which handles all this.

henryiii:
Currently my time is being gobbled up by scikit-build-core (in preparation for a 1.0 release). I can rebase this and retry, though, and try to get it into a testable condition.

felixfontein (commented):
Regarding disk cache usage: I think this is a good idea! I looked a bit into python-discovery/'s cache implementation, and it seems quite sensible. It checks mtime and hash of interpreters before re-using the cached data, to avoid stale information when interpreters are updated.

nox/virtualenv.py
henryiii: Thanks, that's helpful!

henryiii:
I redid this with Claude Opus using your suggestion, then rubber ducked against GPT 5.5.

This is going to be a bit tricky to review, though, since it changes both code and tests. I'd appreciate a review, that would help.

<details><summary>:robot: What has changed (user facing) :robot:</summary>
<p>

I now have enough to answer precisely. Here's the breakdown.

## New user-facing capabilities

1. **PEP 440 version ranges** in `python=` — `@nox.session(python=">=3.14")`, `">=3.11,<3.13"`, `"==3.12.*"`. Nox picks the first installed interpreter that satisfies the spec. **venv / virtualenv / uv only** (conda backends pass the spec through unchanged, so ranges don't work there — documented).
2. **Range auto-download**: with `--download-python auto/always`, a range installs the floor of its lowest bound (`>=3.14` → downloads `3.14`). A bound with no floor (`<3.14`, `!=3.12`) can't be downloaded and errors.
3. **Broader interpreter discovery**: now also searches version managers (pyenv, mise, asdf, uv) and the Windows registry (PEP 514), in addition to PATH and the `py` launcher.
4. **Two new runtime dependencies**: `python-discovery>=1` and `platformdirs>=2` (platformdirs was already transitive). A discovery cache is written under the OS cache dir (`~/.cache/nox/python-discovery`, etc.).

## What could stop working

The whole interpreter-resolution algorithm in `_find_python` was swapped from `shutil.which` + `py.exe`/PATH-version-probe to `python-discovery`. Two concrete regressions:

1. **Arbitrary-named interpreters on PATH no longer resolve.** The old code did `shutil.which(interpreter)` for *any* string, so `python="mypython312"` (a custom-named binary on PATH) would be found and used. python-discovery only recognizes Python-shaped specs (versions, `pythonX.Y`, `pypyX.Y`, paths) — I verified `get_interpreter("bash")` returns `None`. **Workaround:** pass the absolute path (`python="/path/to/mypython312"`), which still works.

2. **Interpreter *selection* may differ when several match.** When multiple interpreters satisfy a spec (e.g. a pyenv 3.12 *and* a PATH `python3.12`), python-discovery may choose a different one than `shutil.which` did — and it now considers sources `shutil.which` never looked at (version managers). The chosen interpreter is also now reported as a **full path** rather than a bare name like `python3.12`.

## Not affected
- All previously-supported spec forms still work: `"3.12"`, `"python3.12"`, `"3.13t"` (free-threaded), `"pypy3.11"`, `"3.6-32"` (Windows 32-bit), absolute paths — I confirmed `3.x-32` and path forms parse.
- `--download-python` semantics (never/auto/always, `[pbs]` extra) are unchanged for concrete versions.
- Conda/mamba/micromamba backends are untouched.

The selection-difference (#2) is the one most likely to surprise someone in CI; it's inherent to replacing the discovery engine and is the risk to watch as the matrix runs. Want me to note both regressions in the PR description so reviewers/users see them?

</p>
</details> 

Also, I went all in in `python-discovery` this time, instead of having an exact match on path being used immediately. That does have some effects (see above), maybe this should still be included and short circuit the search?

felixfontein:
I played around with this a bit, it works great! The only obvious downside is some excessive logging, which can be shut off like this:
```diff
diff --git a/nox/logger.py b/nox/logger.py
index c7b37c2..5e82c8f 100644
--- a/nox/logger.py
+++ b/nox/logger.py
@@ -157,3 +157,5 @@ def setup_logging(
     logging.getLogger("sh").setLevel(logging.WARNING)
     logging.getLogger("httpx").setLevel(logging.WARNING)
     logging.getLogger("httpcore").setLevel(logging.WARNING)
+    logging.getLogger("filelock").setLevel(logging.INFO)
+    logging.getLogger("python_discovery").setLevel(logging.WARNING)
```
(`filelock` was quiet for me with level `INFO`, but maybe it should be set to `WARNING` as well for consistency.)

Regarding the regressions:
1. I think the first one is *probably* OK. My *guess* is that this was a rather esoteric feature used only by very few people, so switching to absolute paths should be OK.
2. This is more serious, but IMO also OK. My *guess* is that most users want nox to find *a* Python 3.14 (for example) interpreter, and don't mind which one is found if they have multiple different ones installed. But this is definitely something that should be mentioned in the release notes.

> The chosen interpreter is also now reported as a **full path** rather than a bare name like `python3.12`.

Where exactly is that reported? I wasn't able to see it, also with extra verbosity (`-v`, `-vv`).

> Want me to note both regressions in the PR description so reviewers/users see them?

I would definitely do that, though likely in reverse order (the more serious one first).

felixfontein:
I read through all changes, and didn't spot something else to comment on.

henryiii:
Before:

```console
$ uv run nox -v -s tests-3.10
nox > Creating virtual environment (uv) using python3.10 in .nox/tests-3-10
Using CPython 3.10.20
Creating virtual environment at: .nox/tests-3-10
Activate with: source .nox/tests-3-10/bin/activate.fish
```

After:

```console
$ uv run nox -v -s tests-3.10
nox > Creating virtual environment (uv) using python3.10 in .nox/tests-3-10
Using CPython 3.10.20 interpreter at: /Users/henryfs/.local/bin/python3.10
Creating virtual environment at: .nox/tests-3-10
Activate with: source .nox/tests-3-10/bin/activate.fish
```

(only visible in verbose mode, `-v`)

felixfontein:
Interesting, both before and after I get something like this:
```
$ nox -v -e ansible-test-sanity-devel
nox > Running session ansible-test-sanity-devel
nox > Creating virtual environment (virtualenv) using python in .nox/ansible-test-sanity-devel
created virtual environment CPython3.13.11.final.0-64-x86_64 in 152ms
  creator CPython3Posix(dest=/path/to/project/.nox/ansible-test-sanity-devel, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, via=copy, app_data_dir=/home/user/.cache/virtualenv)
    added seed packages: pip==25.0.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator,XonshActivator
```
I guess this is because `virtualenv` and not `uv` are used to create the virtual environments.

henryiii:
I think we should try to make a release soon, then put this in, so it has some time on main.

felixfontein:
Sounds like a good plan! That will give it more testing and exposure while keeping the risk low(ish).

felixfontein:
@henryiii thanks a lot for implementing this!

zzzeek:
just so you know this broke our CI where we still have some tests for python 3.7.   My own Claude here has isolated the issue in tox-dev/python-discovery and all that, we can post an issue there, but before I get into all that should I still be expecting nox to be able to run the Python 3.7 interpreter at all?    otherwise I would just pin an earlier version of nox on the CI environment in question until we can get rid of that 3.7 suite

zzzeek:
added https://github.com/tox-dev/python-discovery/issues/116

henryiii:
Good question. I know virtualenv doesn't support older Python versions. I'd say we try to support running 3.8+ (depending on the backend, I think uv handled 3.8+ fine), and would be fine to support older if reasonable.

Any details? What's breaking?

zzzeek:
python-discovery has a file _py_info.py which is invoked under the interpreter in question, and the file has a walrus operator in it

per claude

> Claude: With CPython 3.7.17 at `/opt/python3.7/bin/python3.7`, on `PATH`:
> 
> ```python
> from python_discovery import get_interpreter
> print(get_interpreter("3.7"))                          # None
> print(get_interpreter("/opt/python3.7/bin/python3.7"))  # RuntimeError
> print(get_interpreter("3.9"))                          # works
> ```
> 
> The spec form returns `None`; the absolute-path form raises. Both share one cause,
> visible at `INFO`/`DEBUG`:
> 
> ```
> DEBUG python_discovery._cached_py_info: get interpreter info via cmd:
>       /opt/python3.7/bin/python3.7 .../python_discovery/_py_info.py <hash> <hash>
> DEBUG python_discovery._cached_py_info: first subprocess attempt failed for
>       /opt/python3.7/bin/python3.7 (failed to query ... with code 1 err:
>   File ".../python_discovery/_py_info.py", line 183
>     if (tk_lib := tcl.eval("set tk_library")) and os.path.isdir(tk_lib):
>                ^
> SyntaxError: invalid syntax
> ), retrying
> INFO  python_discovery._cached_py_info: failed to query /opt/python3.7/bin/python3.7 ...
> ```
> 
> It probes `python3` and `python3.7`, twice each, fails identically every time, and
> drops the candidate. `get_interpreter("3.9")` succeeds against
> `/opt/python3.9/bin/python3` in the same process, so `PATH` scanning itself is fine —
> 3.7 is the only version affected, since `:=` is 3.8 syntax.

zzzeek:
for virtualenv we are using venv as the default_venv_backend.  it's included with each python interpreter

python 3.7 is not like the end-all be-all here but we (SQLAlchemy) are going to always run into this kind of thing if our python deprecation policy doesnt line up with that of nox / python-discovery, since the Python assembly line of beta->released->EOL is a pretty constant churn now

PR #1096, opened 2026-05-31
fix(project): return empty dict from load_toml when script file missing

### LINKED ISSUE #1094: `FileNotFoundError` when no `noxfile.py` is present
### Current Behavior

Running `nox` in a directory with no `noxfile.py` raises an unhandled `FileNotFoundError` traceback.

```
  File ".../bin/nox", line 10, in <module>
    sys.exit(main())
  File ".../nox/_cli.py", line 204, in nox_main
    _main(main_ep=True)
  File ".../nox/_cli.py", line 231, in _main
    toml_config = load_toml(os.path.expandvars(noxfile), missing_ok=True)
  File ".../nox/project.py", line 63, in load_toml
    return _load_script_block(filepath, missing_ok=missing_ok)
  File ".../nox/project.py", line 75, in _load_script_block
    script = filepath.read_text(encoding="utf-8")
  ...
FileNotFoundError: [Errno 2] No such file or directory: 'noxfile.py'
```

### Expected Behavior

The friendly error message (and non-zero exit) that the missing-noxfile handler in `load_nox_module` already produces:

```
nox > Failed to load Noxfile /tmp/empty/noxfile.py, no such file exists.
```

### Steps To Reproduce

1. `mkdir /tmp/empty && cd /tmp/empty`
2. Run `nox -l`
3. See the traceback above instead of the friendly error.

### Environment

```Markdown
- OS: macOS (Darwin 24.6.0)
- Python: 3.14.2
- Nox: 2026.04.10
```

### Anything else?

#881 added an early call in `nox/_cli.py:231` to scan the noxfile for PEP 723 metadata:

https://github.com/wntrblm/nox/blob/c537fe16e4f0f074d7923af047ad558cdd24505d/nox/_cli.py#L169

It should probably use `load_nox_module`-style loading, to handle missing-file instead of erroring out entirely.

SAY-5:
Closes #1094.

`load_toml("noxfile.py", missing_ok=True)` is called from `_cli.py` to scan for a PEP 723 script block, but it leaked `FileNotFoundError` when the noxfile didn't exist instead of falling back to the friendly "Failed to load Noxfile ... no such file exists" message produced later in `load_nox_module`. Treat a missing file the same as a missing script block when `missing_ok=True` so the existing handler runs.

### DIFF
--- a/nox/project.py
+++ b/nox/project.py
@@ -74,7 +74,12 @@ def _load_toml_file(filepath: Path) -> dict[str, Any]:
 
 def _load_script_block(filepath: Path, *, missing_ok: bool) -> dict[str, Any]:
     name = "script"
-    script = filepath.read_text(encoding="utf-8")
+    try:
+        script = filepath.read_text(encoding="utf-8")
+    except FileNotFoundError:
+        if missing_ok:
+            return {}
+        raise
     matches = list(filter(lambda m: m.group("type") == name, REGEX.finditer(script)))
 
     if not matches:
--- a/tests/test_toml.py
+++ b/tests/test_toml.py
@@ -109,3 +109,14 @@ def test_load_non_recognised_extension() -> None:
     msg = "Extension must be .py or .toml, got .txt"
     with pytest.raises(ValueError, match=msg):
         nox.project.load_toml("some.txt")
+
+
+def test_load_missing_script_missing_ok(tmp_path: Path) -> None:
+    filepath = tmp_path / "noxfile.py"
+    assert nox.project.load_toml(filepath, missing_ok=True) == {}
+
+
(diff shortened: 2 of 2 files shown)


henryiii (approved):

nox/project.py
```diff
@@ -74,6 +74,8 @@ def _load_toml_file(filepath: Path) -> dict[str, Any]:
 
 def _load_script_block(filepath: Path, *, missing_ok: bool) -> dict[str, Any]:
     name = "script"
+    if missing_ok and not filepath.is_file():
+        return {}
     script = filepath.read_text(encoding="utf-8")
```
henryiii: EAFP is better than LBYL

(try / except)

PR #1100, opened 2026-06-09
Sort classifier-derived Python versions

### LINKED ISSUE #1074: Sort the return value of `python_versions`

If called with `max_version=None`,  [`python_versions`](https://github.com/wntrblm/nox/blob/5a901362f78d0ac1a6e7ef4b8b856e7f944553ee/nox/project.py#L94-L143) returns a list of versions read from the Trove classifiers. I propose to sort that list before returning it, so that the oldest supported Python version is `python_versions()[0]`, and the most recent one `python_versions()[-1]`.

For instance:

```python
from packaging.version import Version
```

then this line

https://github.com/wntrblm/nox/blob/5a901362f78d0ac1a6e7ef4b8b856e7f944553ee/nox/project.py#L124 

becomes

```python
    return sorted(from_classifiers, key=Version)
```

Of course, any sane `pyproject.toml` will have the classifiers already sorted, but should we rely on common sense?

### Anything else?

I can prepare a pull request.

Himanshuagrawal4:
## Summary
- sort classifier-derived `python_versions()` results by default
- add `sort=False` to preserve classifier order when callers need it
- cover both default sorted behavior and opt-out behavior in tests

Closes #1074.

## Validation
- temporary venv `python -m pytest tests/test_project.py`
- `python -m compileall -q nox`
- git diff --check

### DIFF
--- a/nox/project.py
+++ b/nox/project.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 
 import packaging.requirements
 import packaging.specifiers
+import packaging.version
 from dependency_groups import resolve
 
 if TYPE_CHECKING:
@@ -94,7 +95,7 @@ def _load_script_block(filepath: Path, *, missing_ok: bool) -> dict[str, Any]:
 
 
 def python_versions(
-    pyproject: dict[str, Any], *, max_version: str | None = None
+    pyproject: dict[str, Any], *, max_version: str | None = None, sort: bool = True
 ) -> list[str]:
     """
     Read a list of supported Python versions. Without ``max_version``, this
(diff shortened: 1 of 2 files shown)


henryiii:
I think this is good, covers keeping the original order and also sorting, though not sure if it should be default on or default off. I _think_ default on is fine. Any arguments for this order vs. the opposite order? (Low to high vs. high to low).

Himanshuagrawal4:
Thanks! My thinking was that low-to-high matches the usual way supported Python versions are listed in classifiers and CI matrices: oldest supported first, newest last. It also makes the output read like the natural progression of supported versions, and adding a new Python version tends to extend the end of the list.

I do not feel strongly if the project prefers newest-first, but low-to-high felt like the least surprising default to me.

henryiii:
This matches the order from the non-classifier form, so I think it's good. :)

henryiii (approved):

Issue #1102, opened 2026-06-10
Code review over nox


henryiii:
I used Claude Fable 5 to do a sweep of the codebase looking for bugs, performance gains, simplifications, and modernizations. Here's the report. So far, I've had an excellent success rate with Fable on these sorts of sweeps; it tests most findings as it goes.

On the module stuff, I think using 3.15 lazy imports would likely be better than trying to fix this the old-fashioned way with getattr or something.

:robot: _AI text below_ :robot:

All spot-checks confirmed the agents' claims. Here's the consolidated review.

# Nox codebase review

I ran four parallel reviews (bugs, simplifications, modernizations, performance) over the ~6,400 lines in `nox/`, with empirical verification — the bug findings were reproduced against an editable install, and the performance numbers were measured on this machine. Overall the codebase is in very good shape (ruff `select = ["ALL"]`, mypy strict, 100% coverage, and the 3.9 drop was thorough), but there are several real bugs worth fixing.

## Bugs (ordered by severity)

1. **Empty parametrize crashes every nox invocation** — `nox/manifest.py:262` + `nox/_resolver.py:166`. A noxfile containing `@nox.parametrize("x", [])` makes *all* commands (even `nox -l`) fail with exit 3, because the placeholder session has `signatures=[]` and never enters the dependency graph, so the topo sort hits a raw `KeyError`. This is a regression — nox 2024.4.15 correctly reported "no parameters available" and ran the rest. Fix: seed the graph from `self._queue` too (e.g. `dependency_graph.setdefault(s, ...)`).

2. **Fully non-ASCII session names destroy `.nox`** — `nox/sessions.py:89` (`_normalize_path`). A name like `"测试"` normalizes to the empty string, so the venv `location` becomes the envdir itself; `_clean_location()` then `shutil.rmtree`s `.nox`, deleting every other session's venv. Fix: fall back to the existing sha1-hash branch when the normalized name is empty.

3. **Script mode installs deps with the wrong pip on Windows** — `nox/_cli.py:175`. The non-uv branch runs bare `["pip", "install"]` with the venv's PATH only in `env`; Windows' `CreateProcess` searches the *parent's* PATH, so the outer pip is used (or none found). The correct pattern is three lines below for nox itself: `shutil.which("pip", path=env["PATH"])`, or better, `<venv>/python -m pip`.

[...]

PR #1106, opened 2026-06-11
chore: remove Python 3.9-era pins/backports and tighten generator annotations

henryiii:

Small cleanup now that Python 3.9 support has been dropped:

- **`.pre-commit-config.yaml`**: un-pin `pytest<9` in the mypy hook's `additional_dependencies` — the pin existed only because pytest 9 requires Python 3.10+.
- **`.pre-commit-config.yaml`**: remove the `importlib_metadata` and `importlib_resources` backports from the mypy hook. They are imported nowhere — `grep -rn 'importlib_metadata\|importlib_resources' nox/ tests/ .github/ docs/ noxfile.py pyproject.toml` returns no matches; the code uses stdlib `importlib.metadata` / `importlib.resources`. (`tomli` is kept: `nox/project.py` still imports it on Python < 3.11.)
- **`nox/_cli.py`**, **`nox/sessions.py`**: tighten pure-generator return annotations from `Generator[X, None, None]` to `Iterator[X]`.

Verified locally: `prek run --all-files` (including the mypy hook with the slimmed dependency set) and the full test suite (688 passed) both pass.

Part of #1102.

### DIFF
--- a/nox/_cli.py
+++ b/nox/_cli.py
@@ -39,7 +39,7 @@ from nox.project import load_toml
 
 if TYPE_CHECKING:
     from argparse import Namespace
-    from collections.abc import Generator
+    from collections.abc import Iterator
 
 __all__ = ["execute_workflow", "main", "nox_main"]
 
@@ -72,7 +72,7 @@ def execute_workflow(args: Namespace) -> int:
 
 def get_dependencies(
     req: packaging.requirements.Requirement,
-) -> Generator[packaging.requirements.Requirement, None, None]:
+) -> Iterator[packaging.requirements.Requirement]:
     """
     Gets all dependencies. Raises ModuleNotFoundError if a package is not installed.
     """
--- a/nox/sessions.py
+++ b/nox/sessions.py
@@ -47,6 +47,7 @@ if typing.TYPE_CHECKING:
     import argparse
     from collections.abc import (
         Callable,
+        Generator,
         Iterable,
         Iterator,
         Mapping,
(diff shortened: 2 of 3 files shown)


nox/sessions.py
```diff
@@ -47,7 +47,6 @@
     import argparse
     from collections.abc import (
         Callable,
```
henryiii, line 49:
```suggestion
        Callable,
        Generator,
```

nox/sessions.py
```diff
@@ -73,7 +72,7 @@ def __dir__() -> list[str]:
 
 
 @contextlib.contextmanager
-def _chdir(path: str) -> Generator[None, None, None]:
+def _chdir(path: str) -> Iterator[None]:
```
henryiii:
```suggestion
def _chdir(path: str) -> Generator[None, None, None]:
```

This was intentional - contextmanager _should_ (it doesn't) force Generator, as it doesn't actually work on other Iterables.

nox/_cli.py
```diff
@@ -72,7 +72,7 @@ def execute_workflow(args: Namespace) -> int:
 
 def get_dependencies(
     req: packaging.requirements.Requirement,
-) -> Generator[packaging.requirements.Requirement, None, None]:
+) -> Iterator[packaging.requirements.Requirement]:
```
henryiii, line 75: This is not a context manager, so it can be an Iterable. It is technically a Generator, but Iterable is a subset.

nox/sessions.py
henryiii: Eventually (3.14+ or something like that) we can drop the `None, None` though.

PR #1112, opened 2026-06-11
refactor: assorted simplifications and dead-code removal

henryiii:

A batch of reviewed, behavior-preserving simplifications, grouped into four commits. The only user-visible change is the first one.

### User-visible fix
- **`nox/_options.py`**: the `--download-python` flag string was passed twice to its `Option`, so `nox --help` rendered it as "`--download-python, --download-python`". One copy removed; argparse behavior is otherwise identical.

### Deduplication
- **`nox/virtualenv.py`**: the uv-install and python-build-standalone-install blocks in `VirtualEnv._resolved_interpreter` were duplicated verbatim between the `"always"` match arms and the `"auto"` default. They are now a single private `_install_python` helper; the `match` only needs `download_python` since the backend dispatch lives in the helper (backends are validated to `{uv, venv, virtualenv}` in `__init__`).
- **`nox/logger.py`**: `NoxFormatter` and `NoxColoredFormatter` had byte-identical `format()` bodies and `_simple_fmt` setup — factored into a `_SimpleOutputMixin`. The idempotent `logging.addLevelName` calls moved from `LoggerWithSuccessAndOutput.__init__` to module level next to the level constants (they ran on every logger instantiation; module import registers them once before any logger exists).
- **`nox/registry.py`**: `name or func.__name__` was computed twice into two variables; now computed once.
- **`nox/manifest.py`**: `__contains__`/`__getitem__` built throwaway `self._queue + self._consumed` lists; both now iterate `itertools.chain(...)`, and `__contains__` shares the same match expression via `any()`.
- **`nox/_parametrize.py`**: `Param.__str__` builds its output with a single generator expression instead of two intermediates.

### Micro-simplifications
- **`nox/virtualenv.py`**: `CondaEnv.is_offline()` checked `gethostbyname(...) is None`, but `gethostbyname` returns a `str` or raises — the check was always `False`, so the success path now returns `False` directly (the `# pragma: no cover` on the DNS-failure path is unchanged). Also dropped a redundant `f"{version}"` wrapper, and `venv_params or []` (default `()`) is now `list(venv_params)`, which is equivalent for every existing caller but always stores an owned list.
- **`nox/command.py` / `nox/_option_set.py`**: deleted runtime `collections.abc` imports that duplicated the `TYPE_CHECKING` imports; the names are annotation-only under `from __future__ import annotations`.
- **`nox/sessions.py`**: `nox.project.load_toml()` always returns a `dict`, so the `or {}` fallback was dead; `sys.platform.startswith("win32")` switched to the codebase's usual `startswith("win")` (equivalent on CPython, where the Windows value is exactly `"win32"`).
- **`nox/_resolver.py`**: removed a no-op trailing `else: return` and an unnecessary `nonlocal visited` (only item-assignment), and replaced the post-sort filter's redundant `hash(node) == hash(root)` clause (equality already implies equal hashes; the root is an `object()` sentinel) with a plain generator expression.
- **`nox/manifest.py`**: `if not len(self._queue)` → `if not self._queue`.
- **`nox/tox_to_nox.py`**: the `skip_install`/`use_develop` if/else collapses to `config[name][option] = section[option] != "False"` under the existing truthiness guard — identical semantics for the string values tox emits.
- **`nox/project.py`**: `list(filter(lambda ...))` → list comprehension.
- **`tests/test_sessions.py`**: the `SessionRunner` mock func now sets `venv_params = []` to match the real `Func` API (it previously leaked a bare `Mock` into `VirtualEnv`).

### Vestigial removals
- **`nox/sessions.py`**: `_normalize_path`'s `isinstance(path, bytes)` decode branch was a Python 2 leftover — the only caller passes `str`. Branch removed, annotation narrowed to `str`, and the bytes case dropped from its direct unit test.
- **`nox/manifest.py`**: `Manifest.next()` was a Python 2-style alias for `__next__`, referenced only by its own test (not in docs). Removed; the test uses the `next()` builtin.
- **`nox/_decorators.py`**: `FunctionDecorator` was a one-method class whose only subclass anywhere was `Func`; its `__new__` is folded into `Func` and the `cast` dropped. Nothing else imports it (checked `nox/`, `tests/`, `docs/`).

### Skipped items
- None — all 17 reviewed items were applied.

### Verification
- `pytest tests/` — 688 passed locally (macOS), plus the tox-gated `test_tox_to_nox.py` suite run with tox installed.
- Coverage compared against `main` with `coverage run -m pytest -m "not conda"`: no new uncovered lines or branches on any touched file; remaining misses are the pre-existing Windows-/conda-/extra-gated regions.
- `prek -a` — all hooks pass (ruff, strict mypy, codespell, etc.).

Part of #1102.

### DIFF
--- a/nox/_decorators.py
+++ b/nox/_decorators.py
@@ -29,22 +29,11 @@ if TYPE_CHECKING:
 
 T = TypeVar("T", bound=Callable[..., Any])
 
-__all__ = ["Call", "Func", "FunctionDecorator", "_copy_func"]
+__all__ = ["Call", "Func", "_copy_func"]
 
 
 def __dir__() -> list[str]:
     return __all__
-
-
-class FunctionDecorator:
-    """This is a function decorator."""
-
-    def __new__(  # noqa: PYI034
-        cls: Any, func: Callable[..., Any], *args: Any, **kwargs: Any
-    ) -> FunctionDecorator:
-        self = super().__new__(cls)
-        functools.update_wrapper(self, func)
-        return cast("FunctionDecorator", self)
 
 
 def _copy_func(src: T, name: str | None = None) -> T:
@@ -63,8 +52,15 @@ def _copy_func(src: T, name: str | None = None) -> T:
     return cast("T", dst)
 
 
-class Func(FunctionDecorator):
+class Func:
     """This is a function decorator that adds additional Nox-specific metadata."""
+
+    def __new__(  # noqa: PYI034
+        cls, func: Callable[..., Any], *args: Any, **kwargs: Any
+    ) -> Func:
+        self = super().__new__(cls)
+        functools.update_wrapper(self, func)
+        return self
 
     def __init__(
         self,
--- a/nox/_options.py
+++ b/nox/_options.py
@@ -548,7 +548,6 @@ options.add_options(
     _option_set.Option(
         "download_python",
         "--download-python",
-        "--download-python",
         noxfile=True,
         group=options.groups["python"],
         default=lambda: os.getenv("NOX_DOWNLOAD_PYTHON"),
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -139,9 +139,8 @@ def test_iteration() -> None:
     assert len(manifest._consumed) == 1
     assert len(manifest._queue) == 1
 
-    # The .next() or .__next__() methods can be called directly according
-    # to Python's data model.
-    bar = manifest.next()
+    # The second item should be our "bar" session.
+    bar = next(manifest)
     assert bar.func == sessions["bar"]
     assert bar in manifest._consumed
     assert bar not in manifest._queue
--- a/nox/_cli.py
+++ b/nox/_cli.py
@@ -80,7 +80,7 @@ def get_dependencies(
 
     def expand(
         req: packaging.requirements.Requirement,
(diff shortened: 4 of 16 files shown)


nox/_options.py
```diff
@@ -548,7 +548,6 @@ def _tag_completer(
     _option_set.Option(
         "download_python",
         "--download-python",
-        "--download-python",
```
henryiii, line 551: I have no idea how this happened. :)

nox/_decorators.py
```diff
@@ -29,24 +29,13 @@
 
 T = TypeVar("T", bound=Callable[..., Any])
 
-__all__ = ["Call", "Func", "FunctionDecorator", "_copy_func"]
```
henryiii, line 32: I think this is okay, it's undocumented an not used anywhere.

tests/test_manifest.py
```diff
@@ -139,9 +139,8 @@ def test_iteration() -> None:
     assert len(manifest._consumed) == 1
     assert len(manifest._queue) == 1
 
-    # The .next() or .__next__() methods can be called directly according
-    # to Python's data model.
-    bar = manifest.next()
+    # The second item should be our "bar" session.
```
henryiii:
```suggestion
```

PR #1119, opened 2026-06-17
chore: update pre-commit (and dependabot)

henryiii:
Adding pre-commit to dependabot, bumping monthly, with a 7 day cooldown. Also bumping our (slightly old) pins.

- **chore: update dependabot (1M, pre-commit, 7days)**
- **chore: bump pre-commit**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,19 +19,19 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/tox-dev/pyproject-fmt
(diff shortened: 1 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -44,15 +44,14 @@ repos:
           - dependency-groups>=1.2
           - humanize
           - jinja2
-          - orjson # Faster mypy
```
henryiii, line 47: Not needed anymore (uses sqlite now), plus this broke free-threaded Python

PR #1120, opened 2026-06-18
ci: reduce the number of times we contact conda-forge servers on non-windows

henryiii:
I've pulled out just the reduced number of conda-forge server interactions from #1118, which is an easy win and should make the other parts of that PR easier to review on thier own.


henryiii:
Ahh, copied this from a stale checkout, so it has the explanation for why it's different from before still. 😳

PR #1117, opened 2026-06-17
fix: support uv venv backend under MSYS2/MinGW (#1088)

henryiii:
Okay, I think this is a valid fix for #1088. @Tiphereth-A, would be good if you could take a look to be sure. :)

:robot: _AI text below_ :robot:

Fixes #1088.

### Problem

Under MSYS2/MinGW, `uv` cannot inspect the native interpreter — it reports an unknown operating system (e.g. `mingw_x86_64_ucrt_gnu`). Because script mode runs `uv venv -p <sys.executable> ...`, venv creation failed even though a bare `uv venv` works fine.

### Fix

- In `VirtualEnv.create`, skip `-p` for the `uv` backend on MinGW when no interpreter is explicitly requested, letting uv pick the default (which works). An explicit interpreter is still passed through, and non-MinGW behavior is unchanged.
- uv builds a Windows-style venv (`Scripts`/`python.exe`) even under MinGW, whereas the native interpreter and virtualenv/venv use a POSIX layout (`bin`/`python`). A new `_windows_layout` property centralizes this so `bin_paths` and the environment-reuse check select the right directory per backend.

Based on the patch from @Tiphereth-A in msys2/MINGW-packages#28759.

### Tests

Added tests for the MinGW uv layout, the MinGW no-`-p` command, and the explicit-interpreter uv path. `nox/virtualenv.py` and `tests/test_virtualenv.py` pass lint and the full `test_virtualenv.py` suite.

Issue #1088, opened 2026-04-23
Creating virtual environment failed when using uv in MSYS2


Tiphereth-A:
### Current Behavior

I am trying to build nox in MSYS2, and I have found that nox fails to create virtual environment when using MSYS2-build uv as the venv backend, whereas `uv venv` works fine.

### Expected Behavior

nox should create venv successfully.

### Steps To Reproduce

1. Clone this repo
2. Install all dependencies under MSYS2
3. run `python -m nox`
4. It reports

```
nox > Creating virtual environment (uv) using python.exe in .nox\_nox_script_mode
nox > Command uv venv -p 'C:\msys64\ucrt64\bin\python.exe' 'K:\nox\.nox\_nox_script_mode' --clear failed with exit code 2:
nox > error: Failed to inspect Python interpreter from provided path at `C:\msys64\ucrt64\bin\python.exe`
  Caused by: Can't use Python at `C:\msys64\ucrt64\bin\python.exe`
  Caused by: Unknown operating system: `mingw_x86_64_ucrt_gnu`

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "K:\nox\nox\__main__.py", line 34, in <module>
    main()
    ~~~~^^
  File "K:\nox\nox\_cli.py", line 204, in nox_main
    _main(main_ep=True)
    ~~~~~^^^^^^^^^^^^^^
  File "K:\nox\nox\_cli.py", line 267, in _main
    run_script_mode(
    ~~~~~~~~~~~~~~~^
        noxfile,
        ^^^^^^^^
    ...<4 lines>...
        download_python=download_python,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "K:\nox\nox\_cli.py", line 171, in run_script_mode
    venv.create()
    ~~~~~~~~~~~^^
  File "K:\nox\nox\virtualenv.py", line 879, in create
    nox.command.run(cmd, silent=True, log=nox.options.verbose or False)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "K:\nox\nox\command.py", line 208, in run
    raise CommandFailed(msg)
nox.command.CommandFailed: Returned code 2
```

### Environment

```Markdown
- OS: Windows 11
[...]

PR #1122, opened 2026-06-27
ci: experimental MSYS2/MinGW test run

henryiii:
Adding MSYS2 to help us with #1117.

## What

Adds an **experimental, non-blocking** (`continue-on-error: true`) CI job that runs the test suite under a real **MSYS2 / MINGW64** Python, plus the test/noxfile changes needed for the non-uv MinGW paths to pass.

This PR is **independent of #1117** (no shared commits; either can merge first).

## Why

`nox/virtualenv.py` has MinGW-specific logic gated on `_IS_MINGW` that, until now, was only covered by **mocked** unit tests (`@mock.patch("nox.virtualenv._IS_MINGW", new=True)`) — nothing ever ran it with `_IS_MINGW` actually `True`. This job exercises those paths on a genuine MSYS2 runner and feeds coverage into the `cover` job.

## What's included

- **CI job** (`msys2`): MSYS2/MINGW64, virtualenv backend (forced, so the interpreter under test stays MinGW), uv from `astral-sh/setup-uv` with `path-type: inherit` (no uv CLI exists in MSYS2 repos — the `msys` `uv` package only ships the `uv_build` backend), `mingw-w64-x86_64-gcc` for native build deps. Python version is whatever MSYS2 ships (detected at runtime).
- **noxfile:** skips installing `uv` into the test session on MinGW (no wheel; the uv tests `@has_uv`-skip when absent).
- **Test corrections** for the native MinGW venv layout — these fix behavior that predates #1117 (the `bin/` layout from #901, not the uv backend):
  - `test_bin_paths` / `test_create` — gate the `Scripts/`-layout assertions on `not IS_MINGW`; `test_create` gains a MinGW branch (POSIX `bin/` dir holding `.exe` executables — `python.exe`/`pip.exe`, lowercase `lib`).
  - `test_bin_windows` — pins `_IS_MINGW=False` so the Windows case is unaffected by the live MinGW host.
  - All are no-ops off MinGW.

## Remaining red is #1117's, by design

With this PR's corrections in place, the only remaining `msys2` failures are the **uv-backend** tests, which fail with `uv venv -p .../bin/python.EXE → Failed to inspect Python interpreter` — i.e. the **#1088 bug that #1117 fixes** (`test_uv_creation`, `test_create_reuse_uv_environment`, `test_stale_environment[uv-virtualenv-False]`, and the three `test_noxfile_script_mode*`). The job is `continue-on-error`, so the overall run is green; once #1117 lands it is fully green too (**726 passed, 15 skipped**, 100% coverage gate intact).

## Caveats

- MSYS2 is a rolling release; its Python version and packages move over time. Kept non-blocking until proven stable — could be promoted to a required check later (ideally after #1117 lands so the job is green).
- uv cannot introspect the MinGW interpreter (the limitation #1117 documents), so the uv venvs those tests create use uv's own CPython; the uv code path runs for real, but the interpreter inside those particular venvs is not MinGW.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,11 +75,9 @@ jobs:
           include-hidden-files: true
           if-no-files-found: error
 
-  # Experimental: exercise the MinGW-specific code paths (nox.virtualenv._IS_MINGW)
-  # on a real MSYS2 runner. Non-blocking until proven stable.
   # No uv CLI exists in any MSYS2 repo (the msys `uv` package only ships the
   # uv_build backend), so the uv backend is exercised via the standalone uv
-  # binary from setup-uv, which lands on the Windows PATH the MINGW64 shell sees.
+  # binary from setup-uv
   msys2:
     runs-on: windows-latest
     continue-on-error: true
@@ -94,7 +92,7 @@ jobs:
         with:
           msystem: MINGW64
           update: true
-          # inherit the runner PATH so the standalone uv (added to GITHUB_PATH
+          # Inherit the runner PATH so the standalone uv (added to GITHUB_PATH
           # by setup-uv, which runs after this step) is visible to MinGW Python
           path-type: inherit
           install: >-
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -75,8 +75,54 @@ jobs:
           include-hidden-files: true
           if-no-files-found: error
 
+  # No uv CLI exists in any MSYS2 repo (the msys `uv` package only ships the
```
ognevny, line 78: but what about https://packages.msys2.org/base/mingw-w64-uv? it may be broken a bit (https://github.com/msys2/MINGW-packages/issues/21091), but it exists

.github/workflows/ci.yml
```diff
@@ -75,8 +75,54 @@ jobs:
...
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v6.0.3
+        with:
+          persist-credentials: false
+      - uses: msys2/setup-msys2@v2.32.0
+        with:
+          msystem: MINGW64
```
ognevny, line 93: we recommend using UCRT64 environment, because MINGW64 is about to be deprecated

henryiii: Thanks! I'll try to make a follow up soon

PR #1124, opened 2026-06-28
[WIP] Add a new method Session.run_capture() which allows to capture stdout and stderr separately, even while showing the result

### LINKED ISSUE #1053: Allow session.run(quiet=True) to return stdout and stderr separately

When calling programs that print machine-readable output on stdout (for example as JSON) and warnings on stderr (for example as text messages), having the combined output is not helpful since you now have valid JSON mixed with lines that make it invalid. Being able to obtain these outputs separately (for example as with `subprocess.run`) would make this a lot easier then trying to separate the combined output into parsable JSON and unrelated text.

This could be an additional parameter to `session.run()`, like `capture: bool = False`. Though that function already has quite a few arguments.

`capture` could work by returning `None` if the command isn't run, or a tuple `(stdout, stderr)` (or `(stdout, stderr, rc)`, while we're at it - see "Anything else?" below).

Directly using `subprocess.run()`. Getting that to use the session's venv though and global config parameters like `no_install` is non-trivial, and I'm not sure whether it's fully possible at all without accessing private nox internals.

### Anything else?

Having access to the exact return code (and the possibility to disable nox's return code check) could also be helpful in some cases.

felixfontein:
`Session.run_capture()`'s interface is similar to `Session.run()`, except:

* It always returns `tuple[str, str, int]` (unless the command isn't run, then it returns `None`) with return code, stdout, and stderr;
* If `silent=True` is specified and the program exits abnormally, stderr is shown; if `silent=False` (default), all stdout and stderr is shown in near real-time (similar to [`tee`](https://en.wikipedia.org/wiki/Tee_(command)), you catch the output while the user can still see it);
* There is a `success_all: bool` parameter which, when set to `True`, does not fail on any return code; that allows the caller to completely handle the return code, without having to provide an exhaustive list of possible return codes for `success_codes`.

The implementation of `silent=False` was somewhat tricky. I found https://stackoverflow.com/questions/5045771/python-how-to-prevent-subprocesses-from-receiving-ctrl-c-control-c-sigint and the resulting https://github.com/pycontribs/subprocess-tee/blob/main/src/subprocess_tee/__init__.py implementation, but that didn't work as I expected in particular with respect to `KeyboardInterrupt`. I had to add a SIGINT signal handler and cancel tasks myself. (By default SIGINT kills the event loop, and not even `finally` blocks in tasks are handled... You can only catch `KeyboardInterrupt` completely outside the `loop.run_until_complete()` call.)

I also considered not using asyncio, but selecting on stdout/stderr, i.e. basically reimplementing `Popen.communicate()`. But then I looked closer in how `Popen.communicate()` is implemented in the standard library, and decided it's way too complex to try to do it this way, *especially* when it also has to work on Windows.

This is a work in progress since it doesn't have tests yet. Also I only did a few tests on Linux (I don't have access to Windows, but can do some manual testing on macOS later). Before I spent even more time I'm also curious on whether this has a chance of being merged :)

Fixes #1053.

### DIFF
--- a/nox/popen.py
+++ b/nox/popen.py
@@ -170,11 +170,10 @@ class _TeeSubprocess:
     def __init__(
         self,
         *,
-        loop: asyncio.AbstractEventLoop,
         interrupt_timeout: float | None,
         terminate_timeout: float | None,
     ) -> None:
-        self.loop = loop
+        self.loop = asyncio.new_event_loop()
         self.stdout: list[bytes] = []
         self.stderr: list[bytes] = []
         self.interrupt_timeout = interrupt_timeout
@@ -353,6 +352,7 @@ class _TeeSubprocess:
             )
         finally:
             self.loop.remove_signal_handler(signal.SIGINT)
+            self.loop.close()
         sys.stdout.flush()
         if is_canceled:
             raise KeyboardInterrupt
@@ -371,9 +371,7 @@ def tee_popen(
     Return a tuple ``(return_code, stdout, stderr)``.
     Standard output and standard error are also printed to ``sys.stdout``.
     """
-    loop = asyncio.get_event_loop()
     teesub = _TeeSubprocess(
-        loop=loop,
         interrupt_timeout=interrupt_timeout,
         terminate_timeout=terminate_timeout,
     )
--- a/nox/popen.py
+++ b/nox/popen.py
@@ -180,6 +180,8 @@ class _TeeSubprocess:
         self.terminate_timeout = terminate_timeout
         self.collectors: list[asyncio.Task[None] | asyncio.Task[int]] = []
         self.main_task: asyncio.Task[bool] | None = None
+        self.is_terminating = False
+        self.do_terminate = False
 
     async def _shutdown_process(self, proc: Process) -> None:
         """Gracefully shutdown a child process."""
@@ -203,6 +205,8 @@ class _TeeSubprocess:
         Return ``True`` if the process had a non-zero exit code during shutdown.
         """
         try:
+            if self.do_terminate:
+                self._handle_sigint()
             await proc.wait()
         except asyncio.CancelledError:
             # SIGINT causes the task to be cancelled. We first need to uncancel it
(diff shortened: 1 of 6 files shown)


henryiii:
Edit: ahh, sorry, too many tabs open, got comments mixed up. I tried something new here, I had Claude Fable plan out what it would do, but base it on this PR, so it has two comments you can look at. Should I push them here or make a PR to your fork so you can let me know what you think?

felixfontein:
I guess both is fine for me; whatever you prefer. Pushing here has the advantage of all discussion happening here (and not also in PRs in other repos.)

nox/popen.py
```diff
@@ -100,14 +147,197 @@ def popen(
...
+        self,
+        stream: asyncio.StreamReader,
+        out_stream: BinaryIO,
+        out_buffer: list[bytes],
+    ) -> None:
+        """
+        Read a stream chunk by chunk, append it to ``out_buffer``, and write it to ``out_stream``.
+        """
+        while True:
+            chunk = await stream.readline()
```
gotmax23: `stream.readline()` will fail if the line is too long. See https://github.com/ansible-community/antsibull-core/commit/d910be2ec7f521160cba57d9f1d74b59e4596346.

felixfontein: Thanks, I've re-implemented that here in 3c6064fb11676e3a632b88cbf1887c37a3ae3b99. The main difference to your implementation is that it has a lot of comments that explain what exactly is happening, without having to resort to the docs or asyncio.streams' sources.

felixfontein:
I tried to debug the failing tests a bit on a Macbook; there the tests pass if the tests are run sequentially (without `--numprocesses=auto`), but some always fail when things are run in parallel. The "crashed" workers are due to `KeyboardInterrupted` escaping from tests; adding `try`/`except KeyboardInterrupt as exc: raise ValueError("Caught KeyboardInterrupt") from exc` around all of them avoided the "crashed" workers, but still had the other problems.

I don't have access to Windows, so no idea how to work around the `NotImplementedError` there...

felixfontein:
I'm wondering whether it won't be easier to simply stop using async functions and explicitly reimplement this using threads.

felixfontein:
@henryiii I'm still curious about the comments (commits?) you promised ;)

PR #1129, opened 2026-07-05
feat: add `return_code` property to `CommandFailed` exception

### LINKED ISSUE #1128: Include return_code in CommandFailed

Right now, it is impossible to get hold of the return code of a command run by `Session.run()` if `success_codes` is specified and contains more than one element.

I would suggest to extend the `CommandFailed` exception as follows:
```diff
--- a/nox/command.py
+++ b/nox/command.py
@@ -45,9 +45,10 @@ ExternalType = Literal["error", True, False]
 class CommandFailed(Exception):
     """Raised when an executed command returns a non-success status code."""
 
-    def __init__(self, reason: str | None = None) -> None:
+    def __init__(self, reason: str | None = None, *, return_code: int | None = None) -> None:
         super().__init__(reason)
         self.reason = reason
+        self.return_code = return_code
```
and modify `run()` as follows:
```diff
--- a/nox/command.py
+++ b/nox/command.py
@@ -206,7 +207,7 @@ def run(
                 logger.error(output)
 
             msg = f"Returned code {return_code}"
-            raise CommandFailed(msg)
+            raise CommandFailed(msg, return_code=return_code)
 
         if output:
             logger.output(output)
```
Then callers can catch `CommandFailure` and ignore / specially handle specific return codes, instead of having to pass these codes as `success_codes` without being able to react on them.

Right now you can catch `CommandFailure` and extract the return code from its message, which is clumsy and error-prone.

Adding another option to `Session.run()` to allow returning the return code as well would make the interface more complex (and typing more challenging).

Obviously this can also be achieved with #1124 (by using the new functionality added there), but this is a much more low-cost and low-maintenance solution.

### Anything else?

_No response_

felixfontein:
Example implementation of #1128.

Fixes #1128.

### DIFF
--- a/nox/command.py
+++ b/nox/command.py
@@ -45,9 +45,12 @@ ExternalType = Literal["error", True, False]
 class CommandFailed(Exception):
     """Raised when an executed command returns a non-success status code."""
 
-    def __init__(self, reason: str | None = None) -> None:
+    def __init__(
+        self, reason: str | None = None, *, return_code: int | None = None
+    ) -> None:
         super().__init__(reason)
         self.reason = reason
+        self.return_code = return_code
 
 
 def which(
@@ -206,7 +209,7 @@ def run(
                 logger.error(output)
 
             msg = f"Returned code {return_code}"
-            raise CommandFailed(msg)
+            raise CommandFailed(msg, return_code=return_code)
 
         if output:
             logger.output(output)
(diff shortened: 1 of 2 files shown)


henryiii (approved):
This looks great to me, had Claude look to see if there were any other places that it should be passed in, it said there weren't any, this is the only one with a proper subprocess.

felixfontein:
@henryiii (and Claude) thanks a lot for reviewing and merging!

PR #1131, opened 2026-07-07
feat: run sessions in parallel with --parallel/-j

### LINKED ISSUE #544: Support nox sessions running in parallel

This has brought up in other issues a few times (see #198 #88), on the latter one it seemed like there was some conflation of
two different issues (supporting session parallelism and allowing sessions to launch background tasks). So I've broken this out into it's own issue.

#198 Brings up the idea of running tasks in the background from within a session (i.e. spinning up a webserver with something like `session.background`)

This issue is to track the idea of allowing Nox sessions to run in parallel

The parallelism would be opt-in with a flag or some parameter to the session decorator (I have a preference for the former)

Such that you could run e.g. `nox --parallel` or `nox --jobs 4` or similar so that the selected sessions would run in parallel.

My only real ideas for implementation could be a threadpool or wrapping each nox session in it's own subprocess as suggested in https://github.com/theacodes/nox/issues/88#issuecomment-383637043. I'm not sure which is best here.

You could also spin up an asyncio event loop but I suspect this would require a lot of code changes as `async` and `await` tend to spread everywhere as soon as you introduce them. There's also the issue that asyncio is never _actually_ parallel so it's speedup factor would be heavily dependent on the work done by the sessions (CPU vs IO bound).

I think this would be reasonably complex to implement, especially coordinating and handling the output from the sessions (see https://github.com/theacodes/nox/issues/88#issuecomment-383637043 also) but I think, if done correctly would be hugely powerful

There's no real one I can think of, other than some sort of `xargs` hackery?

henryiii:
Suggested order:

- #1145 
- #1144 
- This PR can then lose the long duplicated option list

TODO:

- [x] Mark experimental: note it in the docs and add a warning for now

Adding parallel job execution. Each session runs in its own `nox` subprocess with its output buffered and printed as a contiguous block when it finishes; an interactive terminal additionally shows a live status board of running sessions.

Design features:

- `-j`/`--parallel N` to run nox in parallel. (`NOX_PARALLEL`/`nox.options.parallel` work too, of course)
- Opt-in via `@nox.session(allow_parallel=True)` or `nox.options.allow_parallel` or `--allow-parallel`. Default is non-parallel, though like other settings you can set this at the global level to true, then opt-out on some sessions instead. You can try it without modifying a noxfile by using the CLI. Warning shown if nothing opts in and parallel is requested.
- If `allow_parallel` is false, those sections are exclusive. (Like for collecting coverage.)
- `requires=` dependencies work
- `notify` does _not_ work, and throws an error in parallel mode. Use `requires=` instead.
- `session.interactive` is `False` in parallel sessions.

Design choices:

- Should based-on-cores be "auto" (current) or 0?
- There's quite a bit of duplication in `_parallel.py`, so I made some PRs (above) to refactor.

https://github.com/user-attachments/assets/b1bdeab6-8c06-443a-be1a-fbabc36bfed3

Closes #544.

Assisted-by: ClaudeCode:claude-opus-4.8
Assisted-by: ClaudeCode:claude-fable-5
Assisted-by: Codex:gpt-5.5 (for review)
Assisted-by: Codex:gpt-5.6-sol (for review)


henryiii:
> You are an adversarial reviewer for the new feature in this branch. Do you see any problems? Anything that could be done or written more cleanly? Can you break it?


<details>
<summary>:robot: AI review :robot:</summary>

All finders and verification are done. Here's the adversarial review.

## What I actually broke (reproduced live)

**1. `session.notify` under `--parallel` runs the notified session twice, concurrently.** With `a` (notifies `c`) and `c` both selected, `nox -j4` printed `c`'s body twice: once inside `a`'s child (`--no-dependencies` doesn't suppress notify) and once as `c`'s own scheduled child. With a real virtualenv, two processes build/use the same envdir simultaneously — exactly the corruption the `busy_envdirs` guard exists to prevent, and the summary says "Ran 2 sessions" while 3 ran. The docs saying notify is unsupported doesn't stop anyone; children should hard-error (or at least warn-and-drop) on `notify()` when launched by the parallel runner.

**2. `nox.options.parallel = "bogus"` in a noxfile crashes with a raw traceback** (uncaught `ValueError` out of `run_manifest` at `nox/tasks.py:409`), instead of the `logger.error(...) → exit 3` style everywhere else. Related inconsistency: `nox.options.parallel = 0` is silently treated as sequential (the `if raw_parallel else 1` falsy check), while `--parallel 0` on the CLI errors.

**3. A stale `NOX_PARALLEL=bogus` env var breaks every nox command**, including `nox --version` and `nox -l` — argparse applies the `type` converter to string defaults at parse time. The message is clean, but no other nox env var kills unrelated invocations like this.

**4. `--no-dependencies` is a public flag that silently disables safety checks.** `nox --no-dependencies -s deploy` where `deploy` has `requires=["no_such_session"]` (a typo) runs happily; without the flag it's exit 3 with "Session not found". It also skips cycle detection and the prerequisite-succeeded guard in `SessionRunner.execute`. If it's "mainly used internally", consider validating `requires` anyway and only skipping queueing/result-checking — or hide the flag.

**5. `\r` handling is dead code and progress bars explode.** `Popen(text=True)` uses universal newlines, which translates `\r` to `\n` — so `_preview_text`'s `rsplit("\r", 1)` never fires, and a session printing `"10%\r50%\r100%"` produced three separate lines in the buffered block (verified). A real pip/uv progress bar that detects a pseudo-TTY would bloat the buffer and the final output. Pass `newline=""` (or read bytes) if the `\r` logic is to work as commented.

## Confirmed by inspection, not run

**6. Ctrl-C leaves grandchildren running.** The `KeyboardInterrupt` handler at `nox/_parallel.py:501` only calls `proc.terminate()` on the child nox processes — no kill escalation (nox already has `popen.shutdown_process` with terminate→kill timeouts), and SIGTERM to a child nox kills it without it cleaning up *its* subprocess (pytest, compilers), which keeps running and writing to the terminal. Process groups (`start_new_session=True` + killpg) would fix both.

**7. `--stop-on-first-error` semantics quietly changed.** Sequentially, nothing after the failure runs or is reported. In parallel, in-flight sessions run to completion and land in the summary and `--report` JSON. Probably the right tradeoff, but it should be documented; the docs currently only say "stops launching new sessions".

**8. `_child_argv` is a drift trap — and has already drifted.** It hand-forwards ~18 options, and `--add-timestamp` is already missing: `nox -ts -j2` silently loses timestamps in child output. Every future global option must be remembered here or parallel/sequential behavior diverges silently. Worth a test that diffs the forwarded set against the option registry, or a declarative "forward this option" attribute on `Option` itself.

**9. Unbounded output buffering.** Each child's full output is held in a parent-side `list[str]` until it finishes. A verbose multi-hour session (or several) can eat RAM the sequential path never would. You already have a `TemporaryDirectory` per session — spooling to a file there is nearly free.

## Cleanliness (smaller)

- `_Reporter.aborted()` is `finished()` minus the harmless `_active`/`_preview` pops — since `ABORTED` is falsy, `finished(name, result, "")` does the identical thing. One method can go.
- The `on_line` preview callback runs an ANSI-regex sub plus an RLock acquire **per output line even when there's no TTY** (CI). Pass `on_line=None` when `not reporter.tty`, and even on a TTY consider storing the raw line and stripping only at render time (4 Hz) rather than per line.
- `_COLORS`/`_make_color_formatter` hand-roll ANSI codes while `tasks.py` already uses `colorlog`'s `parse_colors` for the session list — two palettes that will drift.
- `_render` formats each running line twice (plain `head` for width math, then a rebuilt colored `line`) — the layout lives in two places that must agree.
- Board cosmetics: a SKIPPED child counts toward `passed`.
- One thing I tried to break and couldn't, which deserves a comment: the `while progressed:` fixpoint loop in `schedule_ready` looks removable but is load-bearing — when a failure aborts a whole dependency chain and no futures remain, the cascade must complete in that single call or downstream sessions vanish from the summary (I verified the chain `root→mid→leaf` all report `aborted`). A reviewer *will* try to simplify it away.

Things that held up under attack: selector round-tripping with quotes/spaces in parametrized signatures, mixed exclusive/parallel scheduling, abort cascades, exit codes matching sequential, the `--parallel 1` recursion guard, fake-parent sessions being stripped before the queue, and NOX_PARALLEL/noxfile-recursion via the explicit `--parallel 1` override.

```json
[
  {"file": "nox/_parallel.py", "line": 474, "summary": "session.notify in a parallel child runs the notified session concurrently with its independently scheduled run, racing on the same envdir", "failure_scenario": "Session a notifies c; both selected with allow_parallel=True; nox -j2 runs c twice at once (reproduced) — with a venv backend, two processes build .nox/c simultaneously and corrupt it"},
  {"file": "nox/tasks.py", "line": 409, "summary": "Noxfile-set nox.options.parallel is parsed unguarded: invalid values crash with a raw traceback and 0/\"\" are silently treated as sequential", "failure_scenario": "nox.options.parallel = \"bogus\" → uncaught ValueError traceback (reproduced) instead of logger.error + exit 3; nox.options.parallel = 0 runs sequentially with no error while --parallel 0 on the CLI errors"},
  {"file": "nox/_options.py", "line": 657, "summary": "An invalid NOX_PARALLEL env var makes every nox invocation fail, including nox --version and nox -l", "failure_scenario": "export NOX_PARALLEL=bogus → argparse applies _parallel_arg to the string default on every parse, so even nox --version exits with an argument error (reproduced)"},
  {"file": "nox/tasks.py", "line": 265, "summary": "Public --no-dependencies flag silently skips requires validation (typos, cycles) and the prerequisite-success guard", "failure_scenario": "Session with requires=[\"no_such_session\"] runs successfully under --no-dependencies (reproduced) where the resolver previously reported the missing session and exited 3"},
  {"file": "nox/_parallel.py", "line": 501, "summary": "Ctrl-C only sends terminate() to the immediate child nox processes: no kill escalation and grandchild processes are orphaned", "failure_scenario": "nox -j4 running pytest in each child; Ctrl-C SIGTERMs the child nox, which dies without cleaning up pytest, so the test processes keep running; a child ignoring SIGTERM is never killed (popen.shutdown_process already implements the escalation)"},
  {"file": "nox/_parallel.py", "line": 369, "summary": "text=True universal-newline translation converts \\r to \\n, making _preview_text's rsplit(\"\\r\") dead code and multiplying progress-bar redraws into separate buffered lines", "failure_scenario": "A child printing \"10%\\r50%\\r100%\" yields three lines in the output block (reproduced); a real progress bar redrawing per tick bloats the buffered output that sequential mode renders as one line"},
  {"file": "nox/_parallel.py", "line": 316, "summary": "_child_argv hand-forwards options and has already drifted: --add-timestamp is not forwarded, and future options will silently diverge under -j", "failure_scenario": "nox -ts -j2: children run without --add-timestamp so buffered session output loses requested timestamps; any new global option is silently ignored in parallel mode unless someone remembers this list"},
  {"file": "nox/_parallel.py", "line": 496, "summary": "--stop-on-first-error only stops new scheduling: in-flight sessions run to completion and appear in the summary and --report, unlike sequential mode", "failure_scenario": "nox -j4 -x with 4 independent sessions: one fails first, the other 3 still complete and land in the report where sequential mode would have run only the failing one"},
  {"file": "nox/_parallel.py", "line": 356, "summary": "Each child's entire output is buffered unbounded in parent RAM until the session finishes", "failure_scenario": "Several long verbose sessions (GBs of log output) in flight hold every line in memory and can OOM the parent, where sequential mode streams to the terminal; spooling to the already-created temp dir avoids it"},
  {"file": "nox/_parallel.py", "line": 236, "summary": "_Reporter.aborted() duplicates finished(): since ABORTED is falsy, finished(name, result, \"\") already increments _failed and emits the identical block", "failure_scenario": "Two board/emit paths that must be edited in lock-step; a future fix to locking or block format applied to one silently diverges the abort path"}
]
```

The one thematic takeaway: the design (re-invoking `nox` per session) is sound and held up well under direct attack — the surviving problems are all at the *boundaries* of that design: option forwarding (№2, №3, №7), the child's residual autonomy (№1 notify), and signal/lifecycle handling (№5). Tightening the child contract — "children may not notify, must receive the full forwarded option set, and die with their process group" — would close most of this list at once.

</details>

edgarrmondragon:
This one's exciting!

In case it's useful, I think the per-session output is explicitly colored in black so it has little contrast on dark terminals:


https://github.com/user-attachments/assets/24c9f025-f815-415c-8975-248a2f6128aa

henryiii:
It works for me, the "bold black" color shows up as grey in my color scheme. I'm trying "thin" from color log instead now, is that better? I'm not sure what "thin" actually is, I'm guessing it's the 'dim' code, which hopefully is pretty well supported these days. Probably need to check and see how it looks on Windows.

edgarrmondragon:
> I'm trying "thin" from color log instead now, is that better? I'm not sure what "thin" actually is, I'm guessing it's the 'dim' code, which hopefully is pretty well supported these days.

That does improve the contrast!

https://github.com/user-attachments/assets/8364475d-1db6-4280-91cc-4a1361b3d428

I also tried on another terminal with a different color scheme, and the "bold" variant looks better than https://github.com/wntrblm/nox/pull/1131#issuecomment-5113226689, but again the "thin" variant has better contrast.

> Probably need to check and see how it looks on Windows.

Can't help with that unfortunately, but happy to check on macos with different terminals and color schemes.

PR #1142, opened 2026-07-20
feat: support requires-python in script mode

henryiii:
This is a followup to #1086 adding support for `requires-python` from PEP 723 in noxfiles. This means you can write a noxfile for a modern version of Python, and from older Pythons, nox will relaunch itself with that Python, just like it currently does with dependencies.

It started pretty simple but several iterations of review caught some things (stale rebuild, for example). For beta versions, I went with requiring the beta flag, for example, `>=3.15.0b1` would be required to get a 3.15 resolution while it's in beta, since `3.15b1` does not pass `>=3.15` in PEP 440 versions.

Based on #1148, which carries the general interpreter-range work this feature uses (upper-bound-only downloads and the conda range spec fix).

- If the running interpreter fails the spec, nox re-execs into an environment built from a matching interpreter, downloading a version from the range if none is found (controlled by the existing `--download-python` settings).
- A reused script environment is rebuilt if its Python no longer satisfies the spec.
- `requires-python` without `dependencies` implies `nox` itself.
- The check uses the full PEP 440 version built from `sys.version_info`, so a beta predates its final release and dev builds like `3.13.0+` cannot break parsing.
- The `none` script backend errors on a mismatch, since it cannot switch interpreters.

An invalid specifier or an unsatisfiable spec exits with a clear message instead of a traceback.


henryiii:
Dropping the output of `/simplfy` here, as there were a couple of things we could possibly do in a followup:

Tests pass (205 passed) and lint is clean. Summary of the `/simplify` pass:

**Fixed (6):**
- `nox/_cli.py:373` — merged the two-step `dependencies` branch into one condition (`if dependencies is not None or requires_python is not None:` with the `["nox"]` default inside), so "script mode is active" is expressed once. Behavior for an explicit empty `dependencies = []` is unchanged.
- `nox/_cli.py:196` — `_venv_python_version` now passes the existing python-discovery disk cache (`nox.virtualenv._get_python_discovery_cache()`) to `PythonInfo.from_exe`, with a fallback to an uncached probe on `OSError` (mirroring `_discover_interpreter`). This removes a subprocess launch from every script-mode startup that reuses an env with `requires-python`.
- `nox/virtualenv.py:252` — `_concrete_install_target` now builds the `SpecifierSet` directly from the parsed clauses (`f"{op}{version.removesuffix('t')}"`) instead of regex-stripping a serialized string, which also let the near-unreachable `InvalidSpecifier` catch go away.
- `tests/test_virtualenv.py:403` — `test_condaenv_python_dep` uses the existing `make_conda` fixture instead of hand-constructing `CondaEnv` (the comment justifying the bypass was wrong — the fixture never touches a conda binary).
- `tests/test__cli.py:251` — `make_fake_script_venv` merges dicts before constructing the `SimpleNamespace` instead of mutating `__dict__` after.
- `tests/test__cli.py:332` — dropped the dead `calls = script_mode_exec` alias.

**Skipped (noted, intentionally):**
- Moving the stale-reuse recreate logic from `run_script_mode` into `VirtualEnv._clean_location`'s reuse-validation hooks (altitude finding) — the right long-term home, but a real refactor of the venv layer well beyond this diff.
- Splitting `check_requires_python`'s parse-and-check into separate steps to remove the ordering comment in `_main` — would ripple through three call sites and tests for modest gain.
- Verifying the lower-bound branch of `_concrete_install_target` against the full specifier set (like the new upper-bound branch does) — changes pre-existing behavior, belongs in its own change.
- A shared PEP 440 "matches" helper with `_version.py`, the conda first-character spec classifier, and parametrizing the two similar download-range tests — all judged too minor to be worth the churn.

henryiii:
Okay, this is under 500 lines (not counting removed lines), and mostly tests, the changes are clean now.

PR #1148, opened 2026-07-31
fix: download upper-bound-only Python ranges

henryiii:
I'm trying to make #1124 simpler, and this was a required but generally useful fix from that PR.

- `feat`: an upper-bound-only range (`<3.14`) had no installable floor, so `--download-python` could not help. The highest `X.Y` below the tightest upper bound is installed instead (`<3.14` installs `3.13`), stepping down past `!=` exclusions (`<3.14,!=3.13.*` installs `3.12`). Free-threaded `t` suffixes are preserved.

Also updates the `config.rst` sentence that said ranges with no lower bound cannot be downloaded.


felixfontein:
I think this should reference to #1142, not to #1124, in case anyone is confused by the reference in the future :)

PR #1160, opened 2026-08-11
fix: support legacy encodings in parallel reporter

### LINKED ISSUE #1159: `nox --parallel` fails when reporting success on `windows-2025`
### Current Behavior

While trying out the new `--parallel` flag, I ran into an issue where `nox` is failing when reporting success on `windows-2025`:

```
  File "D:\a\_temp\setup-uv-cache\archive-v0\ehbmMaLZ8RjVP79z\Lib\site-packages\nox\_parallel.py", line 509, in worker
    reporter.finished(name, result, output)
  File "D:\a\_temp\setup-uv-cache\archive-v0\ehbmMaLZ8RjVP79z\Lib\site-packages\nox\_parallel.py", line 279, in finished
    self._emit_block(name, result, output)
  File "D:\a\_temp\setup-uv-cache\archive-v0\ehbmMaLZ8RjVP79z\Lib\site-packages\nox\_parallel.py", line 243, in _emit_block
    self.stream.write(
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'charmap' codec can't encode character '\u2713' in position 11: character maps to <undefined>
```

The same session exits sucessfully on `macos-26`, `macos-26-intel`, `ubuntu-24.04`, and `ubuntu-24.04-arm`.

### Expected Behavior

The session exits successfully also on `windows-2025`

### Steps To Reproduce

Run `uvx nox -s tests` using, e.g., this `noxfile.py`: 

```python
import nox

nox.needs_version = ">=2026.08.10"
nox.options.default_venv_backend = "uv"
nox.options.parallel = 5

PYTHON_ALL_VERSIONS = ["3.10", "3.11", "3.12", "3.13", "3.14"]

@nox.session(python=PYTHON_ALL_VERSIONS, reuse_venv=True, default=True, allow_parallel=True)
def tests(session: nox.Session) -> None:
    session.run("uv", "run", "pytest")
```

### Environment

```Markdown
- OS: Windows
- Python: 3.12 (but more may be affected)
- Nox: 2026.08.10
```

### Anything else?

_No response_

nightcityblade:
Fixes #1159

## Summary

- Preserve Unicode status symbols when the output stream supports them.
- Fall back to ASCII status symbols for legacy encodings such as Windows `cp1252`.
- Add a regression test that writes a successful parallel-session result to a `cp1252` stream.

## Test Plan

- `uv run pytest tests/test_parallel.py -q` (57 passed)
- `uvx prek run --files nox/_parallel.py tests/test_parallel.py`
- `uv run coverage report --show-missing nox/_parallel.py` (100% statement and branch coverage)

### DIFF
--- a/nox/_parallel.py
+++ b/nox/_parallel.py
@@ -67,6 +67,7 @@ if TYPE_CHECKING:
     from nox.sessions import SessionRunner
 
 _SPINNER = "⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏"
+_ASCII_SPINNER = "|/-\\"
 _SYMBOLS = {
     Status.SUCCESS: "✓",
     Status.SKIPPED: "⊘",
@@ -127,6 +128,15 @@ def _status_symbol(status: Status, encoding: str | None) -> str:
     except UnicodeEncodeError:
         return _ASCII_SYMBOLS[status]
     return symbol
+
+
+def _spinner_frame(spin: int, encoding: str | None) -> str:
+    frame = _SPINNER[spin % len(_SPINNER)]
+    try:
+        frame.encode(encoding or "utf-8")
+    except UnicodeEncodeError:
+        return _ASCII_SPINNER[spin % len(_ASCII_SPINNER)]
+    return frame
 
 
 @dataclasses.dataclass(kw_only=True)
@@ -206,7 +216,7 @@ class _Reporter:
             header = plain_header[: width - 1]
         lines = [self._banner(width), header]
 
-        frame = _SPINNER[self._spin % len(_SPINNER)]
+        frame = _spinner_frame(self._spin, self.stream.encoding)
         for name, start in self._active.items():
             # Plain and colored renderings are built from the same segments so
             # the width math can't drift from what is actually displayed.
--- a/nox/_parallel.py
+++ b/nox/_parallel.py
@@ -73,6 +73,12 @@ _SYMBOLS = {
     Status.FAILED: "✗",
     Status.ABORTED: "↯",
 }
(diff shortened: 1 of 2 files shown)


nox/_parallel.py
edgarrmondragon, line 69: Should these braille symbols also be handled? I don't think they're compatible with cp1252, but perhaps not as big an issue since they're only displayed on TTYs?

henryiii: I think it would be nice to handle too, probably.

nightcityblade: Good catch — agreed. Addressed in `8491918` by using the braille spinner only when the output stream's encoding can represent it; legacy encodings now get an ASCII `|/-\` animation instead. I added cp1252 regression coverage for the TTY render path.

Validation: `tests/test_parallel.py` (58 passed) and the full pre-commit suite on both changed files (including Ruff and mypy) passed.

nightcityblade:
CI note: the prior `dist` run built the sdist and wheel successfully and passed `check-wheel-contents`; only `twine check --strict` failed because the isolated build produced Metadata-Version 2.5 and Twine rejected that metadata version. The latest `main` CD run passed at this PR's exact base SHA (`57f9e10`), while this branch changes only `_parallel.py` and its tests, so I left the unrelated packaging configuration out of this fix.

PR #1166, opened 2026-08-13
Fix the zombie process leak issue in the _parallel.py file

wu-zhao-min:
Problem description (EN): The _stop_procs function in the _parallel.py file sends SIGKILL (via _signal_group(proc, kill=True) ) to child processes that do not exit within the terminate timeout, but it does not call proc.wait() afterwards to reap the killed process.

Hazard statement (EN): During parallel interrupt/shutdown, each SIGKILL ed child process becomes a zombie (POSIX) or retains an unreleased process handle (Windows). Under long-running parallel stress, zombies accumulate, leaking file descriptors and exhausting the PID table, which may eventually cause new process spawning to fail.

Repair solution (EN): Add a proc.wait() call immediately after _signal_group(proc, kill=True) in the except subprocess.TimeoutExpired branch to reap the killed process and release its file descriptors/handle.

### DIFF
--- a/nox/_parallel.py
+++ b/nox/_parallel.py
@@ -477,6 +477,7 @@ def _stop_procs(procs: Collection[subprocess.Popen[bytes]]) -> None:
             proc.wait(timeout=max(0.0, deadline - time.monotonic()))
         except subprocess.TimeoutExpired:  # noqa: PERF203 - cold shutdown path
             _signal_group(proc, kill=True)
+            proc.wait()
 
 
 def run_manifest_parallel(


henryiii:
I don't believe this is true. If you have a reproduction, or clear details of how it can actually happen, open an issue. Report below.

Every child process is created inside `with subprocess.Popen(...) as proc:` in a worker thread (see `_Children.spawn`). When `_stop_procs` sends SIGKILL, the child's stdout hits EOF, the worker's read loop ends, and the `Popen` context manager's `__exit__` calls `proc.wait()` — that reaps the process and closes its pipes/handles. The executor also joins those worker threads before `run_manifest_parallel` returns, so the reap always happens.

There's also no way for zombies to accumulate: `_stop_procs` runs once, on `KeyboardInterrupt`, and nox re-raises and exits right after. The window between the SIGKILL and the worker's reap is a momentary (and normal) zombie state, not a leak.

One concern with the change itself: `_signal_group` suppresses `PermissionError`, so if the kill was suppressed, the added unbounded `proc.wait()` could block the interrupt path forever — the surrounding code uses deadline-bounded waits for that reason.

PR #1171, opened 2026-08-26
docs: clarify supported session Python versions

### LINKED ISSUE #681: Document supported (session) Python versions

I'm unable to find documentation on which old Python version sessions nox intends to support.

https://github.com/wntrblm/nox/issues/525#issuecomment-1000337239 says
> And to be clear, this about installing Nox on Python 3.6. Nox will still create and run sessions on old Pythons.

It would be nice to elaborate on exactly which "old Pythons" in the docs :)

Find the above info in docs.

_No response_

### Anything else?

I've found that at least 3.3 is doable currently but requires an additional `virtualenv<15.2` constraint, ditto 3.4 with `virtualenv<20.5`.

Not sure if there's anything nox could do to make that Just Work without users having to figure out and specify the constraints.

Himanshuagrawal4:
### Summary

- document how interpreter discovery and the selected environment backend
  determine session Python support
- avoid implying that all Nox sessions use virtualenv
- point readers targeting old Python versions to the existing EOL guidance

### Testing

- targeted pre-commit suite
- Sphinx HTML build

Closes #681

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -123,6 +123,13 @@ You can tell Nox to use a different Python interpreter/version by specifying the
     .. _python-discovery: https://pypi.org/project/python-discovery/
     .. _PEP 514: https://peps.python.org/pep-0514/
 
+Nox does not impose a separate supported-version range on the Python
+interpreters used by sessions. A target version works when Nox can discover or
+download its interpreter and the selected environment backend can create an
+environment with it. Backends and the packages they use may have their own
+Python support ranges; see the end-of-life Python guidance below when targeting
+older versions.
+
 You can also tell Nox to run your session against multiple Python interpreters. Nox will create a separate virtualenv and run the session for each interpreter you specify. For example, this session will run twice - once for Python 3.13 and once for Python 3.14:
 
 .. code-block:: python
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -41,7 +41,7 @@ To run both of these sessions, just run::
 
     nox
 
-For each session, Nox will automatically create a `virtualenv`_ with the appropriate interpreter, install the specified dependencies, and run the commands in order.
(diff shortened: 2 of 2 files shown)


Himanshuagrawal4:
The CI jobs were cancelled before running any steps. GitHub reports that a hosted runner could not pick them up after multiple attempts. I tried rerunning CI, but GitHub requires repository admin rights. Could a maintainer rerun it when reviewing? The Read the Docs build passed. The separate CD run also reports a workflow startup error: https://github.com/wntrblm/nox/actions/runs/32985808654.

docs/config.rst
```diff
@@ -123,6 +123,13 @@ You can tell Nox to use a different Python interpreter/version by specifying the
     .. _python-discovery: https://pypi.org/project/python-discovery/
     .. _PEP 514: https://peps.python.org/pep-0514/
 
+Nox does not impose a separate supported-version range on the Python
+interpreters used by sessions. A target version works when Nox can discover or
+download its interpreter and the selected environment backend can create an
+environment with it. Backends and the packages they use may have their own
+Python support ranges; see the end-of-life Python guidance below when targeting
+older versions.
```
henryiii, line 131:
```suggestion
Python support ranges. Any non end-of-life Python should work, and
a few older Python versions usually work too depending on the backend.
```

Maybe this? I think currently uv is 3.7+, virtualenv is somewhere around 3.8+, venv I think is 3.6+ (limit from python-discovery). Not sure about conda.

Himanshuagrawal4: Applied your suggested wording in 5bae189, thanks. The hooks for docs/config.rst pass, and a warning-as-error Sphinx build passes with the two Markdown symlinks materialized in a temporary build copy on Windows. The earlier macOS 3.14 failure was a Conda dependency-resolution conflict during setup; this change only touches the wording. Codex helped apply and check this follow-up.

Issue #1021, opened 2025-10-28
nox virtualenv backend uses the sys.executable to run virtualenv with the -p flag targeting the interpreter, which leads to incorrect packages installed in the virtual enviroment


zzzeek:
### Current Behavior

the virtualenv backend runs virtualenv with the "-p" option to indicate the desired interpreter.   for obsolete versions of python like 3.7, this installs packages in terms of the system interpreter (apparently!  maybe this is a bug in virtualenv. seems strange).   This breaks immediately with current pip version on Python 3.7, see reproducer.    using the venv backend, it uses the resolved interpreter all the way and things work correctly.

### Expected Behavior

Assuming this is what virtualenv does, I would imagine we'd want to run virtualenv using the resolved interpreter also, not using the "-p" option.   there seem to be options to virtualenv on how to install pip/setuptools also, but anyway, see the reproducer, something to ...document ?  understand?  not really sure.

### Steps To Reproduce

noxfile:

```
import nox

@nox.session(python=["3.7"])
def helloworld(session: nox.Session) -> None:
    session.install("typing-extensions")

```

now make sure you have an old 3.7 interpreter and a new interpreter like 3.13 around.

make a py 3.13 environment and install nox using python 3.13:

```
[jenkins@9d93bd4d66c5 foo]$ /opt/python3.13/bin/python3 -m venv .venv
[jenkins@9d93bd4d66c5 foo]$ .venv/bin/pip install nox
Collecting nox
  Using cached nox-2025.10.16-py3-none-any.whl.metadata (5.1 kB)
  # ... packages install ...

Collecting virtualenv>=20.15 (from nox)
  Using cached virtualenv-20.35.3-py3-none-any.whl.metadata (4.6 kB)

# ... more
Installing collected packages: distlib, platformdirs, packaging, humanize, filelock, colorlog, attrs, argcomplete, virtualenv, dependency-groups, nox
Successfully installed argcomplete-3.6.3 attrs-25.4.0 colorlog-6.10.1 dependency-groups-1.3.1 distlib-0.4.0 filelock-3.20.0 humanize-4.14.0 nox-2025.10.16 packaging-25.0 platformdirs-4.5.0 virtualenv-20.35.3

[notice] A new release of pip is available: 25.2 -> 25.3
[notice] To update, run: /tmp/foo/.venv/bin/python3 -m pip install --upgrade pip

```

[...]

PR #1178, opened 2026-09-06
fix: skip uv backend when the session Python is below uv's floor

### LINKED ISSUE #833: Setting venv_backend='uv|virtualenv' breaks if 'uv' is installed but the version of python your session is running against isn't supported by uv.
### Current Behavior

`nox.options.default_venv_backend = "uv|virtualenv"` will always try to use `uv` if it is installed, even when the session's python version is incompatible with `uv`.

### Expected Behavior

Nox would fall back to the alternate venv backend if the python version requested is earlier than the earliest supported version of the backend.

### Steps To Reproduce

1. Create a nox file with the following contents:
```python
import nox
import sys

# prefer UV if it is available
# Note for older versions of nox we can't actually use this syntax.
nox.options.default_venv_backend = "uv|virtualenv"


@nox.session(python=[f"{sys.version_info.major}.{sys.version_info.minor}", "3.6"])
def hello(session):
    "A simple test session"
    session.install("cowsay")
    session.run("cowsay", "-t", f"Hello from Python {session.python}")
```
2. Invoke `nox -s hello-3.8`. Output:
```plaintext
nox > Running session hello-3.8
nox > Creating virtual environment (uv) using python3.8 in .nox/hello-3-8
nox > uv pip install cowsay
nox > cowsay -t 'Hello from Python 3.8'
  _____________________
| Hello from Python 3.8 |
  =====================
                     \
                      \
                        ^__^
                        (oo)\_______
                        (__)\       )\/\
                            ||----w |
                            ||     ||
nox > Session hello-3.8 was successful.
```
3. Invoke `nox -s hello-3.6`. output:
```plaintext
nox > Running session hello-3.6
nox > Creating virtual environment (uv) using python3.6 in .nox/hello-3-6
nox > Command uv venv -p python3.6 /grmn/prj/aoem/tools/build/bakelite/nox-simple-reproduction/.nox/hello-3-6 failed with exit code 1:
  × Can't use Python at `/usr/bin/python3.6`
  ╰─▶ Python 3.6.15 is not supported. Please use Python 3.8 or newer.
nox > Session hello-3.6 failed.
```

I would expect the output for step 3 to be:
```plaintext
nox > cowsay -t 'Hello from Python 3.6'
Error processing line 1 of /grmn/prj/aoem/tools/build/bakelite/nox-simple-reproduction/.nox/hello-3-6/lib/python3.6/site-packages/_virtualenv.pth:

    File "/usr/lib/python3.6/site.py", line 174, in addpackage
      exec(line)
    File "<string>", line 1, in <module>
    File "/grmn/prj/aoem/tools/build/bakelite/nox-simple-reproduction/.nox/hello-3-6/lib/python3.6/site-packages/_virtualenv.py", line 3
                                       ^
  SyntaxError: future feature annotations is not defined

Remainder of file ignored
Error processing line 1 of /grmn/prj/aoem/tools/build/bakelite/nox-simple-reproduction/.nox/hello-3-6/lib/python3.6/site-packages/_virtualenv.pth:

    File "/usr/lib/python3.6/site.py", line 174, in addpackage
      exec(line)
    File "<string>", line 1, in <module>
    File "/grmn/prj/aoem/tools/build/bakelite/nox-simple-reproduction/.nox/hello-3-6/lib/python3.6/site-packages/_virtualenv.py", line 3
                                       ^
  SyntaxError: future feature annotations is not defined

Remainder of file ignored
  _____________________
| Hello from Python 3.6 |
  =====================
                     \
                      \
                        ^__^
                        (oo)\_______
                        (__)\       )\/\
                            ||----w |
                            ||     ||
nox > Session hello-3.6 was successful.
```

### Environment

```Markdown
- OS: ubuntu 22.04
- Python: 3.8, 3.6
- Nox: 2024.4.15

Note: nox is invoked in python 3.8. Python 3.6 is installed using deadsnakes.
```

### Anything else?

One possible solution would be to allow us to parameterize the venv_backend so that we can explicitly invoke the desired backend with the desired python version.  But it seems simpler to just add logic to consider `uv` absent if our target session python is earlier than python 3.8

eknathdj:
## Problem

Closes #833.

`nox.options.default_venv_backend = "uv|virtualenv"` always picks `uv` when uv is
installed, even for a session whose Python uv refuses to build. uv fails instead
of the fallback being used:

```
nox > Creating virtual environment (uv) using python3.6 in .nox/hello-3-6
nox > Command uv venv -p python3.6 ... failed with exit code 1:
  × Can't use Python at `/usr/bin/python3.6`
  ╰─▶ Python 3.6.15 is not supported. Please use Python 3.8 or newer.
```

`get_virtualenv` chose the backend from availability alone and never looked at
its own `interpreter` argument.

## Change

`get_virtualenv` now also skips a backend that is known to reject the requested
interpreter. `uv` is the only backend with a known floor (Python 3.8), so
`uv|virtualenv` uses `virtualenv` for older sessions.

Two deliberate limits keep this conservative:

- **A backend is only skipped while a fallback remains.** As the last candidate
  it is still used, so a plain `venv_backend="uv"` keeps reporting uv's own
  error rather than nox's "No backends present".
- **Only a concrete `major.minor` can rule uv out.** A bare name, a path, or a
  PEP 440 range resolves to no definite version, so the backend keeps making its
  own decision and behavior there is unchanged.

The skip is logged at debug level so the switch isn't silent.

This follows the approach you suggested for the equivalent `virtualenv`/EOL case
in #1021 — "automatically dropping back to venv when virtualenv is targeting
versions it can't support" and hard-coding the supported range. The version is a
single documented constant, `_UV_MIN_PYTHON`.

## Tests

Added to `tests/test_virtualenv.py`, all hermetic — no uv binary, no subprocess,
no network:

- falls back for `3.6`, `python3.7`, `pypy3.7`, `3.7t`
- keeps uv for `3.8`, `3.12`, `python3.13`, `pypy3.10`
- keeps uv when the version is unknown: `None`, `python`, `>=3.6`, `<3.8`
- keeps uv when it is the only backend, so uv's own error still surfaces

Verified on Windows with Python 3.14:

- `pytest tests/test_virtualenv.py` → 168 passed, 43 skipped
- `pytest tests/test_sessions.py` → 163 passed, 1 skipped
- `ruff check` and `ruff format --check` clean on both changed files
- `coverage` reports no missing lines or partial branches in the new code

## Docs

`docs/usage.rst` and `docs/config.rst` said the first *available* backend is
selected, which is no longer the whole story; both now mention the uv floor.

---

AI assistance was used while investigating and preparing this change. I reviewed
the result and ran the tests, lint, and coverage above myself. Happy to adjust
the approach, the constant, or the log level.


henryiii:
Ran a quick review. Not entirely correct, but some important points. Will try to review properly once back from traveling.

<details>
<summary>Fable 5.1 medium /code-review:</summary>

The code review of commit 6387eea (uv backend floor fallback) finished. Findings, most severe first:

1. **The floor constant may be wrong.** The reviewer found uv 0.12.12 only rejects Python below 3.6, so `_UV_MIN_PYTHON = (3, 8)` demotes working 3.6 and 3.7 sessions to virtualenv. The floor also depends on uv version. Suggested fix: key it on the uv version alongside the existing 0.4.16 and 0.8 gates, or drop the constant and fall back when `uv venv` fails. The docs in config.rst and usage.rst repeat the 3.8 claim.

2. **The "fallback left" guard is positional, not availability based.** With `uv|conda`, uv present, conda absent, and `python="3.7"`, uv is skipped and the user gets "No backends present" even though uv is installed. Compute the available backends first and only skip when the unsupported one is not the last available.

3. **`pypy-3.7` bypasses the floor.** The check parses the raw spec before the `pypy-` normalisation at nox/virtualenv.py:749, and that spelling parses with no major version, so it is treated as unknown and keeps uv. `pypy3.7` falls back. Normalise before the check.

4. **Range specs that resolve below the floor keep uv.** Specs like `<3.8` or `==3.7.*` are classified unknown, but `_concrete_install_target` already turns them into a concrete version. This reproduces the original #833 symptom. Reuse that helper so both paths share one interpretation.

5. **The skip is only logged at debug**, and uv-specific `venv_params` are passed to virtualenv unchanged. A user sees an unexplained virtualenv failure. Log at info like the existing "uv binary not found" line.

6. **Missing commit trailer.** The commit has no `Assisted-by:` line. Amend before opening a PR.

7. **Lazy import cost.** The new `python_discovery` import runs on every `uv|fallback` session, including the reuse path, at about 21 ms per process. A local regex for major.minor would avoid it.

8. **Test duplication.** Three tests in tests/test_virtualenv.py share a body and differ only in inputs. One parametrized test would cover all cases.

I did not independently verify the uv 3.6 floor claim in item 1. It is the one that changes the design, so it is worth confirming with `uv venv -p 3.7` before acting on the rest.

</details>

PR #1180, opened 2026-09-08
fix: fall back to ASCII for the parallel banner and status header

dylanpulver:
`#1160` gave the parallel reporter ASCII fallbacks for its spinner and status symbols. Two other non-ASCII literals in the same file still reach the console unchecked, and one of them is the very first thing `nox --parallel` writes.

`_Reporter.__enter__` writes the banner verbatim in non-TTY mode (`_parallel.py:174`), and `_EXPERIMENTAL` contains an em dash:

```
cp437     banner(em dash): UnicodeEncodeError ('—')     spinner+symbol (#1160): ok
cp932     banner(em dash): UnicodeEncodeError ('—')     spinner+symbol (#1160): ok
ascii     banner(em dash): UnicodeEncodeError ('—')     spinner+symbol (#1160): ok
cp1252    banner(em dash): ok                           spinner+symbol (#1160): ok
```

So on a console using any common legacy encoding except cp1252, `--parallel` raises before a single session runs. The status header (`running N · passed N · …`) has the same problem with `·`, which fails on cp932 and ascii.

Which characters survive where:

| character | cp1252 | cp437 | cp850 | cp932 | ascii | latin-1 |
|---|---|---|---|---|---|---|
| `—` em dash, banner | ok | **fail** | **fail** | **fail** | **fail** | **fail** |
| `·` middle dot, header | ok | ok | ok | **fail** | **fail** | ok |
| `✓` symbols (fixed in #1160) | fail | fail | fail | fail | fail | fail |

`_encodable()` sits next to the existing `_status_symbol()` and `_spinner_frame()` and works the same way — try the console encoding, fall back on `UnicodeEncodeError` — so the banner and header degrade instead of raising. Only punctuation is translated (`—` to `-`, `·` to `|`); on a UTF-8 console nothing changes.

`pytest tests/`: **892 passed, 47 skipped, 1 xpassed**. Reverting `_parallel.py` fails exactly the two new tests and nothing else. `nox/_parallel.py` stays at **100% coverage** (0 missing, 0 partial branches) per CONTRIBUTING. `pre-commit run --files nox/_parallel.py tests/test_parallel.py` clean, including ruff, mypy and codespell.

https://claude.ai/code/session_011M5uTyCU4WcNTsPvGrErDo


henryiii:
Thanks!

PR #1181, opened 2026-09-10
fix: keep quoted paths when converting tox commands

r3wretrhy:
`tox-to-nox` split each tox command with `str.split()`. A quoted path with spaces such as `--path '/tmp/foo bar'` was therefore broken into extra argv tokens.

Use `shlex.split` so quoting is preserved, then wrap tokens the same way as before.

The added test fails on the unpatched tree and passes with the change.


henryiii (approved):

henryiii:
Looks good, thanks!

PR #1182, opened 2026-09-11
Expose parallel execution state on sessions

### LINKED ISSUE #1177: Allow user sessions to detect when they are running in parallel

Commonly, parallelism is configured at the level of a job. `sphinx-build -j auto`, `pytest-xdist`, etc.
When combined with a parallel runner (`nox -j auto`), this can result in many more parallel processes being spun up than there are available resources to run them, and make run-times net-net worse.

In `tox` usage, I have many projects where I detect parallel usage in this awkward way in a `toxfile.py`:
```python
@impl
def tox_before_run_commands(tox_env: ToxEnv) -> None:
    parallel_detected = tox_env.options.command in ("p", "run-parallel")
    if parallel_detected:
        setenv = tox_env.conf.load("set_env")
        setenv.update({"TOX_PARALLEL": "1", "PYTEST_XDIST_AUTO_NUM_WORKERS": "0"})
```

I would like to do something similar with the new `nox -j` flag, but it's not clear where I can read the value.

Imagine a modest matrix of ~20 jobs, each of which is run with pytest-xdist, for fast individual runs. When run with `nox -j`, they should all drop down to not using xdist. In practice, that's much faster than `20*num_cpus` processes.

My favorite possibility is some new API on the session object:
```python
if session.parallel:
    ...
```

More generically, possibly a way to read the options which were passed?
```python
# how would the generic option reader be shaped?
# would it expose options only from the CLI invocation?
if session.read_option("parallel") != 0:
   ...
```

In any case, this information needs to pass down to the individual subprocs for parallel sessions.

Reading the original `sys.argv` is possible but has a few downsides.
Probably there are other similar workarounds possible, but nothing looks today like a clean API.

### Anything else?

The addition of the parallel mode flag makes it possible for me to migrate projects from `tox`. I think it's a very significant step forward for `nox`, but this gap makes it harder to apply in existing projects.

nickita-khylkouski:
## Summary
- add Session.parallel so session code can detect Nox parallel workers
- mark child processes with an internal hidden option
- document using the property to avoid nested parallelism

Closes #1177

## Tests
- full pytest suite (921 passed, 25 skipped, 1 xpassed)
- Ruff formatting and checks
- mypy


henryiii (approved):
Looks good, similar to how we provide access to other things that nox sets like the venv backend.

PR #1184, opened 2026-09-19
fix: preserve Python literals in converted tox commands

jakezwang:
`tox-to-nox` generates invalid Python for commands such as `python -c "print('hello')"`: the argument becomes `'print('hello')'`. Backslashes can also change meaning, for example turning a Windows path's `\n` and `\t` into control characters.

Serialize each argument as a Python string literal after shell tokenization. Three regression cases run the converter against real tox configuration, parse the generated file, and check the resulting argument values. This complements the quoted-path handling in #1181.

Validation: all 16 converter tests, all pre-commit checks (including mypy), and the strict documentation build pass. The full non-conda suite has 925 passes and one failure in `test_noxfile_script_mode_url_req`; pristine main reproduces the same `2024.10.09` versus `2024.10.9` mismatch. Coverage is unchanged, with the modified line covered.


henryiii:
Tox produces different output on Windows, so that's why we have to add sys.platform dependence here. @gaborbernat Claude suggests that if `tox config --format json` left these as lists, then most of this custom parsing goes away and we could just use it as is. If that's something that you'd be interested in, I could open an issue in tox.

Short answer: tox has no code that does what nox is doing, and the closest thing it does have is the same shape and length as nox's function.

**tox only goes one way.** `StrConvert.to_command` (`tox/config/loader/str_convert.py`) turns the raw ini text into a `Command(args)` list. `shell_cmd` (`tox/execute/request.py:86`) turns args back into a display string. tox never reads that string back — it's for logging and for the `tox config` ini/json dumps. So there's no tox parser for nox to reuse or copy; nox is inverting a *display* format.

**tox's parse direction uses POSIX shlex on every platform.** On Windows it doesn't switch to a Windows parser; it runs a 40-line hand-rolled pre-pass, `_win32_process_path_backslash` (lines 102–141), that rewrites path backslashes so POSIX `shlex` won't eat them. That function is a character-by-character loop with `quote` and `path_start` state flags — the same structure and nearly the same length as nox's `_split_windows_command`. So 45 lines for this is not nox doing something unusual; it's what the problem costs.

**The asymmetry is the real finding.** On POSIX, `shell_cmd` uses `shlex.quote`, which `shlex.split` inverts exactly — nox's POSIX path is on solid ground. On Windows it uses `list2cmdline`, which happens to be exactly invertible, but nothing in tox guarantees that: it's a formatting helper with no round-trip contract, and tox changing it would silently break nox.

That points at the actually-simplest fix, which is upstream: `tox config --format json` already has `Command.args` in hand and chooses to stringify it. If tox emitted `commands` as lists in JSON mode, nox could delete `_split_windows_command`, `_split_command`, and both test blocks, and drop the `shlex`/`subprocess` lazy imports. Worth an issue on tox — want me to draft one?

gaborbernat:
Do you have an example of this? This technically would be a breaking change, but want to understand when and why you'd need this.

henryiii:
It's to avoid the 46 line (possibly fragile) function we are adding to undo this. We are running `tox config` so we can access the tox config programmatically as JSON, so having a list joined into a string according to a platform specific custom merge function that we just have to undo is counter-intuitive when we asked for structured JSON.

I know it's a breaking change which is why I asked here first. It would even break us at first. :) Not sure where/how this is currently being used. "No" is a valid answer. Could also be a new "STRUCTURED_JSON" output format or some other flag. If it's at all interesting, I can open an issue in tox.

henryiii:
This is for `tox-to-nox`, a tool that reads `tox` files and writes equivalent `nox` files. The old version read the files directly, but the tox 4 version uses `tox config --format JSON` instead.

gaborbernat:
Yeah, I read the description. I want some real configs of tox and outputs where the issue occurs.

henryiii:
The fix in this PR was `return ", ".join([f"'{item}'" for item in seq])` being changed to  `", ".join(repr(str(item)) for item in seq)`, and with that it hit problems in our own test suite on Windows. The new tests hit this bug (no old ones did AFAICT).

@jakezwang do you have an example of a toxfile that would hit this?

Three, all from `test_commands_preserve_python_literal_characters` (`tests/test_tox_to_nox.py:211`). Each builds a tox.ini of this shape with `commands` swapped:

```ini
[tox]
envlist = lint

[testenv:lint]
basepython = python3.13
commands = <command>
```

| `commands =` line             | Windows result before fix                       | Expected         |
| ----------------------------- | ----------------------------------------------- | ---------------- |
| `python -c "print('hello')"`  | `session.run('python', '-c', 'print(hello)')`   | `print('hello')` |
| `python --path 'C:\new\test'` | `session.run('python', '--path', 'C:newtest')`  | `C:\new\test`    |
| `python --name "O'Reilly"`    | `ValueError: No closing quotation` from `shlex` | `O'Reilly`       |

All three pass on Linux/macOS, because tox re-emits `tox config` with `shlex.quote` there and with `subprocess.list2cmdline` on Windows — only the Windows quoting confused the POSIX `shlex.split`.

The new `test_split_windows_command_roundtrip` (`:193`) covers the same class without needing Windows: it feeds `list2cmdline` output for argv lists like `["python", "--path", "C:\\ends\\with\\sep\\"]`, `["python", "--text", 'a "quoted" word']` and `["python", "--empty", ""]` back through `_split_windows_command`.

gaborbernat:
I meant from tox change POV, your AI text explains the impact on nox, not on tox input/output.

henryiii:
To print that properly I need to run on Windows. I can try to do that soon. Basically though this prints different results based on the OS it runs on: `tox config --format json`:

```ini
[tox]
envlist = lint

[testenv:lint]
basepython = python3.13
commands = python -c "print('hello')"
```

macOS trimmed:

```json
{
  "env": {
    "lint": {
      "commands": [
        "python -c 'print('\"'\"'hello'\"'\"')'"
      ],
    }
  }
}
```

Windows trimmed I have to run Windows for, but I think it's different from the tests above, it's making this with a Windows specific custom quoting.

By the way, the key quotes are missing from the colored default output on a terminal. Makes it hard to copy/paste. I did `tox config --format json | cat` to get something I could copy/paste.

gaborbernat:
What if the config is in TOML not INI?

PR #1185, opened 2026-09-19
feat: free threaded from classifiers

henryiii:
We should add the free threaded versions if there's a free-threading classifier present when pulling versions from classifiers. Can be enabled/disabled expliclity too.

Noticed in https://github.com/scikit-hep/vector/pull/745.


docs/cookbook.rst
```diff
@@ -182,6 +182,46 @@ By default, ``nox-uv`` also validates that the lockfile is up-to-date.
...
+      "Programming Language :: Python :: 3.15",
+      "Programming Language :: Python :: Free Threading :: 3 - Stable",
+    ]
+
+.. code-block:: python
+
+    import nox
+
+    PYPROJECT = nox.project.load_toml("pyproject.toml")
+    # ["3.12", "3.13", "3.14", "3.14t, "3.15", "5.15t"]
```
edgarrmondragon:
```suggestion
    # ["3.12", "3.13", "3.14", "3.14t, "3.15", "3.15t"]
```

typo?

henryiii: Yeah, I added this by hand. 😳

